PR #5, opened 2020-05-13
Catala Logo Proposal

kaepora:
I propose and contribute the following logo to the Catala project.

Here it is in PNG format for quick judging:


denismerigoux:
Thanks very Nadim for this beautiful proposal ! We're going to discuss it in a meeting tomorrow and I'll get back to you :)

denismerigoux:
So we discussed it at the meeting and we generally think that there are too many things going on in the logo. The scales of justice are definitely a keeper, however we're not fan of the QED black square and of the detailed book. 

So if we want to keep the scales of justice, maybe they can be associated with a element coming from the sequent calculus rules typographical aesthetic (https://en.wikipedia.org/wiki/Sequent_calculus) ? 

In this spirit, I'm thinking of something like this : 

```  
      (stylized legislative bill/book)  :  CATALA
   --------------------------------------------------------
                        (scales of justice)
``` 

The structure (A : B/C) corresponds to a default logic inference rule, the prerequisite being the legislative text, the justification being CATALA and the conclusion being justice :)

denismerigoux:
Something like this
logo

denismerigoux:
Or simpler like this
logo

denismerigoux:
As the logo settled on this https://github.com/CatalaLang/catala/raw/master/doc/logo.png, I'm closing the PR.

PR #8, opened 2020-05-13
Move Pygments up

kaepora:
This moves the Pygments subsection above the Generating website assets subsection, given that the former is a dependency to the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,37 @@ and install the library. Makefile aliases are here to help you: running
 
 builds the compiler from its OCaml sources.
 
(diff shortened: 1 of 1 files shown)


denismerigoux:
Thanks very much for this proposal, indeed the ways things were it didn't work and you could believe that you had to do `make pygments` for everything to work, etc. But now I added this dependency in the `Makefile`, and in the `README.md` the Pygments section is more explanatory than really necessary for building the website assets. So I'm closing the PR :)

Issue #9, opened 2020-05-14
Render mathematical signes in Unicode in literate programming output


denismerigoux:
To make things prettier. Involves modifying `src/catala/literate/{html.ml, latex.ml}`.

PR #12, opened 2020-05-15
Pretty print maths symbols in html and latex outputs

skodt:
#9

### DIFF
--- a/src/catala/literate/html.ml
+++ b/src/catala/literate/html.ml
@@ -153,13 +153,16 @@ let program_item_to_html (i : A.program_item) (custom_pygments : string option)
           | _ -> "#" )
           (pre_html (Pos.unmark a.law_article_name))
     | A.CodeBlock (_, c) | A.MetadataBlock (_, c) ->
-        let syms = R.regexp "!=|<=|>=|--|->" in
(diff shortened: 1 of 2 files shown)


skodt:
[allocations_familiales.pdf](https://github.com/CatalaLang/catala/files/4634642/allocations_familiales.pdf)

src/catala/literate/html.ml
```diff
@@ -153,11 +153,24 @@ let program_item_to_html (i : A.program_item) (custom_pygments : string option)
           | _ -> "#" )
           (pre_html (Pos.unmark a.law_article_name))
     | A.CodeBlock (_, c) | A.MetadataBlock (_, c) ->
+        let date = "\\d\\d/\\d\\d/\\d\\d\\d\\d" in
+        let syms = R.regexp (date ^ "|!=|<=|>=|--|->|\*|\/") in
```
denismerigoux, line 157: We're missing `=` here, in Catala that's the equal comparison and not definition so it should also be rendered in fancy font.

skodt: Something like ＝compared to the normal = (https://www.fileformat.info/info/unicode/char/ff1d/browsertest.htm)

PR #23, opened 2020-08-04
Cgi total rule

jemsab:
Here is the rule that completes article 1A of the French Tax Code, per my previous PRs. 

Concerning the use of the keywords "égal à" and "contexte", they do not appear to be defined in the "reading_guide.tex". Should I add an issue on the repo ?

### DIFF
--- a/examples/code_general_impots/Section_I_Dispositions_generales.catala
+++ b/examples/code_general_impots/Section_I_Dispositions_generales.catala
@@ -38,4 +38,11 @@ donnée bénéfices_non_commerciaux contenu montant
 donnée revenus_capitaux_mobiliers contenu montant
 donnée plus_values contenu montant
 donnée total contenu montant
+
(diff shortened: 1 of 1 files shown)


denismerigoux:
Great ! You nailed it. One small thing: could you uniformize the naming with _accents_ ? Right now some fields of `RevenuNetGlobal`have _accents_ in their names, but other don't. I would recommend putting _accents_ everywhere since the goal is to stick to the French text as close as possible.

Concerning the reading guide, you are right for the missing part. The truth is that the reading guide is completely out of date and should be redone. You can definitely file an issue of that. Moreover the reading guide should be written as HTML rather than in LaTeX because we want it to appear here : https://catala-lang.org/en/guide. There's already a tutorial for programmer that showcases most of the syntax here : https://catala-lang.org/en/examples/tutorial.

PR #27, opened 2020-09-12
Backend refactoring

denismerigoux:
The backend code was still buggy and far from the formalization, especially due to a unclear scheme of renaming subscopes. I know @nchataing you thought of it as a clever way of not doing the bookkeeping, but for multiple reasons I decided to get rid of it :
* Subscope renaming was only done one level down, and not completely recursively This is OK with the current formalization but not future-proof if we decide someday to allow programmers to redefine subscopes multiple levels down.
* As I mentioned, far from the formalization where we describe the subscope redefinition bookkeeping.
* The system was based on giving subscope variables a string name like `x::y`. This string name was crucial in the algorithm for name resolution, which I didn't feel very comfortable with.
*  Your `Uid.t` namespace was mixing together local variables (binders), scope variables and subscope variables. I hard a hard time figuring out what was happening by reading the code because of this confusion, which is why I switched to separate namespaces (enforced by a refactoring of `uid.ml`) that "dumbs down" the code.

The PR is still missing the refactoring of the interpreter to build, where I'll have to do the bookkeeping of changing subscope variables with caller prefixes and stuff. My guess is that this bookkeeping is actually not that big and is a small price to pay for a code that's more readable and more future-proof.

I also fix various bugs I think and improve error-handling.


denismerigoux:
Update: the rewrite is complete, I'm merging the PR.

PR #30, opened 2020-10-13
Include catala formalisation

nchataing:


denismerigoux:
Thanks for the PR ! Here's a summary of the list of changes we agreed to in today's meeting. 

## Default logic

* Change the list structure of defaults to a tree following the suggestion by Pierre-Évariste Dagand
* Change the definition of the `++` operator to operator just on two defaults, and create a new merged one 
* Add the beta-reudction rule for default evaluation

## Scope language

*Confirmation : we don't have many reorderings of rules when doing subscope inclusion so we don't actually need the dynamic rewriting operator `Theta`!
* We define the scope language only by showing its encoding in the scope language (as a pure function)
* Figure out the correct way to delay the default evaluation until inside the subscope

```
scope S:
  int x
  int y
  bool b

  x := b ? 1
  x := ~b ? -1

  y := x < 0 ? 42
  y := x > 0 ? 1337

" fun (x',y',b') ->
"   let b := b' ++ [] in
"   let x := x' ++ [b ? 1 | ~b ? -1] in
"   let y := y' ++ [x > 0 ? 42 | x < 0 ? 1337] in
"   (x,y,b)

scope S':
  scope s : S
  int res
  s.b := true
  s.y := s.x >= 0 ? -1
  res := s.y

" fun (res) ->
"   let (s.x,s.y,s.b) = 
"     S{s} (* function S, but if you see s.x, treat it as x, etc. *)
"       ([],[s.x >= 0 ? -1],[true ? true]) in 
"   let res = res ++ [true ? s.y] in
"   (res)
```

PR #45, opened 2020-12-27
Ongoing Catala development

### LINKED ISSUE #36: Add polymorphism into the language
The language does not have any polymorphism right now. Support for polymorphism should be added in the default calculus representation via a classic Big-Lambda type abstraction with type variables. Big lambdas should be explicitly annotated in order to avoid big-lambda inference in the type checking algorithm.

Then, the polymorphism should be reflected upwards in the intermediate representations, and reflected as minimally as possible in the surface syntax. In a first thought, surface syntax should not change at all.

### LINKED ISSUE #37: Add collection types into the language
Once #36 is solved, we can add a primitive polymorphic collection type into the language with its associated operators. This type should be present all the way from the surface syntax to the default calculus.

Once this type is added, the open question of #17 should be reactivated; maybe with a special `map` operator that maps a scope (viewed as a function) onto a collection.

Collections are fixed-length arrays, whose length can be provided at runtime. Their natural extraction target are C arrays.

denismerigoux:
* [x] Add collections and fix #37 
* [ ] Update tutorial


denismerigoux:
I'll merge this here to favor smaller scoped Pull Requests from now on.

PR #56, opened 2021-01-06
Handle extrema operators on collections

### LINKED ISSUE #52: Implement maximum and minimum aggregation function
Catala supports `maximum` and `minimum` aggregation functions for collections. Right now, these are only present in the surface syntax.

One has to fill the holes like this one : 

https://github.com/CatalaLang/catala/blob/933fd024842993f2f458f132e12f480c875b8d38/src/catala/catala_surface/desugaring.ml#L362-L364

To do that, the expression `maximum money for salary in salaries of (salary - $200)` should desugar to 

```
fold (fun acc salary -> let tmp = (salary - 200) in if acc < tmp then tmp else acc) <INIT> salaries
```

The problem lies within the `<INIT>` expression. What to put here? I guess this should be provided by the user, so we have to update the syntax to something like `maximum money context $0 for salary in salaries of (salary - $200)`

nchataing:
Fixes #52.

A possible improvement lies in the desugaring : the `predicate` is calculated twice in the body of the fold function (once for the comparison and once for the result). We could instead put the `if predicate < acc then predicate else acc` under an abstraction to mimick a let-expression inside the scope language.


src/catala/catala_surface/desugaring.ml
```diff
@@ -394,9 +406,17 @@ let rec translate_expr (scope : Scopelang.Ast.ScopeName.t) (ctxt : Name_resoluti
...
-            Errors.raise_spanned_error "Unsupported feature: minimum and maximum"
-              (Pos.get_position op')
+        | Ast.Aggregate (Ast.AggregateExtremum (max_or_min, t, _)) ->
+            let op_kind =
+              match t with
+              | Ast.Integer -> Dcalc.Ast.KInt
+              | Ast.Decimal -> Dcalc.Ast.KRat
+              | Ast.Money -> Dcalc.Ast.KMoney
+              | Ast.Duration -> Dcalc.Ast.KDuration
+              | _ -> assert false
```
denismerigoux, line 416: This should raise an error message like `Errors.raise_spanned_error (Format.asprintf "It does not make sense to get the maximum of values having the %a type" ...) (Pos.get_position op')

denismerigoux (changes requested):
Yes we should avoid code duplication and put the `predicate` under a let-binding. If there's no `Scopelang.Ast.make_let`, then you should create one on the model of 

https://github.com/CatalaLang/catala/blob/3ae9f7b67abb3e1c98604a0a85de670262090d8d/src/catala/default_calculus/ast.ml#L134-L144

denismerigoux (approved):

denismerigoux:
Thanks Nicolas !

PR #57, opened 2021-01-06
Implement quick pattern check desugaring

### LINKED ISSUE #50: Implement quick pattern-check desugaring
The Catala surface syntax contains a handy sugar that allows to check whether a value of a sum type falls into one of the sum type cases:

https://github.com/CatalaLang/catala/blob/933fd024842993f2f458f132e12f480c875b8d38/src/catala/catala_surface/ast.ml#L132

This AST pattern is not handled by desugaring yet: 

https://github.com/CatalaLang/catala/blob/933fd024842993f2f458f132e12f480c875b8d38/src/catala/catala_surface/desugaring.ml#L488

The desugaring should be implemented in the following way. Suppose you have:

```
declaration enumeration Foo:
  -- Case1
  -- Case2
  -- Case3 content integer
```

Then `x with pattern Case2` should be desugared to: 

```
match x with pattern 
  -- Case1 : false 
  -- Case2: true 
  -- Case3 content y: false
```

One can take inspiration from 

https://github.com/CatalaLang/catala/blob/933fd024842993f2f458f132e12f480c875b8d38/src/catala/catala_surface/desugaring.ml#L259-L343

nchataing:
Fixes #50

I'm not sure about which positions we should put in 'ghost' expressions.

### DIFF
--- a/src/catala/catala_surface/desugaring.ml
+++ b/src/catala/catala_surface/desugaring.ml
@@ -341,6 +341,42 @@ let rec translate_expr (scope : Scopelang.Ast.ScopeName.t) (ctxt : Name_resoluti
         (fun e1 cases_d -> (Scopelang.Ast.EMatch (e1, Option.get e_uid, cases_d), pos))
         e1
         (LiftEnumConstructorMap.lift_box cases_d)
+  | TestMatchCase (e1, constructor) ->
+      let possible_c_uids =
+        try Desugared.Ast.IdentMap.find (Pos.unmark constructor) ctxt.constructor_idmap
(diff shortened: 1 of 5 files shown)


denismerigoux (approved):
Thanks! For the ghost `Pos.t` it's your guess, put the closest thing that you have on hand. They're going to be used in error messages later so if they turn out to yield weird error messages we can change them.

You could also put a example in the test suite that gives an error message because the constructor does not belong to the right enumeration.

denismerigoux (approved):

PR #62, opened 2021-01-13
CamlCase -> CamelCase

xbonnetain:
Camel case comes from the shape of a camel, and not from OCaml.

This PR fixes all the occurences of "CamlCase" in the repository to "CamelCase".


denismerigoux:
Merci de ta bonne vigilance typographique :)

PR #68, opened 2021-01-19
add legislative text

jemsab:


jemsab:
Hi, @denismerigoux, quick question : can I split a "déclaration structure" in two or should I define things in a metadata section ? 

In my current work article 4A defines certains properties (tax residence status) for a person and 4B adds elements to that definition (place of residence). I would therefore like to extend the definition as is possible in the Swift  programming language (https://docs.swift.org/swift-book/LanguageGuide/Extensions.html). 

It seems to me be cleaner to keep definitions as close as possible to the legislative text.

Thanks for your advice

denismerigoux:
Hi James, thanks again for this contribution ! 

The proposal you're making is interesting. I had thought of it myself at some point, because it would indeed allows to put structure definitions closer to where they are in the legislative text.

In Catala you currently can't split a structure declaration in two (ore more) pieces. The rationale behind this design is that a Catala structure is morally like a C struct or an OCaml record type : is it not a method. The Swift extensions that you are mentioning are from more object-oriented inspiration. There's a point to be said about keeping the struct declaration in one piece: in a glance, you can see all the fields at your disposal rather than having to collect information through scattered pieces. 

Catala already scatters rules and definitions and makes it harder for the programmer to see all the ways in which one variable is defined. Doing the same for structure and scope declaration would go even further in that direction, and I fear it would hinder the programming experience. The only advantage would be to keep the structure declaration closer to the text ; but recalling my discussions with Liane and Sarah they said that lawyers are not interested in looking at structure declarations or scope declarations, they only care about the rules. Moreover, the legislative text does not contain much things that correspond to structure declarations, as most data structures are implicit in the text and have to be inferred from context.

That's why the current method I propose is to create a `prologue.catala_fr` file that you include at the beginning of your master file 
and that contains all the declarations. Doing that allows the programmer to quickly see how data is organized and functions connect to each other. 

For all these reasons, I do not particularly favor adding to Catala the possibility of splitting struct or scope declarations. But that could change if we see more compelling arguments for it.

jemsab:
Duly noted, thanks for such a quick reply

denismerigoux:
Closing for #146

PR #70, opened 2021-01-21
Add syntactic sugar for unlabeled exceptions. Fixes #59

### LINKED ISSUE #59: Syntactic sugar for one base case and exceptions
Right now, we have a label system to declare exceptions to rules : 

https://github.com/CatalaLang/catala/blob/839310d5ba685c68ebb6fe8fbb106df64da9d589/examples/tutorial_en/tutorial_en.catala_en#L355-L362

But in many cases, when defining variable `x` of scope `A`, you only have one base case and several exceptions to it. To make the syntax easier to read in this case, we could have a syntactic sugar of the form: 

```
definition income_tax equals two_brackets.tax_formula of individual.income  
  
# Then, you can declare the exception by referring back to the label 
exception definition income_tax under condition  
individual.income <=$ $10,000 
consequence equals $0 
 ```
 
 The desugaring pass could make sure that there's only one `definition` without the `exception` tag, making it the base case. If there are multiple `definition` without `exception`, then the desugaring pass throws an error message saying you should revert to a label to disambiguate.

This involves modyfing `src/catala/catala_surface/parser.mly` as well as `src/catala/catala_surface/desugaring.ml`

### PR DESCRIPTION
**Pull request by R1kM**
This PR adds support for unlabeled exceptions, when there is no ambiguity.
This is done using syntactic sugar in the Catala surface language.

The parser and surface AST are changed to support unlabeled exceptions: exceptions are now an ident option option instead of an ident option, where Some None represents an unlabeled exception.

The desugaring of a scope use is enhanced as follows:
- We first collect all definitions with unlabeled exceptions as a set of ScopeDef (function gather_unnamed_exceptions)
- For each unlabeled exception, we attribute a fresh rulename to the corresponding default definition (function generate_default_rulenames). At this stage, we raise an error if there is an ambiguity, i.e. if there are several possible default definitions, or labeled definitions for this scopeDef. The fresh rulename is stored in a map, which is now part of the scope_use context.
- When generating a rule name inside process_def, we lookup previously generated names for default definitions. If the definition is not a default definition, or is not referred to by any unlabeled exception, the name generation proceeds as prior to this PR
- When looking up the parent rule inside process_def for an unlabeled exception, we fetch the previously generated rulename using the rulename map inside of the scope_use context. If no default definition exists for this definition, we raise an error at this stage.

This PR also defines get_def_key as an auxiliary function, since it's now used in several places.
It also provides several tests for this new feature in test_exception.

### CODE REVIEW

**Comment by denismerigoux** [comment]
Thanks a lot Aymeric! This should fix #59 :)

File: src/catala/catala_surface/ast.ml [hunk 1]
```diff
@@ -218,7 +218,7 @@ and expression =
 
 type rule = {
   rule_label : ident Pos.marked option;
-  rule_exception_to : ident Pos.marked option;
+  rule_exception_to : ident Pos.marked option option;
```
denismerigoux, line 221: Indeed, we need an `option option` here but we might as well make it a proper sum type with informative variants like

```OCaml
type exception_to = 
  | NotAnException
  | ExceptionToLabel of ident Pos.marked 
  | ExceptionToImplicitBaseCase
```

src/catala/catala_surface/ast.ml
```diff
@@ -229,7 +229,7 @@ type rule = {
 
 type definition = {
   definition_label : ident Pos.marked option;
-  definition_exception_to : ident Pos.marked option;
+  definition_exception_to : ident Pos.marked option option;
```
denismerigoux, line 232: Same as above

src/catala/catala_surface/desugaring.ml
```diff
@@ -747,29 +747,35 @@ let process_default (ctxt : Name_resolution.context) (scope : Scopelang.Ast.Scop
     exception_to_rule;
   }
 
+let get_def_key (def : Ast.definition)
```
denismerigoux, line 750: Thanks for the refactoring!

src/catala/catala_surface/desugaring.ml
```diff
@@ -893,6 +920,104 @@ let process_scope_use_item (precond : Ast.expression Pos.marked option)
...
+  (rulemap : Desugared.Ast.RuleName.t Desugared.Ast.ScopeDefMap.t) 
+  (item : Ast.scope_use_item Pos.marked)
+  : Desugared.Ast.RuleName.t Desugared.Ast.ScopeDefMap.t =
+  let is_exception (def : Ast.definition) = match def.Ast.definition_exception_to with
+    | None -> false | _ -> true
+  in
+  match Pos.unmark item with
+  | Ast.Rule rule ->
+    let consequence_expr = Ast.Literal (Ast.LBool (Pos.unmark rule.rule_consequence)) in
+    let def =
```
denismerigoux, line 969: This might be refactored into `rule_to_def` since it's used in quite a number of places?

tests/test_exception/ambiguous_unlabeled_exception.catala.A.out
```diff
@@ -0,0 +1,6 @@
+[ERROR] An unlabeled exception refers to several definitions, including this one. Disambiguate by using labels.
```
denismerigoux, line 1: Very neat error message, I like it!

src/catala/catala_surface/desugaring.ml
```diff
@@ -804,10 +810,31 @@ let process_def (precond : Scopelang.Ast.expr Pos.marked Bindlib.box option)
                   def.definition_name
             | Some label -> label )
     in
+    let is_exception (def : Ast.definition) = match def.Ast.definition_exception_to with
+      | None -> false | _ -> true
+    in
+    (* If we had previously defined a rulename for this default definition during
```
denismerigoux, line 816: OK, I understand what happens.

denismerigoux (approved):
Thanks again for this superb PR. I left a few comments that might help keep the code even prettier. 

Also: do you want to go through the examples (allocations familiales, US Tax code) and simplify the code using this new feature? Else it will be a pleasure for me to do it :)

src/catala/catala_surface/ast.ml
R1kM: I modified the datatype and the code to use this sum type

src/catala/catala_surface/ast.ml
R1kM: Same as above

src/catala/catala_surface/desugaring.ml
R1kM: Good point, I just pushed the refactoring

R1kM:
Thanks for the comments, I modified the code and the PR accordingly.

I'd be happy to propagate this to the examples. Should I maybe do it in a separate PR, once this one is merged?

denismerigoux:
Yep, let's do it in a separate PR.

PR #71, opened 2021-01-22
Improve unlabeled exceptions, and use them in US tax code and allocations familiales examples

R1kM:
The previous support for unlabeled exceptions was limited to definitions and exceptions defined inside of the same code block.
The file tests/test_exception/split_unlabeled_exception.catala shows an example that was previously failing, although it should be considered a valid use of unlabeled exceptions. Examples in the US tax code and in allocations familiales, although larger, follow a similar pattern. Thus using unlabeled exceptions was not possible there.

This PR enables support for unlabeled exceptions in the same scope across code blocks.
The main difference is that the generation of default rulenames for default definitions is done during the name resolution pass, instead of during the desugaring pass. This allows to have a global context with rulenames for default definitions already generated, instead of generating and forgetting them after each iteration over a scope use during desugaring.

For each scope, default_rulemap is now a map to a new datatype, unique_name, which, for a given scope definition, differentiates between 
* The absence of a default definition (no associated value)
* Ambiguous default definitions (either several default definitions possible, or a labeled definition)
* A Unique default definition, for which a fresh rulename has been generated

During the desugaring phase, it is then sufficient to check that every unlabeled exception is associated to a unique default definition.

Using this improvement, this PR also simplifies the existing examples to use unlabeled exceptions when possible.

### DIFF
--- a/examples/allocations_familiales/decrets_divers.catala_fr
+++ b/examples/allocations_familiales/decrets_divers.catala_fr
@@ -74,14 +74,14 @@ tranche est celle dont les revenus sont supérieurs au plafond de base de
 
 /*
 champ d'application AllocationsFamiliales :
-  exception base_plafond_I_d521_3
+  exception
   définition plafond_I_d521_3 sous condition 
     date_courante >=@ |01/01/2020| et date_courante <=@ |31/12/2020|
   conséquence égal à 57 759 € +€
     5 775 € *€ (entier_vers_décimal de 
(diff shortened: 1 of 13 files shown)


denismerigoux (approved):
Well spotted! I didn't realize this bug while reviewing your last PR, but indeed this exception analysis should have gone into the main name resolution pass instead of the desugaring. 

Thanks for having changed the examples ; as I thought this syntactic sugar removes a lot of syntax noise for the programmer.

PR #72, opened 2021-01-24
Better structure for the testing framework

R1kM:
This PR proposes a different structure for the testing framework.

For each folder, tests are split in two different folders, depending on whether they are good or bad tests. Their outputs are stored in a different folder, called output.
This enables an easier browsing of the unit tests when trying to understand Catala features


denismerigoux:
Thanks Aymeric for yet another great PR! I see you've kept the Makefile system for detecting and running tests, it's fine by me. Just a little comment though: what happens if there are two tests with the same name in the `good` and `bad` folders of a subdirectory ?

In this case, I would expect at least an error message coming from the Makefile and saying: "please rename your test". Alternatively, you could move the `outputs` folders into `good` and `bad`: doing that, you can have the same test name in `good` and `bad` but there would be no ambiguity as to which output file corresponds to it.

R1kM:
That's a very good question. At the moment, this will lead to an error when calling the catala compiler. Let's assume your duplicated file is called simple.catala. The current system will try to call `$(CATALA) good/simple.catala bad/simple.catala`, and the compiler will complain about being passed two files. The error message will not be as explicit as "please rename your test", but it should hopefully give sufficient information to debug.

Do you think that having the outputs folder in the good and bad directory would be a better solution? If so, I'm happy to make the change.

denismerigoux:
I think I would prefer having `outputs` inside `good` and `bad`. Thanks for the comment!

R1kM:
Thanks for the feedback. I modified the infrastructure accordingly, and moved the test files to the appropriate folders

denismerigoux (approved):
Looks good to me, thanks for implementing my suggestion!

PR #75, opened 2021-01-26
Disambiguation syntax for fields and constructors. Fixes #74

### LINKED ISSUE #74: Add disambiguation syntax for enum constructors and struct fields
## The problem

Currently, it is possible in Catala for structs to share field names and for enums to share constructor names. E.g:

```
declaration structure Foo:
  data field1 content int 
  data field2 content bool

declaration structure Bar:
  data field1 content date 
  data field3 content bool
```

But then, field access becomes ambiguous:

```
... baz.field1 # Is it Foo.field1 or Bar.field1 ?
```

Right now, an error is thrown in this case:

https://github.com/CatalaLang/catala/blob/5973c677cf1f500e9030ffcf934f56c0384c8413/src/catala/catala_surface/desugaring.ml#L58-L84

https://github.com/CatalaLang/catala/blob/5973c677cf1f500e9030ffcf934f56c0384c8413/src/catala/catala_surface/desugaring.ml#L232-L243

However, it is not currently syntactically possible to disambiguate using `Foo.field1` or `Bar.field1` ; currently, the only solution is to rename the fields.

## How to fix it ?

First, the parser has to be updated here:

https://github.com/CatalaLang/catala/blob/5973c677cf1f500e9030ffcf934f56c0384c8413/src/catala/catala_surface/parser.mly#L115-L117

`ident` should be turned into something like `option(constructor DOT) ident` and the AST should be updated accordingly:

https://github.com/CatalaLang/catala/blob/master/src/catala/catala_surface/ast.ml#L266

Then, the desugaring should be updated. I don't known at this point whether we would need to update this part

https://github.com/CatalaLang/catala/blob/5973c677cf1f500e9030ffcf934f56c0384c8413/src/catala/catala_surface/desugaring.ml#L208-L245

R1kM:
This PR adds disambiguation syntax for struct fields and enum constructors.

If `x` is a struct of type `Foo` containing a field `f`, the syntax `x.Foo.f` is now allowed to access field `f`.
This enables disambiguation when several structs contain the field f.

This PR also adds a few good and bad tests for this new syntax.


denismerigoux:
Thanks again Aymeric! While in you're in that issue streak mode, do you think you can extend this PR for ambiguous enum constructor names? The same problem appears with things like:

```
declaration enum Foo:
  -- Case1
  -- Case2

declaration enum Bar:
  -- Case1
  -- Case3
  
 ... match x with pattern -- Case1 : ... # To which enum does Case1 refer to ?
 # -> match x with pattern -- Foo.Case1: ...
```

denismerigoux:
Also : which `ocamlformat` version are you using? Don't forget to `make build` before you commit so that the code that you write is autoformatted. The project uses `ocamlformat 0.15.0`, which is the version that's compatible with all the opam dependencies of the project for OCaml v4.09.1.

R1kM:
Thanks for the comments,

Regarding enums, very good point, I initially thought that Dotted in the grammar was covering them.
I extended EnumInject and the match_case_patterns in the AST (and parser) to support possibly qualified enum names. Did I forget another node?
See the test test_enum/good/disambiguated_cases.catala which recaps the new possible behvaiours.

Regarding `ocamlformat`, I'm using the right version, but had forgotten to run `make build`. I was only running `make` from the tests/ folder. This should now be fixed, sorry about that

denismerigoux (approved):
Looks good to me, thanks!

PR #76, opened 2021-02-12
Add .mlis for all of Catala

AltGr:
Some important notes:

- this is on top of the compiling_defaults branch, so should wait for
  it to be merged

- it also includes a renaming of the directories below `src/catala`,
  so that they match the corresponding library names

- it's a first pass: in particular, I did almost no filtering in the
  published interfaces (even when there was an 'API' header towards
  the end; decided it was easier to generate everything first). It
  should be easy to go through the .mlis and just clear everything
  that we don't want shared anymore now, testing that it still
  compiles along the way.

- as a corollary, I decided to keep OCamlDoc comments in both
  interface and implem for now, so that the cleanup doesn't lose
  information. We might want to strip duplicate comments from .ml
  files later on, e.g. if they start diverging...

- you may want to squash the commits before merge, because there is a
  bit of noise in the middle (huge generated mlis that are cleaned in
  a subsequent commit)


denismerigoux:
Wouahou the compilation speedup is actually huge, at least x2 or x3!

denismerigoux:
I agree with all your remarks; I'll probably find some time to curate the `.mli` this week-end. I would prefer to do that before merging this PR.

denismerigoux (approved):

PR #82, opened 2021-02-24
For EN syntax highlighting, use a pygments plugin rather than a fork …

protz:
Pygments has a plugin mechanism that allows easily registering new languages without having to override a complete pygments setup. This is convenient, e.g. if pygments is managed by your distribution or package manager, and you wish to keep things modular.

This replaces the install script by a simple `python setup develop`, which sets up symlinks so that pygments picks up on any improvement to the lexer directly.


denismerigoux:
Thanks Jonathan! I was not aware of this plugin mechanism, but it sounds like the right solution to adopt. Before merging, I'll need to extend this to FR and NV syntax highlighting, and fix `literate/latex.ml` that was previously redefining the location of the `pygmentize` executable in generated `.tex` files.

denismerigoux:
Ok I fixed everything to use the new setup, thanks again @protz !

PR #111, opened 2021-04-30
Remove parser tokens for builtin functions

### LINKED ISSUE #66: Remove parser tokens for builtin functions
Right now, builtin functions have a parser token associated with them:

https://github.com/CatalaLang/catala/blob/7a835ac5f654e42beb82bf25f570606136584c54/src/catala/catala_surface/parser.mly#L147-L158

Altough this allows supporting multi-language frontends out of the box, it is cumbersome since it requires touching the parser each time we want to add a new builtin.

Instead, we should add a piece of code in the `qident` expression constructor: 

https://github.com/CatalaLang/catala/blob/7a835ac5f654e42beb82bf25f570606136584c54/src/catala/catala_surface/parser.mly#L107

This piece of code would check would lookup the parser identifier into a multi-language list of identifiers corresponding to builtin functions, that would sit in the `lexer*.ml` files. Careful, one also has to check that we can't define scope parameters, structure fields or bound variables with a name corresponding to a `builtin`.

AltGr:
Closes #66


AltGr:
Quite a bit of functorisation here :sweat_smile: 

Note that I used first-class modules here (to get the `lang -> lexer` mapping) but if you don't like them we could easily do without, by instanciating the different languages and just selecting the main parsing function based on the `lang` variant.

~Marked as draft because there is no mechanism yet to prevent the use of builtin names as variables.~
Added check for builtin names as variables in the last patch, but not 100% happy with this way of failing in the parser.

denismerigoux (changes requested):
Thanks very much @AltGr for this PR! 

I would prefer not to use first-class modules here, as you say we can select the main parsing function by casing on the language. It is the style I have used for various other localized items throughout the codebase.

I think the way you fail in the parser on builtins is fine ; if you want you can create a helper function in https://github.com/CatalaLang/catala/blob/master/src/catala/surface/parse_utils.ml to de-clutter the parser code.

I expect builtin addition to be a frequent thing so could you add a to-do list in the [CONTRIBUTING.md](https://github.com/CatalaLang/catala/blob/master/CONTRIBUTING.md) that explains where to add things in the compiler code base to add a new builtin ?

This is really going to make things way easier so thanks again :)

AltGr:
Sure thing ; I know everyone is not fond of first-class modules, hence why I asked :)

AltGr:
Update:
- removed first-class modules
- added a small guide in `CONTRIBUTING.md`. **Note**: the built-ins still get encoded separately in the dcalc ast, we could further streamline them by relying on a common `Builtin` variant rather than specific unary operators.
- removed the interface for the surface AST, because it's not maintainable due to the generated visitors. The `.ml` file only contains type definitions (explicitely) anyway.

denismerigoux (approved):
Perfect !

denismerigoux:
I would prefer to keep the frontend builtins separated from the backend operators because we might want to add a builtin that is just a sugar for multiple backend operators :)

PR #122, opened 2021-05-17
POC: Using cmake instead of Makefiles

wozniakpl:
Hi all,

What do you think about adding using cmake in this project? It'd be an easier way to maintain buildsystem, dependencies between targets, etc.

1st commit is an example of how `test_examples` would work. Oneliner:
```
mkdir -p build/ && cmake -B build && cmake --build build/ --target test_examples
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ _opam/
 src/**/.merlin
 legifrance_oauth*
 *.html
+build/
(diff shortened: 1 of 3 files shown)


denismerigoux:
Hi Bartosz, thanks for this contribution ! However the main contributors of this project (my PhD advisor @protz and I) are not at all familiar with CMake, and instead are quite fluent with regular `make`. I do agree with you that the current build system could use some improvement, but a potential change to CMake would make it difficult for me to maintain it.

denismerigoux:
Closing the issue since no more activity on this.

PR #126, opened 2021-05-29
fix(compiler): correct the handling of the Division_by_zero exception

EmileRolley:
An invalid index was used to get the null denomitator.

#### Changelog

* Fix and factorize the exception handling in `Dcalc.interpreter`.
* Add some unit tests for `Division_by_zero` and `UncomparableDurations` exceptions.
* Fix the `.run` rule of the `./tests/Makefile` in order to manage all catala file extensions.


denismerigoux (approved):
Thanks for this sweet refactoring !

Issue #105, opened 2021-04-28
Feature request: wildcard in patterns


protz:
It would be convenient to be able to do:

```
match e₁ with pattern
  -- * : e₂
```

the syntax can also be an underscore, although maybe a star will be more understandable ?

denismerigoux:
The location to change to add this feature is
 
https://github.com/CatalaLang/catala/blob/2cad9823f74ebc2fb13dd90a061e9bb4fbcb4da1/src/catala/surface/desugaring.ml#L394-L400

EmileRolley:
> It would be convenient to be able to do:
> 
> ```
> match e₁ with pattern
>   -- * : e₂
> ```
> 
> the syntax can also be an underscore, although maybe a star will be more understandable ?

It could be the keyword `any` (respectively `n'importe quel` for the french syntax).

denismerigoux:
Actually we should ask lawyers about this keyword too :)

EmileRolley:
To be sure, the wildcard should match all cases or all remaning cases?

For example, with this declarations:

```
declaration enumeration E:
  -- Case1 content integer
  -- Case2
  -- Case3

declaration scope A:
  context x content E
  context y content integer
```

Should this code be considered correct?
```
scope A:
  definition x equals Case1 content 2
  definition y equals match x with pattern
    -- Case1 of i : 42
    -- anything : 43
```
And equivalent to:
```
scope A:
  definition x equals Case1 content 2
  definition y equals match x with pattern
    -- Case1 of i : 42
    -- Case2 : 43
    -- Case3 : 43
```

Or should it only be used alone?

```
scope A:
  definition x equals Case1 content 2
  definition y equals match x with pattern
    -- anything : 42
```

denismerigoux:
The wildcard should match remaining cases, as in nearly all languages with pattern matching.

denismerigoux:
As the  pattern matching in `Scopelang`, `Dcalc` and `LCalc` don't support it, you can either add support for wildcard pattern matching in all of those languages, or you can do a workaround by let-binding the wildcard arm expression just before the match and putting the reference to it in all the branches the wildcard stands for.

denismerigoux:
This latter option should not change the semantics because there are no side-effects when evaluating expression that don't have defaults in Catala, but ideally we should add wildcard pattern matching support all the way down.

EmileRolley:
Closed by #129

PR #129, opened 2021-06-03
feat(compiler)!: add wildcard in patterns

EmileRolley:
> Related issue #105 

### Questions

~~1.  I need your help for the choice of the polish keyword @wozniakpl.~~
~~2.  English and french keywords are starting with uppercases in order to match with constuctors. Is it a good idea?~~
~~3.  In order to reduce changes in syntax highlighting files, the french keyword is `N'importe` instead of `N'importe quel`. Is it ok?~~

### Improvement

As @denismerigoux mentionned in #105, there is multiple ways to add this features. With this PR I chose the second one, which consits of *let-binding the wildcard arm expression just before the match and putting the reference to it in all the branches the wildcard stands for*.

However, the wildcard support should be added all the way down and not just in the desugaring pass, so I suggest to open a new issue for this improvement.

### Changelog

* Add support for wildcard in patterns, including:
  * new tokens
  * new surface AST constructor `WildCard` and type `explicit_match_case`
  * modification of the `Surface.Desugaring.disambiguate_match_and_build_expression` function
  * new unit test files
  * refactor lexers in order to manage at least one space between words in multiwords keyword.


wozniakpl:
Hi,

Regarding the polish word for "anything" - "cokolwiek" should be ok.

EmileRolley:
> Hi,
> 
> Regarding the polish word for "anything" - "cokolwiek" should be ok.

Perfect, thanks!

denismerigoux:
> English and french keywords are starting with uppercases in order to match with constructors. Is it a good idea?

I would prefer to stick with keywords starting with lowercase, both because of convention + you might want to call one of your own constructors `Anything`.

> In order to reduce changes in syntax highlighting files, the french keyword is N'importe instead of N'importe quel. Is it ok?

Actually you can keep the syntax highlighting to only `n'importe` since `quel` is already matched as a regular keyword, but I would prefer `n'importe quel` to be the full keyword parser by the compiler.

EmileRolley:
> Actually you can keep the syntax highlighting to only `n'importe` since `quel` is already matched as a regular keyword, but I would prefer `n'importe quel` to be the full keyword parser by the compiler.

By the way, multiple words keywords are lexed with exactly one space between words, is it suitable?

denismerigoux:
It is what we do with https://github.com/CatalaLang/catala/blob/9b16576c5a8d81bf8ac4b92aecfdec6e4ec8f61e/src/catala/surface/lexer_fr.ml#L116-L118. But actually, you can put a `Star white_space` between the two words to be more flexible :)

EmileRolley:
If you think that is preferable, I will refactor lexers with `Plus white_space` then.

src/catala/surface/desugaring.ml
```diff
@@ -728,9 +728,28 @@ and disambiguate_match_and_build_expression (scope : Scopelang.Ast.ScopeName.t)
...
-  let expr, e_name =
-    List.fold_left
-      (fun (cases_d, e_uid) (case, _pos_case) ->
+  let create_var = function
+    | None -> (ctxt, (Scopelang.Ast.Var.make ("_", Pos.no_pos), Pos.no_pos))
+    | Some param ->
+        let ctxt, param_var = Name_resolution.add_def_local_var ctxt param in
+        (ctxt, (param_var, Pos.get_position param))
+  in
+  let bind_case_body c_uid e_uid ctxt param_pos case_body e_binder =
```
denismerigoux, line 737: I would prefer for this kind of functions that the type of the parameters be annotated, as it really improves the readability of the code.

tests/test_enum/bad/output/useless_wildcard.catala_en.A.out
```diff
@@ -0,0 +1,7 @@
+[ERROR] Unreachable match case, all constructors of the enumeration E are already specified.
```
denismerigoux, line 1: Ok so for that one you can actually transform it into a warning, since it does not prevent you from compiling the code :)

denismerigoux (approved):
Thanks very much @EmileRolley ! This is the first PR where you start to dive more deeply into the compiler, reaching now the desugaring phase :) I hope you found the journey interesting, don't hesitate to ask more questions, and please know that your contributions are very valued.

denismerigoux:
Oh BTW when this is merged you can open a new issue about how to add support for pattern match wildcarding in the later intermediate representations of the compiler.

EmileRolley:
> Thanks very much @EmileRolley ! This is the first PR where you start to dive more deeply into the compiler, reaching now the desugaring phase :) I hope you found the journey interesting, don't hesitate to ask more questions, and please know that your contributions are very valued.

You are welcome! It was certainly more complicated than the previous ones but the satisfaction is even greater :innocent:

EmileRolley:
> Oh BTW when this is merged you can open a new issue about how to add support for pattern match wildcarding in the later intermediate representations of the compiler.

Of course!

denismerigoux (approved):

Issue #115, opened 2021-05-05
Feature: annotate exceptions with location information


msprotz:
Working with the JS backend, I get this sort of exceptions:

```
1: Object { t: 9, c: "Runtime.NoValueProvided", l: 23 }
```

if there was a way to annotate the `NoValueProvided` with a location, or better, the actual variable for which no value was provided, it'd be really great!

denismerigoux:
Thanks Jonathan for raising this ! To give `NoValueProvided` a location, the relevant places in the code to change are 

https://github.com/CatalaLang/catala/blob/8ab73627935583495603933f68d43ac6c3f10e0e/src/catala/lcalc/to_ocaml.ml#L229

https://github.com/CatalaLang/catala/blob/8ab73627935583495603933f68d43ac6c3f10e0e/src/catala/runtime.mli#L39

The position should be embedded in the generated code with 

https://github.com/CatalaLang/catala/blob/8ab73627935583495603933f68d43ac6c3f10e0e/src/catala/runtime.mli#L76-L83

in the manner of

https://github.com/CatalaLang/catala/blob/8ab73627935583495603933f68d43ac6c3f10e0e/src/catala/lcalc/to_ocaml.ml#L332-L337

EmileRolley:
Could you give me an example of workflow to get this exception please? @msprotz

denismerigoux:
Fixed by #132

PR #132, opened 2021-06-08
Feat #115: annotate NoValueProvided exn with location informations

EmileRolley:
> Related issue : #115 

## Questions

~~1. Is it what was expected?~~
~~2. Is there a quick way to test it?~~
~~3. I don't really know what to do with the [`Lcalc.Print`](https://github.com/CatalaLang/catala/compare/master...EmileRolley:feat-localized-exn?expand=1#diff-2447a342020cc426d632dfeb5704590bc1223bbb8b37474ea5acbb6fc9cce49aR66) module.~~
~~4. How to update all benchmarking and test files?~~

## Changelog

* Extend the `Runtime.NoValueProvided` exception with location informations.
* Add `.ml` as target for the `examples/Makefile.common.mk` `clean` rule.
* Update some `french_law` files.


denismerigoux:
Hi Emile, thanks for tackling this issue ! Let me answer your questions:

> 1. Is it what was expected?

So here you have to distinguish between the deep embedding `NoValueProvided` exception of `lcalc/ast.mli` (which corresponds to the `ErrorOnEmpty` unary operator of `dcalc/ast.mli`) and the shallow embedding `NoValueProvided` exception of `runtime.ml`. This first one (deep embedding) is internal to the Catala compiler, and is used only in the `dcalc` interpreter. You don't need to attach a position to it because the Catala interpreter already retrieves the position by simply looking the position tag attached to the expression to which `ErrorOnEmpty` is applied to : 

https://github.com/CatalaLang/catala/blob/c718095b2ddcf788e4f886645b8576f5ee667d2e/src/catala/dcalc/interpreter.ml#L403-L410

Issue #115 really concern the second, shallow embedded `NoValueProvided` of `runtime.ml`, which is going to be used in OCaml programs compiled from Catala sources. And this you have correctly dealt with your changes in `lcalc/to_ocaml.ml` notably :)

> 2. Is there a quick way to test it?

We try to avoid the `NoValueProvided` and `ErrorOnEmpty` triggering in our current legislative examples because it is a sign that a definition is missing to handle a certain situation present in reality but not modeled in the law. So you will have to build an example of your own, which can look really simple like

```
declaration scope Foo:
  context x content boolean
```

When you compile this program down to OCaml and run it with any output, since `x` is never defined in function `Foo`, it will always throw a `NoValueProvided` error.

> 3. I don't really know what to do with the Lcalc.Print module.

You don't need to modify it, see answer to 1. :)

> 4. How to update all benchmarking and test files?

The benchmarkings and test files written in OCaml depend on the `runtime.ml` file, so when you update the definition of the `NoValueProvided` exception in `runtime.ml` with a source position payload, the OCaml compiler will tell you where to fix the benchmarks and test files that will have some typing errors.

Hope it clarifies things for you :)

EmileRolley:
Okey, thanks a lot!

EmileRolley:
> We try to avoid the `NoValueProvided` and `ErrorOnEmpty` triggering in our current legislative examples because it is a sign that a definition is missing to handle a certain situation present in reality but not modeled in the law. So you will have to build an example of your own, which can look really simple like
> 
> ```
> declaration scope Foo:
>   context x content boolean
> ```
> 
> When you compile this program down to OCaml and run it with any output, since `x` is never defined in function `Foo`, it will always throw a `NoValueProvided` error.

I compiled this example and add at the end:

```ocaml
let _ =
  let foo_out = foo { x_in = (fun _ -> false) } in
  Printf.printf "x_out = %b\n" foo_out.x_out
```

But no error is thrown... Am I missing something?

denismerigoux:
Did you put `content boolean` or `condition` ?

Try with 

```
declaration scope Foo:
  context x content integer
```

Conditions are `false` by default, hence no error if you put one. But if you had put `content boolean` that's a bug, I should investigate.

EmileRolley:
> Did you put `content boolean` or `condition` ?
> 
> Try with
> 
> ```
> declaration scope Foo:
>   context x content integer
> ```
> 
> Conditions are `false` by default, hence no error if you put one. But if you had put `content boolean` that's a bug, I should investigate.

I already tried with `integer` and no errors too.

I think it's related to the last line of this generated function:

```ocaml
let foo (foo_in : foo_in) =
  let x_ : unit -> integer = foo_in.x_in in
  let x_ : integer =
    try
      handle_default
        [| (fun (_ : _) -> x_ ()) |]
        (fun (_ : _) -> true)
        (fun (_ : _) ->
          handle_default [||]
            (fun (_ : _) -> false)
            (fun (_ : _) -> raise EmptyError))
    with EmptyError ->
      raise
        (NoValueProvided
           {
             filename = "tmp.catala_en";
             start_line = 5;
             start_column = 11;
             end_line = 5;
             end_column = 12;
             law_headings = [ "Test file" ];
           })
  in
  { x_out = x_ }
```

denismerigoux:
You have to put 

```
let _ =
  let foo_out = foo { x_in = (fun _ -> raise EmptyError) } in
  Printf.printf "x_out = %b\n" foo_out.x_out
````
For the program to raise the error. With `false` you give a value for `x`, so it does not throw `NoValueProvided`.

EmileRolley:
Indeed, this works.

denismerigoux (approved):

PR #133, opened 2021-06-09
Fix atom fr syntax highlighting plugin

PhiSeb:


denismerigoux:
Thanks for spotting the missing escapement!

Issue #109, opened 2021-04-28
List concatenation


protz:
The `@` operator is missing for list concatenation... this would be helpful for combining results from multiple sub-scopes, wherein the sub-scopes can return a list of, say, penalties

denismerigoux:
We should indeed add this. Not sure if `@` is the right syntax though, let's ask the lawyers what they think.

EmileRolley:
> We should indeed add this. Not sure if `@` is the right syntax though, let's ask the lawyers what they think.

Any feedback from the lawyers about the desired syntax?

denismerigoux:
Not yet, you can try and ask Liane or Lilya on the french law Zulip channel.

denismerigoux:
Closed by #137

PR #137, opened 2021-07-08
Feat #109: add collection concatenation operator

EmileRolley:
> Related issue: #109 

## Questions

~~1. Currently, the interpreter does not evaluate expresions inside collections during the concatenation. Is it ok?~~

https://github.com/CatalaLang/catala/blob/731513a003d4f38e2e2e9974b0719b3740265027/compiler/dcalc/interpreter.ml#L191-L192

## Changelog

* New operator `++` for collection concatenation.
* Minor refactoring of the formatting functions inside `Print` modules.


compiler/dcalc/interpreter.ml
```diff
@@ -188,6 +188,8 @@ let rec evaluate_operator (ctx : Ast.decl_ctx) (op : A.operator Pos.marked)
         match Pos.unmark (evaluate_operator ctx (Pos.same_pos_as (A.Binop A.Eq) op) args) with
         | A.ELit (A.LBool b) -> A.ELit (A.LBool (not b))
         | _ -> assert false (*should not happen *))
+    | A.Binop A.Concat, [ A.EArray es; A.EArray es2 ] ->
+        A.EArray (es @ es2) (* @EmileRolley NOTE: should it keep lazy evaluated or not? *)
```
denismerigoux, line 192: Keep it like that. `es` and `es2` already have been evaluated at this point because concat is viewed as a function, and when you call a function you always evaluate first the arguments.

src/french_law/law_source/unit_tests/tests_allocations_familiales.ml
```diff
@@ -0,0 +1,4123 @@
+(** This file has been generated by the Catala compiler, do not edit! *)
```
denismerigoux, line 1: This should not appear in the diff, it is in the `.gitignore` of its folder.

denismerigoux (approved):
Looks good to me except the `allocations_familiales_test.ml` which should not be under version control.

compiler/dcalc/interpreter.ml
EmileRolley: Ok I see. I will remove the comment.

denismerigoux (approved):

PR #138, opened 2021-08-18
Cleanup the lexer, and refactor for more generic directives

AltGr:


AltGr:
Now I can add the `mod_use` directives :sweat_smile:

AltGr:
Wait for merge I have more fixes ^^

AltGr:
Should be better now. Be careful I force-pushed

compiler/surface/lexer_common.ml
```diff
@@ -35,17 +35,19 @@ let get_law_heading (lexbuf : lexbuf) : token =
...
 (** Boolean reference, used by the lexer as the mutable state to distinguish whether it is lexing
     code or law. *)
-let is_code : bool ref = ref false
+let context : lexing_context ref = ref Law
 
 (** Mutable string reference that accumulates the string representation of the body of code being
     lexed. This string representation is used in the literate programming backends to faithfully
     capture the spacing pattern of the original program *)
-let code_string_acc : string ref = ref ""
+let code_buffer : Buffer.t = Buffer.create 4000
```
denismerigoux, line 47: I guess it's more efficient like that :)

compiler/surface/lexer_common.ml
```diff
@@ -35,17 +35,19 @@ let get_law_heading (lexbuf : lexbuf) : token =
   let precedence = calc_precedence (String.trim (get_substring 1)) in
   LAW_HEADING (title, article_id, article_expiration_date, precedence)
 
+type lexing_context = Law | Code | Directive | Directive_args
```
denismerigoux, line 38: I like this

compiler/surface/lexer_en.ml
```diff
@@ -500,60 +503,73 @@ let rec lex_code (lexbuf : lexbuf) : token =
...
+        L.get_law_heading lexbuf
+    | _ -> (
+        (* Nested match for lower priority; `_` matches length 0 so we effectively retry the
+           sub-match at the same point *)
+        let lexbuf = lexbuf in
+        (* workaround sedlex bug, see https://github.com/ocaml-community/sedlex/issues/12 *)
+        match%sedlex lexbuf with
+        | Star (Compl '\n'), ('\n' | eof) -> LAW_TEXT (Utf8.lexeme lexbuf)
+        | _ -> L.raise_lexer_error (Pos.from_lpos prev_pos) prev_lexeme)
+  else
```
denismerigoux, line 562: Ok neat, the lexer for law is now aware of beginning of lines. Good ! @EmileRolley we talked about this at some point

compiler/surface/parser.mly
```diff
@@ -29,6 +29,10 @@ end>
 
 %start source_file
 
+(* The token is returned for every line of law text, make them right-associative
+   so that we concat them efficiently as much as possible. *)
+%right LAW_TEXT
```
denismerigoux, line 34: Wouahou neat, that's far beyond my level of parser expertise

denismerigoux (approved):
Thanks a lot Louis ! I agree with you, it's difficult to keep the lexer code in sync as we add more languages. What do you think could be functorized or factorized more that currently ? The problem is that with `sedlex` we can't really break the lexing match into pieces...

AltGr:
Yes, it won't be trivial to factorise... The only idea that comes to mind at the moment would be some kind of custom pre-processing where we have a lot of named placeholders in the lexer code, then a mapping for these placeholders to syntax fragments (sedlex regexps, code, strings...) in separate files  for each language.

That sounds like heavy machinery though; I wonder if existing ppx macro systems could be re-used...

denismerigoux:
I think we'd have to develop our own ppx for that. I'm merging this anyway :)

AltGr:
I don't know, maybe `cppo` would be enough to do the job with a big list of `#define` for each language :thinking:

PR #139, opened 2021-08-19
Factorise lexer translations

AltGr:
@denismerigoux since you asked nicely ;)


compiler/surface/lexer_fr.cppo.ml
```diff
@@ -0,0 +1,141 @@
...
+        Runtime.integer_of_string (R.substitute ~rex:remove_spaces ~subst:(fun _ -> "") units) \
+      in \
+      let cents = \
+        try Runtime.integer_of_string (parts 4) with Not_found -> Runtime.integer_of_int 0 \
+      in \
+      L.update_acc lexbuf; \
+      MONEY_AMOUNT (units, cents)
+
+#define MX_DECIMAL_LITERAL \
+   Plus digit, ',', Star digit -> \
```
AltGr, line 135: Hm actually we could define a macro just for the decimal separator... couldn't we ?

denismerigoux: No, in English the decimal separator is `.` instead of `,` in France.

denismerigoux (approved):
Wouahou this is amazing! Is `cppo` bundled with dune or do we have to add it a an opam dependency ? I love this PR, this will really enable scaling up with many many frontend languages. Do you think you could add a doc section about how to add a new language with this setup here https://github.com/CatalaLang/catala/blob/master/CONTRIBUTING.md#internationalization ?

AltGr:
Cppo isn't bundled, but the dependency was already pulled through `visitors`; I'll add it explicitely.

:+1:  for the doc

compiler/surface/lexer_fr.cppo.ml
AltGr: yes, what I mean is that we could factorise the digit parsing and have a macro only for `.` or `,`

denismerigoux:
BTW: what do the suffixes `MR`, `MS` and `MX` mean ?

AltGr:
It's explained here: https://github.com/CatalaLang/catala/pull/139/files#diff-60a8e1ddb93b0c4c97283d6b44f8f8f9b853b6044f8fba5f92ff27f57d8c67c9R24-R28

catala.opam
```diff
@@ -32,6 +32,7 @@ depends: [
   "js_of_ocaml-ppx" {>= "3.8.0"}
   "camomile" {>= "1.0.2"}
   "odoc" {with-doc}
+  "cppo" {>= "1"}
```
denismerigoux, line 34: The `catala.opam` is generated from `dune-project`, so you should modify this one. There's also a list of `opam` packages to fetch in the `dependencies-ocaml` rule of the top-level `Makefile`.

denismerigoux (approved):

catala.opam
AltGr: Ah yes, just noticed this about the dune file, sorry !

Also, I have patches to avoid the need for listing the dependencies again in the Make file (using `opam install . --deps-only` instead) but haven't cleaned up for pushing yet :)

PR #160, opened 2021-11-28
add menhirLib to Nix dependencies

jakeisnt:
I was having some trouble building Catala with Nix - it was missing the dependency `menhirLib` - so `menhirLib` has been added to the  `default.nix` file here as a dependency of catala. The nix build should be fixed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,13 +12,23 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  build-nix:
+  build-nix-21-05:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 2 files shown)


denismerigoux:
@lIlIlIlIIIIlIIIllIIlIllIIllIII could you review this PR ?

lIlIlIlIIIIlIIIllIIlIllIIllIII:
Thanks for your PR.

I think the issue is due to [menhir split into 3 packages](https://gitlab.inria.fr/fpottier/menhir/-/commit/423432d28163e22baf846f64bc489a5f3a6aa039) last year. The current version in the 21.05 channel is 20190626 (pre-split), whereas the current version on the unstable channel is 20211012 (post-split).

I've added a fix compatible with both the 21.05 and the unstable channel. I've added an CI check to ensure Catala compiles against both channels.

Issue #166, opened 2021-12-10
Issue with ocamldoc


denismerigoux:
The documentation does not build anymore, leaving weird error message like this:

```
dune build @doc
File "dcalc__Ast.cmti":
Couldn't find the following modules:
  Stdlib
File "scopelang__Ast.cmti":
Couldn't find the following modules:
  Stdlib
File "scopelang__Dependency.cmti":
Couldn't find the following modules:
  Graph Stdlib
File "desugared__Ast.cmti":
Couldn't find the following modules:
  Stdlib
File "lcalc__Ast.cmti":
Couldn't find the following modules:
  Stdlib
File "desugared__Dependency.cmti":
Couldn't find the following modules:
  Graph
File "surface__Parser.cmti":
Couldn't find the following modules:
  MenhirLib
File "../../compiler/desugared/.desugared.objs/byte/desugared.odoc":
Couldn't find the following modules:
  Graph Stdlib
File "../../compiler/desugared/.desugared.objs/byte/desugared.odoc":
Failed to lookup type unresolvedroot(Desugared).Ast.RuleMap.t Parent_module: Parent_sig: Unresolved module path r(hidden((root Desugared__Ast))) (Lookup failure (module): (root Desugared__Ast))
File "../../compiler/lcalc/.lcalc.objs/byte/lcalc.odoc":
Couldn't find the following modules:
  Stdlib
File "../../compiler/.runtime.objs/byte/runtime.odoc":
Couldn't find the following modules:
  Stdlib
File "../../compiler/scalc/.scalc.objs/byte/scalc.odoc":
Couldn't find the following modules:
  Stdlib
File "../../compiler/literate/.literate.objs/byte/literate.odoc":
Couldn't find the following modules:
  Stdlib
File "../../compiler/utils/.utils.objs/byte/utils.odoc":
Couldn't find the following modules:
  ANSITerminal Cmdliner Stdlib
File "../../../../compiler/scopelang/scopelang.mld", line 7, characters 43-77:
Failed to resolve reference unresolvedroot(Dcalc).Ast.EDefault Couldn't find "Ast"
File "../../../../compiler/scalc/scalc.mld", line 28, characters 0-27:
Failed to resolve reference unresolvedroot(Lcalc).To_python Couldn't find "To_python"
File "../../../../compiler/lcalc/lcalc.mld", line 29, characters 0-57:
Failed to resolve reference unresolvedroot(Lcalc).To_python Couldn't find "To_python"
[...]

PR #169, opened 2022-01-02
Fix(build/doc): remove warnings due to .ml* files

EmileRolley:
> related issue: #166 
> related scopes: `docs`

### Changelog

* Modifies documentation references to remove compilation warnings.

### Note

> "Part of it seems related to the cppo lexer extension.", said _@denismerigoux in  https://github.com/CatalaLang/catala/issues/166#issue-1077051102_

I don't thinks so, indeed, warnings about unlocatable modules such as `Stdlib` appears when upgrading `odoc` to a version `>= 2.0.0` and this even before the refactoring of the lexers.

**Quickfix**: a solution to remove this warnings is to use `odoc.1.5.3`.

### DIFF
--- a/compiler/desugared/desugared.mld
+++ b/compiler/desugared/desugared.mld
@@ -1,9 +1,9 @@
 {0 Desugared representation }
 
-This representation is the second in the compilation chain 
-(see {{: index.html#architecture} Architecture}). Its main difference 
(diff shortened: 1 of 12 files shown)


denismerigoux:
Thanks for having looked into that! Closes #166.

Issue #112, opened 2021-04-30
Feature: scope context variables decorators (input, output, internal)


denismerigoux:
## The problem

Scopes in Catala can have many context variables. But as the number of context variables grows, it is more and more difficult to figure out which of these variables are output, which are input and which ones are intermediate variables that are not relevant from outside the scope. Catala users have already started using comments to annotate scope context variable declarations with this classification.

While it is the essence of Catala that context variables are neither input nor outputs by default, since they can be redefined by a calling scope, we could benefit from user annotations to enable helper lints and better code generation in the different backends.

## Specification of the decorations

A regular context variable declaration looks like this: 

```
scope Foo:
  context a content bool 
```

This proposal would allow replacing the `context` with the following keywords:

* `input`: this scope variable be defined by the caller, cannot be defined in the scope
* `output`: this scope variable cannot be defined by the caller, has to be defined in the scope
* `internal`: this scope variable cannot be defined by the caller, has to be defined in the scope, and does not appear in the outputs
* `output`: this scope variable can defined by the caller, can be defined in the scope, and appears in the outputs

The classifications `internal`/`input`/`context` on the one hand, and `internal`/`output` on the other hand, form two independent classifications for respectively the input and output behavior of a scope variables. From this independent combinations of two choices between respectively 3 and 2 options are yielded 6 different possibilities for fully qualifying the input/output behavior of a scope variable:

1. internal
2. output
3. input
4. input output
5. context
6. context output

This specification defines  informally two permissiveness lattices between the kinds of scope variables. Here it is, the most permissive being at the top:

```
CONTEXT
        | 
        |  
    INPUT                        OUTPUT
        |                                    |
        |                                    |
 INTERNAL                   INTERNAL
```

## Linting

If we have these four keywords, we can enforce their specification in three different ways.

1. ERROR LINT: When calling a subscope `Foo`, we can ensure that all the variables of `Foo` redefined in the caller are either `context` or `input`, but not `internal`.
2. ERROR LINT: When calling a subscope `Foo`, we can ensure that all the variables of `Foo` used (as outputs) in defining variables of the caller are `output`, but not `internal`
3. WARNING LINT: Inside a scope, we can ensure all variables defined are either `context` or `internal`
4. WARNING LINT: Inside a scope, we can check that all variables marked as `output` or `internal` have at least one definition.

## Code generation 

[...]

PR #175, opened 2022-01-08
Catala as a proof platform

denismerigoux:
## Roadmap

This branch contains the implementation related to the ProLaLa 2022 submission https://hal.archives-ouvertes.fr/hal-03447072/. 

## Main features

This PR brings an new proof functionality to the compiler : it now generates verification conditions tied to the well-behaved execution of each scope definition, and tries to solve those verification conditions using Z3.

## Other changes

* Optimizations brought to Dcalc
* Complete documentation at the OCamldoc level and CLI level
* Test suite for the prover

## TODO

* [x] Separate `z3encoding.ml` into the encoding itself and the logic for launching and executing queries which should be proof-backend generic (functor ?)
* [x] Doc for CLI and error message interpretation


denismerigoux:
Note: we'll probably want #112 implemented to distinguish variables that don't have definitions because they're input vs. those where there's a bug.

compiler/verification/z3encoding.ml
```diff
@@ -0,0 +1,632 @@
...
+        let accessor = List.hd accessors in
+        let proj = Expr.mk_app ctx.ctx_z3 accessor [ head ] in
+        (* The fresh variable should be substituted by a projection into the enum in the body, we
+           add this to the context *)
+        let ctx = add_z3matchsubst fresh_v proj ctx in
+
+        let body = Bindlib.msubst (Pos.unmark e) [| fresh_e |] in
+        translate_expr ctx body
+    (* TODO: Denis, Is this true for constructors with no payload? *)
+    | _ -> failwith "[Z3 encoding] : Arms branches inside VCs should be lambdas"
```
denismerigoux, line 426: Right now all constructors have a payload. The "no payload" case is encoded with a unit payload and yes there is always a lambda.

denismerigoux:
@lIlIlIlIIIIlIIIllIIlIllIIllIII we need to fix the unstable nix build, I'm assigning you to it :)

denismerigoux:
I think this PR is in a pretty decent state to be merged. There are things left to do but they can be dealt with later, this PR is already very big. @R1kM @lIlIlIlIIIIlIIIllIIlIllIIllIII please file issues documenting what remains to be done here (encoding of decimals into Z3 for instance).

PR #177, opened 2022-01-10
Clerk, a build system for Catala

denismerigoux:
## The problem

Catala currently has no build system. This is unfortunate as a build system is a necessity for a modern programming language to handle tasks such as: dependency management, compilation, testing, benchmarking. Right now, these things are managed through Makefiles, which is not ideal as some have pointed out (#122). 

Particularly, the test harness is done through arcane Makefile magic and only supports comparing the output of the interpreter for a given test file. We would want to compare the output of the Catala compiler for different backends on the same file, to cross-test the features of the compiler

## The solution

Enter Clerk, the new build system for Catala. Its interface and philosophy ought to be heavily inspired by `cargo`. This PR is the first to introduce clerk by replicating the basic functionalities of the test harness that already existed. More features will be coming as we hope at some point to have `clerk build` and `clerk test` :)

## Documentation

See `make help_clerk` for the Clerk man page. Clerk is implemented in OCaml, like the Catala compiler. Testing doc has been updated accordingly.


denismerigoux:
Clerk in action:
Capture d’écran du 2022-01-11 12-52-11
Capture d’écran du 2022-01-11 12-52-37

PR #178, opened 2022-01-12
An attempt to simplify deps install and CI

AltGr:
from the docs, the setup-ocaml@v2 github action should already handle
caching...


denismerigoux:
So even if the caching is still not working, I'm going to merge this PR @AltGr if it's okay with you because it brings a needed cleaning of the current CI system. I'll open an issue for the CI caching.

PR #184, opened 2022-01-26
Restore CI by disabling counterexamples printing

denismerigoux:
Recently, the CI failed because Z3 returned different counterexamples from the ones that were enshrined in the expected stdout output. We cannot afford to have the CI failing because of Z3's randomness, so I added an option in the Catala compiler to disable Z3's models and avoid printing the values for the test suite.

Here was an example of failing CI: https://github.com/CatalaLang/catala/runs/4952713428?check_suite_focus=true#step:6:611


R1kM (approved):
It unfortunately seems like the OCaml bindings do not expose facilities to change/set a Z3 seed. These changes look fine to me, please feel free to merge.

Issue #102, opened 2021-03-25
Eliminating spurious errors related to "dependent computations"


msprotz:
Based on a long discussion with @denismerigoux. Here's a Catala example:

```
definition bar under condition paragraph_a_applies consequence equals $0
definition foo under condition paragraph_a_applies consequence equals bar
definition foo under condition paragraph_b_applies consequence equals $1
```

The only output that I care about is the value of `foo`. However, if I call this scope with `paragraph_b_applies = true`, then I get a fatal error because the definition of `bar` is empty, *even though I do not need `bar` for computing my output `foo`*.

The reason is that this gets translated to:

```
let bar = < | paragraph_a_applies: $0 > in
let foo = < < | paragraph_a_applies: bar >; < | paragraph_b_applies: $1 > | > in
bar, foo
```

and the evaluation of `bar` is unconditional

If, however, I were able to annotate `foo` with an attribute that says "this is an output of my top-level scope", then the program above would contain more opportunities for rewriting; for instance, a simple syntactic analysis could uncover that `bar` is only needed in the first computation, and the program could be rewritten as such:

```
let foo = < < | paragraph_a_applies: let bar = < | paragraph_a_applies: $0 > in bar >; < | paragraph_b_applies: $1 > | > in
foo
```

This would provide a superior user experience; rather than providing an un-needed default case for `bar` when `paragraph_b_applies` (which legally does not make sense), we ensure instead that the compiler "figures out" that `bar` is needed only for the computation of `foo` under condition `paragraph_a_applies` and rewrite the default calculus term accordingly.

Questions to work out:
- how to express this rewriting formally
- can we always guarantee that this optimization triggers and therefore eliminates *all* spurious empty errors?
- what is the corresponding standard compiler transformation, and what can we learn from existing implementations?

AltGr:
I only now stumbled upon this old issue, but similar questions keep arising regularly.

At the risk of stating the obvious, inlining or lazy evaluation come to mind, although in a restricted setting. But the actual implementation triggered another idea: what if we were to remove the `ErrorOnEmpty` guarding all (non-output) variables, and move them to the positions where the variables are being used ? Semantics wise, that will probably be more difficult to express (additional wrapping, or another kind of exceptions ?) but it might be worth exploring.

denismerigoux:
Fixed by #970

Issue #89, opened 2021-03-10
Externally-implemented modules for Catala


denismerigoux:
## The problem

In this issue, we assume that Catala has a module system as described in #88. The problem we want to solve is exemplified by #81: we want for Catala programs to rely on abstract scopes and types which cannot be defined or implemented within the limits of the DSL. 

### Syntax

To simplify the implementation, we only allow for whole modules to be externally implemented at a time. Externally implemented modules have to be define in standard `.catala*` files, except that the first line with the module name should bear:

```markdown
> Module Foo external
```

While types and scopes have to be declared in such a file, it is impossible to write any scope usage that might introduce rules or definitions. Modules that use `Foo` can then use any of the externally implemented types and scopes defined in `Foo`.

### Compilation 

When compiling an externally implemented module to a backend like OCaml, the Catala compiler shall create a header file like `*.mli` or `*.h`. It is then up to the user to provide a manually-written implementation in the target language.

### Intererpreter 

> See below

msprotz:
Filling out the "Interpreter" part after the design discussion for this issue.

There are essentially two steps to this process. The first one is to i) go over the types introduced by the external `.catala*` file, then ii) go over the externally-written definitions so as to "make them available" to the Catala interpreter located within the Catala compiler.

The second step is to embed/de-embed interpreter values and pass them to / receive them from externally-implemented definitions.

### External Catala modules written in OCaml, via the interpreter

A first way to go about this is to rely on compiler-libs to perform on-the-fly compilation of the external OCaml files to bytecode, then use the resulting compiled functions from the interpreter. In greater detail:
1. Catala processes `foo.catala_en` which is an external module
2. Catala emits `foo_types.mli` which contains only type declarations (although maybe standalone mlis are deprecated, I can't remember precisely)
3. Catala locates `foo_impl.ml` written by the user
4. Catala starts an OCaml interpreter via compiler-libs, processes `foo_types.mli` then `foo_impl.ml`, then the user-written functions become available via the OCaml interpreter's environment and can be called from Catala's interpreter

### External Catala modules written in OCaml, via compilation and dynlink

To avoid relying on an unstable API such as compiler-libs, one can use instead the regular ocaml compiler... which results in a somewhat heavier process.

Steps 1.-3. remain the same. Then:
4. Catala compiles `foo_types.mli` to `foo_types.cmi`; `foo_impl.ml` to`foo_impl.cmxs` using `-shared`
[...]

PR #200, opened 2022-02-17
Implementation of Section 1001 of the US Tax Code

denismerigoux:


ammkrn:
@denismerigoux 

After trying my hand at this, I'm curious how you want to handle the pattern of exceptional subsections requiring additional or specialized information. If the scope declaration/function signature only has `input` fields that require information sufficient to handle the default case, you may not be able to handle the exceptions or determine whether a given condition is met. On the other hand, if you ask for all of the possible information you might need for the whole section, some signatures will get quite large; without a way to express relationships between arguments large sets of arguments will likely be sources of errors. 

The default rule for 1001 just needs to know the fair market value of property received, and the amount of money received. However, (b)(2) is an exception that wants to know about real property taxes under 164, and (e) is an exception that wants to know about a potential term interest to determine whether the exception applies. If 1001 is done as a single scope, the inputs would need to be something like

```
fmv_property_received content money
money_received content money
property_tax_reimbursed_164 content option<money>
property_tax_paid_by_purchaser_164 content option<money>
term_of_years content option<some struct or enum>
```

Not having `option` and requiring a `$0` input is against the grain for functional programming, but may still be what you want to do (which is fine). Alternatively, you could break up the subsections into their own scopes that take the correct set of inputs, though this means that the caller would have to know the correct subsection. This might not be so bad; it seems like the different type signatures would provide some safety, and the actual statutory implementation seems to represent something more like a private API than a public one. Taxpayers aren't going to be looking at the US Code when determining their tax liability, and in the same vein as our discussion about the 1040, something like a public API implemented on top of the statutory stuff that directly addresses the needs of practitioners might be in line with the layered architecture theme.

You could also create an enum that serves as the argument, though that does move closer to "just use a programming language", and wouldn't handle cases in which there's overlap (if a transaction has characteristics of exception 1 and exception 2).

Lastly, is there a means of calling external resources? Certain parts of the code require retrieval of a value from something like an actuarial or depreciation table (these are supplied by the government) or adjustments for inflation which presumably would need to be part of the ocaml codebase. For test cases, the most common ones are probably the [marginal rate schedule](https://www.irs.gov/newsroom/irs-provides-tax-inflation-adjustments-for-tax-year-2021), and [MACRS depreciation tables](https://www.irs.gov/publications/p946#en_US_2020_publink1000270861).

denismerigoux:
Hi Chris, let me answer your questions.

> After trying my hand at this, I'm curious how you want to handle the pattern of exceptional subsections requiring additional or specialized information. If the scope declaration/function signature only has input fields that require information sufficient to handle the default case, you may not be able to handle the exceptions or determine whether a given condition is met. On the other hand, if you ask for all of the possible information you might need for the whole section, some signatures will get quite large; without a way to express relationships between arguments large sets of arguments will likely be sources of errors.

This is a relevant remark, that shows that you have quite some experience translating law into code and a deep understanding of it. What you are talking about is a known problem, related to https://github.com/CatalaLang/catala/issues/102. There are several mitigation strategies: first, mark your scope's input as `context` instead of `input`, and provide a default definition for those inputs inside the scope when such a default definition makes sense. Second, give more structure to your inputs by using enumerations that capture the relationship between all the pieces of the input. Ultimately, you would want a completely dependently typed language to capture accurately real-world relationships between the inputs in complex pieces of law; unfortunately this would add A LOT of complexity to Catala and we don't plan to turn Catala into a dependently-typed language :)

> The default rule for 1001 just needs to know the fair market value of property received, and the amount of money received. However, (b)(2) is an exception that wants to know about real property taxes under 164, and (e) is an exception that wants to know about a potential term interest to determine whether the exception applies. If 1001 is done as a single scope, the inputs would need to be something like

>```
>fmv_property_received content money
>money_received content money
>property_tax_reimbursed_164 content option<money>
>property_tax_paid_by_purchaser_164 content option<money>
>term_of_years content option<some struct or enum>
>```

> Not having option and requiring a $0 input is against the grain for functional programming, but may still be what you want to do (which is fine). [The 1015 example has a match on Some/None](https://github.com/CatalaLang/catala/blob/c4a7fe36c1bbe69fb0716b77bd7e5eac28d324c8/examples/us_tax_code/section_1015.catala_en#L74), but the file doesn't type check, and option doesn't appear in the syntax cheat sheet, so maybe this was a feature that got axed? Alternatively, you could break up the subsections into their own scopes that take the correct set of inputs, though this means that the caller would have to know the correct subsection. This might not be so bad; it seems like the different type signatures would provide some safety, and the actual statutory implementation seems to represent something more like a private API than a public one. Taxpayers aren't going to be looking at the US Code when determining their tax liability, and in the same vein as our discussion about the 1040, something like a public API implemented on top of the statutory stuff that directly addresses the needs of practitioners might be in line with the layered architecture theme.

> You could also create an enum that serves as the argument, though that does move closer to "just use a programming language", and wouldn't handle cases in which there's overlap (if a transaction has characteristics of exception 1 and exception 2).

I guess you thought of the exact same mitigations strategies as me and brilliantly described them on the example at hand :)

> Lastly, is there a means of calling external resources? Certain parts of the code require retrieval of a value from something like an actuarial or depreciation table (these are supplied by the government) or adjustments for inflation which presumably would need to be part of the ocaml codebase. For test cases, the most common ones are probably the [marginal rate schedule](https://www.irs.gov/newsroom/irs-provides-tax-inflation-adjustments-for-tax-year-2021), and [MACRS depreciation tables](https://www.irs.gov/publications/p946#en_US_2020_publink1000270861).

Not yet, but support for this planned in https://github.com/CatalaLang/catala/issues/89.

denismerigoux:
And yeah the Section 1015 example is an old prototype that we did not retrofit to the current version of Catala, we should create an issue for that.

ammkrn:
I pushed a relevant commit on [this branch](https://github.com/ammkrn/catala/commits/chris_1001)(the first is just a typo fix). After reworking this a couple of times and thinking a lot about how this will intersect with users and other code sections, I have much more articulable positions on some of the issues that were brought up earlier. I'll write them down over the weekend and maybe we can discuss them if there's a meeting next week.

denismerigoux:
Thanks @ammkrn! I think you can move your progress on 1011 into this branch as well, so that we'll merge 1001 and 1011 together.

denismerigoux:
Closing this as it has become stale.

PR #205, opened 2022-02-19
Add support for rationals in Z3 encoding

R1kM:
This PR extends the current Z3 proof backend with support for rationals, encoded as Z3 reals. It also adds a few unit tests for this support.

### DIFF
--- a/compiler/verification/z3backend.ml
+++ b/compiler/verification/z3backend.ml
@@ -115,7 +115,7 @@ let rec print_z3model_expr (ctx : context) (ty : typ Pos.marked) (e : Expr.expr)
        pretty-printing, we print nothing to remain closer from Catala sources *)
     | TUnit -> ""
     | TInt -> Expr.to_string e
-    | TRat -> failwith "[Z3 model]: Pretty-printing of rational literals not supported"
+    | TRat -> Arithmetic.Real.to_decimal_string e 7
     (* TODO: Print the right money symbol according to language *)
     | TMoney ->
         let z3_str = Expr.to_string e in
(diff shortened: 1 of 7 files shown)


compiler/verification/z3backend.ml
```diff
@@ -213,7 +213,7 @@ let translate_typ_lit (ctx : context) (t : typ_lit) : Sort.sort =
   | TBool -> Boolean.mk_sort ctx.ctx_z3
   | TUnit -> fst ctx.ctx_z3unit
   | TInt -> Arithmetic.Integer.mk_sort ctx.ctx_z3
-  | TRat -> failwith "[Z3 encoding] TRat type not supported"
+  | TRat -> Arithmetic.Real.mk_sort ctx.ctx_z3
```
denismerigoux, line 216: Are we sure we want to use Z3's real ? Didn't we agree we would use fixed-precision integers ?

denismerigoux:
I guet the following output for running the `allocations_familiales` example: 

```
[ERROR] The translation to Z3 failed:
[Z3 encoding] application of non-integer binary operator Mult not supported
[ERROR] The translation to Z3 failed:
[Z3 encoding] application of non-integer binary operator Mult not supported
```

and 

```
[ERROR] The translation to Z3 failed:
[Z3 Encoding] Ill-formed VC, a function application does not have a function type
[ERROR] The translation to Z3 failed:
[Z3 Encoding] Ill-formed VC, a function application does not have a function type
```

R1kM:
Thanks for the review. I tried for a while using fixed-precision integers, but it seemed to lead to unexpected errors and losses of precision.
In the end, since we do not use quantifiers, Z3's real theory seems to work reasonably well.
I'll debug the two issues you observed. The former is likely that operations on dates or money are partially supported, I can probably fix it easily.
I'm a bit more surprised by the second one, I'll have a look. Marking this PR as draft in the meantime.

R1kM:
I fixed the first issue, I hadn't added arithmetic operations on money when adding money support.
For the second issue, the problem stems from some functions in VCs having type TAny. We currently do not support those, as we do not know what Z3 sort to give them. I improved the error message to indicate that this is not actually an ill-formed VC.
Note that, interestingly, adding the -O option on the `allocations_familiales` example avoids this case.

compiler/verification/z3backend.ml
```diff
@@ -115,7 +115,7 @@ let rec print_z3model_expr (ctx : context) (ty : typ Pos.marked) (e : Expr.expr)
        pretty-printing, we print nothing to remain closer from Catala sources *)
     | TUnit -> ""
     | TInt -> Expr.to_string e
-    | TRat -> failwith "[Z3 model]: Pretty-printing of rational literals not supported"
+    | TRat -> Arithmetic.Real.to_decimal_string e 7
```
denismerigoux, line 118: Instead of `7` you should use the compiler option that is precisely here for this kind of things: https://github.com/CatalaLang/catala/blob/12b31554d673ce905b9cd31bfe89f7b63682a7bd/compiler/utils/cli.ml#L31

denismerigoux (approved):
Looks good to me, you can merge after changing the precision digits thing!

PR #213, opened 2022-03-04
Draft: add backend plugin support

AltGr:


compiler/utils/cli.ml
```diff
@@ -80,19 +80,61 @@ let backend =
...
+let plugins_dirs =
+  let doc = "Set the given directory to be searched for backend plugins." in
+  let env = Term.env_info "CATALA_PLUGINS" ~doc in
+  let default =
+    let ( / ) = Filename.concat in
+    [ Sys.executable_name / ".." / "lib" / "catala" / "plugins" ]
+  in
+  Arg.(value & opt_all dir default & info [ "plugin-dir" ] ~docv:"DIR" ~env ~doc)
+
+type backend_option_builtin =
```
EmileRolley, line 92: Hi @AltGr, are there any advantages to using polymorphic variants rather than ordinary ones for this case?

AltGr: Yes, see the [individual commit messages](https://github.com/CatalaLang/catala/pull/213/commits/a212d529ad2d91f5617148e4744f2f8ee0ddc4fe):
- the last part is optional, I can roll it back if preferred: it makes use of
  polymorphic variants, which some find distasteful. `driver.ml` proceeds with
  successive refinements of the `Cli.backend_option` type, and in this case they
  fit well by allowing to type these refinements, ensuring both the absence of
  holes and redundance, while removing the need for pervasive `_ -> assert
  false` cases.

I can happily either remove this part (and restore the `assert false` cases) or move it to a separate PR for further discussion, if desired.

EmileRolley: > and in this case they fit well by allowing to type these refinements, ensuring both the absence of holes and redundance, while removing the need for pervasive _ -> assert false cases.

I understand the point, thanks!

AltGr:
What should I do with this ? I propose the following:
- remove patch [9da75f7](https://github.com/CatalaLang/catala/pull/213/commits/9da75f7d63a194cd9af0c1ab667ce5996edc6705) which contains just demo plugins (esp. the `python` one is of no interest since the backend is already built-in !) ; another PR to re-add the jsoo plugin after a little cleanup should be fine
- rebase the rest and merge ? In particular [a212d52](https://github.com/CatalaLang/catala/pull/213/commits/a212d529ad2d91f5617148e4744f2f8ee0ddc4fe) has some refactorings that are already conflicting and will get worse as time goes
- if more debate is needed on plugins, maybe rebase [a212d52](https://github.com/CatalaLang/catala/pull/213/commits/a212d529ad2d91f5617148e4744f2f8ee0ddc4fe) as a separate PR. I'll in any case have to extend the `debug_printf` functions for consistency across the various printing functions ; nothing difficult but it should be done and merged quickly to avoid losing time on conflict resolution :)

AltGr:
Rebased ; the first patch should be fine, the second adding the demo plugins needs some cleanup (removing the dummy python plugin, decide where/how we put the jsoo one, add it to the normal build path…)

AltGr:
Rebased, I think it should be OK now, and the two plugins are properly advertised as examples.

denismerigoux (approved):
It looks good to me code-wise but I think this plugin system is lacking some high-level documentation. 

1. Can you define a `plugin.mld` file in `compiler/plugins` ? Would this file be showed in the odoc output ?
2. Can you write in `CONTRIBUTING.md` a guide on how to define a new backend plugin and make it work ?

AltGr:
I'll do that, thanks!

PR #215, opened 2022-03-07
Fix(clerk): sort subdirectory names

EmileRolley:
Fix the issue encountered by @denismerigoux in https://github.com/CatalaLang/catala/pull/186#pullrequestreview-901484377, by sorting directory names after a `Sys.readdir` call to have determined results from `Ninja_utils.format`.

## Changelog

* Add a new function `readdir_sort`.
* Modify the failing test.


denismerigoux (approved):
Thanks!

PR #218, opened 2022-03-08
Use format strings directly in debug/error/log functions

AltGr:
This avoids many intermediate calls to e.g. `Format.asprintf`; should result
in some cases in "more correct" use of `Format`¹, avoid the computation of
unused debug strings, and make the code more readable.

¹ for `Format` to work as expected, all intermediate calls need to go through 
it. Some cases of formatting to an intermediate string then printing through
Format again are still present, but this makes the situation better.


AltGr:
(note: this should be reviewed and merged if fit with high priority due to the high potential for merge conflicts :) )

compiler/dcalc/interpreter.ml
```diff
@@ -212,48 +211,44 @@ let rec evaluate_operator (ctx : Ast.decl_ctx) (op : A.operator Pos.marked)
...
-                       let expr_str =
-                         Format.asprintf "%a" (Print.format_expr ctx ~debug:false) (e', Pos.no_pos)
-                       in
-                       let expr_str =
-                         Re.Pcre.substitute ~rex:(Re.Pcre.regexp "\n\\s*")
-                           ~subst:(fun _ -> " ")
-                           expr_str
-                       in
-                       Cli.print_with_style [ ANSITerminal.green ] "%s" expr_str))
+              (* FIXME: this usage of Format is broken, Formatting requires that all is formatted in
```
denismerigoux, line 214: I've used `TODO` accross the codebase rather than `FIXME`, I guess we can stick with one keyword only :)

denismerigoux (approved):
Thank you so much! it pained me indeed to have to use `Format.asprintf` all the time....

PR #221, opened 2022-03-08
Command-line: use a record for the options

AltGr:
Should make it much easier and less error-prone to add new options. There is 
still a bit of boiler-plate, but at least it's contained in the Cli.options 
function and doesn't transpire in the interfaces.


denismerigoux (approved):
Thanks so much!!!! I had wished I'd done that a long time ago.

PR #222, opened 2022-03-08
Refactor: group common functions related to backend_option in the Cli module

EmileRolley:
Simply group `catala_backend_option_to_string` and `catala_backend_option_of_string` functions in `utils/cli.ml`.

### DIFF
--- a/compiler/lcalc/ast.ml
+++ b/compiler/lcalc/ast.ml
@@ -121,7 +121,7 @@ let make_none (pos : Pos.t) : expr Pos.marked Bindlib.box =
 let make_some (e : expr Pos.marked Bindlib.box) : expr Pos.marked Bindlib.box =
   let pos = Pos.get_position @@ Bindlib.unbox e in
   let mark : 'a -> 'a Pos.marked = Pos.mark pos in
-  let+ e = e in
+  let+ e in
   mark @@ EInj (e, 1, option_enum, [ (D.TLit D.TUnit, pos); (D.TAny, pos) ])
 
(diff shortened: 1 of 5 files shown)


denismerigoux (approved):

compiler/lcalc/ast.ml
```diff
@@ -134,7 +134,7 @@ let make_matchopt_with_abs_arms
   let pos = Pos.get_position @@ Bindlib.unbox arg in
   let mark : 'a -> 'a Pos.marked = Pos.mark pos in
 
-  let+ arg = arg and+ e_none = e_none and+ e_some = e_some in
```
denismerigoux, line 137: Yea you can't really remove these, it's a monadic bind it does not have the same syntactic sugar as regular binds.

denismerigoux: I just updated `ocamlformat`, we're using `0.20.1` now. On `master` when I run `make format` it doesn't change anything and the `let+ arg = arg` are conserved.

AltGr:
This will conflict with #221 (there is some overlap between what they do!), if it's ok I think #221 can go first, then I can take care of rebasing the remaining improvements from this one.

EmileRolley:
Of course!

denismerigoux (approved):
Thanks Emile!! Once it's rebased/merged on top of `master`, let's go :)

AltGr:
Ah actually it does't conflict! I had in mind another local patch that I didn't put #221 yet. So feel free to merge and I'll handle the rebase of my second part of the patch :)

Issue #207, opened 2022-02-21
Restore JS executable of the Catala compiler


denismerigoux:
## The problem

Right now, if you compile the Catala interpreter with `js_of_ocaml` to a Javascript binary with

    dune build compiler/catala.bc.js --profile release

You will get the following errors:

```
Dummy implementations (raising 'Failure' exception) will be used if they are not available at runtime.
You can prevent the generation of dummy implementations with the commandline option '--disable genprim'
Missing primitives:
  caml_mutex_new
  caml_thread_initialize
  n_add_rec_def
  n_ast_map_contains
  n_ast_map_erase
  n_ast_map_find
  n_ast_map_insert
  n_ast_map_keys
  n_ast_map_reset
  n_ast_map_size
  n_ast_map_to_string
  n_ast_to_string
  n_ast_vector_get
  n_ast_vector_push
  n_ast_vector_resize
  n_ast_vector_set
  n_ast_vector_size
  n_ast_vector_to_string
  n_ast_vector_translate
  n_context_of_ast
  n_context_of_ast_map
  n_context_of_ast_vector
  n_context_of_func_entry
  n_context_of_func_interp
  n_context_of_model
  n_context_of_param_descrs
  n_context_of_params
  n_context_of_solver
  n_context_of_symbol
  n_del_config
  n_func_decl_to_string
  n_func_entry_get_arg
  n_func_entry_get_num_args
  n_func_entry_get_value
  n_func_interp_get_else
  n_func_interp_get_entry
  n_func_interp_get_num_entries
  n_get_app_arg
  n_get_app_decl
  n_get_app_num_args
  n_get_arity
  n_get_as_array_func_decl
  n_get_ast_hash
  n_get_ast_id
  n_get_ast_kind
  n_get_datatype_sort_constructor
  n_get_datatype_sort_constructor_accessor
  n_get_datatype_sort_num_constructors
  n_get_datatype_sort_recognizer
  n_get_decl_ast_parameter
  n_get_decl_double_parameter
  n_get_decl_func_decl_parameter
  n_get_decl_int_parameter
  n_get_decl_kind
  n_get_decl_name
  n_get_decl_num_parameters
  n_get_decl_parameter_kind
  n_get_decl_rational_parameter
  n_get_decl_sort_parameter
  n_get_decl_symbol_parameter
  n_get_domain
  n_get_domain_size
  n_get_full_version
  n_get_func_decl_id
  n_get_range
  n_get_sort
  n_get_sort_id
[...]

PR #223, opened 2022-03-08
Make Z3 an optional dependency

AltGr:
If Catala is compiled without Z3, trying to run it with the backend `Proof`
will yield:
```
[ERROR] This instance of Catala was compiled without Z3 support.
```
and return 124

Note that this doesn't change the `make depends`, opam file or CI to account
for it, it just enables it at the build-system level.

There are also no hooks at this moment to have Catala self-document the
options whith which it was compiled (e.g. in the `--help` screen). But that
could be added in a more general way later, it's probably not really needed
yet.


denismerigoux (changes requested):
Amazing! Altough I would prefer if you didn't rename the `z3backend.ml` into the `z3backend.real.ml`, the "realness" is implicit enough I guess.

Also could you provide some documentation about what `dune` command to run? Maybe create a new target in the top-level `Makefile` ?

AltGr:
> Altough I would prefer if you didn't rename the z3backend.ml into the z3backend.real.ml, the "realness" is implicit enough I guess.

It's exactly what I tried to do at first… but dune enforces it unfortunately :shrug: 

> Also could you provide some documentation about what dune command to run? Maybe create a new target in the top-level Makefile ?

well it's not changed, just running `make build` will work as before, but if `z3` wasn't installed you will still be able to compile. Or did you mean something like a `make depends-except-z3` target ?

denismerigoux:
Yeah an explicit `make depends-except-z3` would be perfect.

denismerigoux:
BTW this should also fix https://github.com/CatalaLang/catala/issues/207, you might want to use the `make depends-except-z3` to build the JS executable of the compiler.

AltGr:
hm wait the CI still installs z3

compiler/verification/dune
```diff
@@ -1,7 +1,17 @@
...
+ (libraries
+  bindlib
+  utils
+  dcalc
+  runtime
+  calendar
+  (select
+   z3backend.ml
+   from
+   (z3 -> z3backend.real.ml)
```
denismerigoux, line 13: Can you here make the case disjunction based on the fact that the user has passed a particular compilation flag or profile?

denismerigoux:
So there are two types of CI: `run-make-all.yml` is executed on stateful CI managed by Inria on which `z3` has been preinstalled. `run-build.yml` in run on GitHub runners that build everything from scratch at each CI run.

denismerigoux (approved):

PR #224, opened 2022-03-08
DCalc optimization pass: Remove if_then_else when both branches are identical

R1kM:
This PR adds a new optimization pass to dcalc that simplifies if_then_else constructs when both branches contain the same term.
To do so, this PR adds a new function called equal_exprs to dcalc/ast.ml. This function compares two dcalc expressions modulo position information, i.e., it erases position information deeply embedded in Catala terms during comparison.
The PR then uses equal_exprs in dcalc/optimizations.ml to rewrite IfThenElse e1 e2 e3 into e2 when e2 and e3 are equal.

This is particularly useful in the context of the proof platform, where VCs are dcalc expressions that often have the shape "if complicated_constraint then b else b" where b is either true or false.


denismerigoux (approved):
Thanks very much!

PR #228, opened 2022-03-16
[Z3encoding] Basic support for arrays

R1kM:
This PR adds a basic support for arrays in the proof platform.
At the moment, it only models the (symbolic) length of arrays, and does not consider its contents. This is apparently sufficient to reason about the current version of the "allocations_familiales".

The changes are the following:
* Model arrays as a symbolic integer in Z3, corresponding to their length
* Add support for the length operator
* Extend the Z3 backend context with a `ctx_z3constraints` field, which stores a list of hypotheses regarding the symbolic variables. These hypotheses are added to the Z3 query in the `solve_vc_encoding` function.
* When creating a new symbolic array, log that its length is positive as a context hypothesis
* Add support for pretty-printing array counterexamples
* Add three unit tests for array verification (both positive and negative)

### DIFF
--- a/compiler/driver.ml
+++ b/compiler/driver.ml
@@ -28,8 +28,7 @@ let extensions =
   [ (".catala_fr", "fr"); (".catala_en", "en"); (".catala_pl", "pl") ]
 
 (** Entry function for the executable. Returns a negative number in case of
-    error. Usage:
-    [driver source_file options]*)
+    error. Usage: [driver source_file options]*)
 let driver source_file (options : Cli.options) : int =
   try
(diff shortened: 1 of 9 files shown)


denismerigoux:
Thanks so much! However I'm still getting:

```
[ERROR] [AllocationsFamiliales.montant_initial_base] The translation to Z3 failed:
[Z3 encoding] EInj unsupported
```

:)

R1kM:
See #230  :)

denismerigoux (approved):

PR #236, opened 2022-03-24
[Z3 encoding] Add support for durations

R1kM:
This PR adds support for durations in the Z3 encoding.
This is mostly straightforward, similar to dates, we encode a duration as a number of days, i.e., an integer for Z3.
To add support for translating duration literals into Z3, this PR also adds a function `duration_to_nb_days` to runtime, which directly calls the underlying function from CalendarLib.
In addition to adding support for duration, this PR also extends support for binary operators to dates, which were already supported, but had some operators that I overlooked.
This PR comes with the usual unit tests for the Z3 encoding.


R1kM:
Thanks Denis for the review! I fixed the PR accordingly.

Issue #245, opened 2022-04-20
Extract runtime to pip-installable package and add CLI?


rprimet:
Looking at the catala repo, I wonder if it would be worth extracting the python runtime (`catala.py`) to a `pip`-installable package (since it's not related to french law) and optionnally provide a generic CLI utility that would let one call a given scope with arguments? (i.e. a generic version of `main.py`)?

Rationale: this would make it easier getting started with using the python backend with arbitrary catala programs by reducing the boilerplate needed to drive the generated code

denismerigoux:
Thanks for your proposals ! Indeed that would be nice to have a `pip` package for `catala.py` to distribute it easily, and another one for `french_law` that depends on `catala`. 

I wonder what CLI you have in mind for this sort of Python interpreter you're describing : how would you pass the arguments ? It seems a bit redundant with the `Interpret` mode of the Catala compiler that would be preferable to use since it already has all the features that you would have to reimplement in Python. The rationale for the Python backend is really to be embedded into your larger Python app.

The boilerplate you might want to reduce is not in `main.py` but rather in `api.py` where we wrap the functions with the right converters between Python native types and Catala values. This wrapper could be auto-generated by the Catala compiler.

denismerigoux:
This has been fixed by #246

PR #246, opened 2022-04-20
Extract py runtime

rprimet:
Open draft PR for discussion, following #245


rprimet:
@denismerigoux I believe the remaining CI error comes from the fact that the python venv may be persistent across CI runs?

denismerigoux:
Yes the Python `venv` is persistent across the short CI runs that checks PR validity. The command for these runs is in https://github.com/CatalaLang/catala/blob/master/.github/workflows/run-make-all.yml. I noticed that this PR changes something concerning the Python venv; what should I do to restore the CI ?

rprimet:
Since the env setup step is suite short, maybe rm'ing the `env` directory and running `setup_env.sh` before each run of `make all` would be a solution ? (Even without deleting the env directory on each run, we should run `pip install . ` in the python runtime directory before each run if we go forward with an installable runtime -- which `setup_env.sh` does)

denismerigoux:
I added you as an admin of the Inria CI project (ci.inria.fr) where the machines are hosted. You can login with user `ci` and a password I'll send you via email. The Catala repo where the actions happen is in `/builds/actions-runner/catala/catala` on each one of the two machines and `tmux a` brings you to a session where the GitHub actions runner listents to income job requests.

french_law/python/src/allocations_familiales.py
```diff
@@ -1,59 +1,58 @@
 # This file has been generated by the Catala compiler, do not edit!
```
denismerigoux, line 1: Did you format this with `autopep8` ?

runtimes/python/catala/src/catala/runtime.py
```diff
@@ -70,6 +70,9 @@ def __eq__(self, other: object) -> bool:
 
     def __str__(self) -> str:
         return self.value.__str__()
+    
+    def __repr__(self) -> str:
```
denismerigoux, line 74: Thanks!

runtimes/python/catala/src/catala/runtime.py
```diff
@@ -70,6 +70,9 @@ def __eq__(self, other: object) -> bool:
```
denismerigoux, line 70: Good idea to put the runtimes in a separate folder. I should do the same for OCaml.

denismerigoux (approved):
Thanks very much @rprimet!

french_law/python/src/allocations_familiales.py
rprimet: Nope, I just did, thanks!

(BTW, any reason for choosing autopep8 over black?)

denismerigoux: I didn't know black :)

denismerigoux (approved):

PR #256, opened 2022-05-05
Fix money parsing

denismerigoux:
Embarrasingly, there was a bug in the parsing of money literals like `0,3 €` or `$0.3` : they were counted as "3 cents" instead of "30 cents" due to a string to integer conversion error. This PR fixes this behavior by padding a 0 on the right to these kinds of money literals.


denismerigoux:
Yes you're right. I implemented your suggestion.

AltGr (approved):

PR #206, opened 2022-02-21
Revert "Merge pull request #196 from CatalaLang/fixup-cmdliner"

lIlIlIlIIIIlIIIllIIlIllIIllIII:
fix nix build reverting the update of  cmdliner from 1.0.4 to 1.1.0

PR #258, opened 2022-05-06
Upgrade cmdliner to 1.1 (some API changes)

AltGr:
also add an opam "depext" for ninja so that ninja-build gets
automatically installed with Clerk (on debian/ubuntu at least, we
should investigate the name of the package required on other distros)


catala.opam
```diff
@@ -37,7 +37,9 @@ depends: [
   "js_of_ocaml-ppx" {>= "3.8.0"}
   "camomile" {>= "1.0.2"}
   "cppo" {>= "1"}
+  "obelisk" {dev}
```
denismerigoux, line 40: Be careful when added these lines, the `opam` team asked me to remove them last time I published: https://github.com/ocaml/opam-repository/pull/20861#issuecomment-1061807613

clerk.opam.template
```diff
@@ -0,0 +1 @@
+depexts: ["ninja-build"] {os-family = "debian"}
```
denismerigoux, line 1: Why is the `.template` file needed ?

dune-project
```diff
@@ -49,7 +49,7 @@
   (bindlib
    (>= 5.0.1))
   (cmdliner
-   (= 1.0.4))
+   (>= 1.1.0))
```
denismerigoux, line 52: So last time we tried that it broke the Nix build and we had to revert: https://github.com/CatalaLang/catala/pull/206. Tagging @lIlIlIlIIIIlIIIllIIlIllIIllIII for support on that issue.

french_law.opam
```diff
@@ -13,6 +13,7 @@ depends: [
   "dune" {>= "2.8"}
   "ocaml" {>= "4.11.0"}
   "catala" {= version}
+  "conf-npm" {dev}
```
denismerigoux, line 16: Ditto https://github.com/ocaml/opam-repository/pull/20861#issuecomment-1061807613

denismerigoux (approved):
Thanks! I'm all for it if it doesn't break the nix build.

AltGr:
Answering several of your questions at once: the `.opam` file are generated from the `.dune_project` file — this is done with the prospect of better integration in the future, but I am not sure of the benefit for us at this point:
- there is some duplication: I didn't touch the `conf-npm` etc. deps, they just got regenerated when I hit `dune build *.opam` as these were not in sync
- `.dune_project` does'nt support `depexts`, so we have to rely on a .template file for `dune` to generate the field. A bit heavyweight...

I might be missing something with potential benefits of using generated opam files (maybe for `dune-release` ?) but using plain opam files and de-duplicating the info in `.dune_project` would seem reasonable at this point. Of course I know I am being partial since I know the opam format by heart and not dune details !

dune-project
AltGr: That's because `1.1` had breaking API changes¹ ; so the new constraint should be safe for now, but if we want to be defensive we could add `>= "1.1.0" & < "1.2"`. That could apply to all the other dependencies as well though, the approach of the opam repository maintainers is usually to only add these constraints as they arise.

¹ Actually probably not really, I think they kept compatibility but just with deprecation warnings. Does the Nix build by any chance have warnings-as-errors :scream: ?

AltGr:
(I have been opening discussions above, but it'd be perfectly fine to remove the spurious `{dev}` constraints from their source in the `dune_project` file and keep the template file to just fix this PR!)

dune-project
denismerigoux: In my opinion no need to add `>= "1.1.0" & < "1.2"`, it's a problem of the Nix build that ought to be fixed. I think we have warning-as-errors everywhere in this project, including in non-nix builds with `dune build`.

denismerigoux:
Not particularly happy with the opam generation by dune either, I would be in favor of manually creating and editing the opam files! But in that case we should remove the dependencies information for the `dune-project` file to avoid duplication.

dune-project
AltGr: > I think we have warning-as-errors everywhere in this project

this is fine for development (and obvious for CI) but shouldn't be enabled for releases and packages (esp. for the "deprecation" warning, we were bitten quite a lot by packages breaking this way in the early days of the opam repo!)

dune has reasonable defaults for the `--profile=dev` and `--profile=release` options though so that should be ok.

denismerigoux:
As you can see the nix build is failing again: https://github.com/CatalaLang/catala/runs/6349860983?check_suite_focus=true @lIlIlIlIIIIlIIIllIIlIllIIllIII

denismerigoux:
That's because nix has its own list of dependencies inside `.nix/` and does not rely on opam. Instead, each opam package has a mirror nix package whose versions are not always maintained in sync with the opam ones.

PR #259, opened 2022-05-10
Fix ocamlformat crashes, and tweak some settings

AltGr:
Disclaimer: I am aware this just a matter of personal preferences and don't intend to start a war.

Then again, if there is no strong disagreement... :)

We should wait for the right moment to merge this atomically without creating too much conflicts (it's easy to rebase, but the other branches might not).

Goodies:
- fixed some files where ocamlformat didn't pass
- add a git-blame ignore file to keep `git blame` working reasonably
- a small script to reformat a local branch (based on origin/master; make sure you install the proper version of ocamlformat):

```bash

if [ $# -gt 0 ]; then
    echo "Reformats the current branch patch by patch, according to the ocamlformat"
    echo "config at origin/master. Use before rebase."
    exit 1
fi

reformat() {
    git checkout origin/master .ocamlformat
    dune build @fmt 2>/dev/null || true
    dune promote
}

BRANCH=$(git branch --show-current)
if [ -z "$BRANCH" ]; then echo 'Please be on a branch' >&2; exit 2; fi

TMP_BRANCH=tmp-reformat-$BRANCH

trap "git checkout --force $BRANCH; git branch -D $TMP_BRANCH" EXIT

BASE=$(git merge-base origin/master "$BRANCH")
COMMITS=$(git log --format=%H --reverse "$BASE".."$BRANCH")

git checkout -B "$TMP_BRANCH" "$BASE"
reformat
git diff --quiet || git commit -am "Formatting: sync settings with master"

for COMMIT in $COMMITS; do
  git checkout "$COMMIT"
  reformat
  git diff --quiet || git commit -am "tmp-reformat"
  C=$(git rev-parse HEAD)

  git checkout "$TMP_BRANCH"
  git diff --binary HEAD..$C | git apply --index -
  git commit -C "$COMMIT"
done

git checkout "$BRANCH"
git reset "$TMP_BRANCH" --hard
```

### DIFF
--- a/.github/workflows/run-make-all.yml
+++ b/.github/workflows/run-make-all.yml
@@ -21,6 +21,20 @@ jobs:
         run: |
           rm -rf french_law/python/env
           ./french_law/python/setup_env.sh
+      - name: Install dependencies
+        run: |
+          opam exec -- make dependencies
+      - name: Check promoted files
+        run: |
+          rm -f bad-promote
+          opam exec -- make check-promoted > promotion.out 2>&1 || touch bad-promote
       - name: Make all
         run: |
-          OCAMLRUNPARAM=b opam exec -- make dependencies all -B
+          OCAMLRUNPARAM=b opam exec -- make all
+      - name: Forward result from promotion check
+        run: |
+          if [ -e bad-promote ]; then
+            echo "[ERROR] Some promoted files were not up-to-date";
+            cat promotion.out;
+            exit 1
+          fi
--- a/Makefile
+++ b/Makefile
@@ -40,21 +40,13 @@ dependencies-with-z3: dependencies-ocaml-with-z3 dependencies-js init-submodules
 COMPILER_DIR=compiler
 BUILD_SYSTEM_DIR=build_system
 
-format:
-	dune build @fmt --auto-promote >/dev/null
-
 #> build_dev				: Builds the Catala compiler, without formatting code
-build_dev:
-	dune build @update-parser-messages --auto-promote | true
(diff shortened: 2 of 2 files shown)


AltGr:
Hm the CI runner doesn't have `ocamlformat.0.21.0`: an `opam update` should be run.
Or should that be done unconditionally on the CI ?

denismerigoux (approved):

denismerigoux:
So I tried to do `opam update` and `opam install ocamlformat.0.21.0` on the CI machines but then it recompiled a bunch of packages and `easy-format` building keeps failing with

    # [ERROR] /builds/.opam exists, but does not appear to be a valid opam root. Please remove it and use `opam init', or specify a different `--root' argument

I tried `rm -rf ~/.opam` + `opam init` multiple times and reinstalling opam from the installation script but I keep getting the same error. At this point I've sent an email to you (louis.gesbert@inria.fr) about getting root SSH access to you on the CI machines, can you take a look ?

denismerigoux:
To connect the runners to GitHub actions again you will have to spin a `tmux` session and launch `~/actions_runner/run.sh` in there.

AltGr:
it's an opam sandbox error: your opam root is outside of `/home` and thus invisible from within the sandbox. During its build the `easy-format` package tries to run `opam var prefix` and fails since it can't load the root.
I think there is already a workaround on master :thinking:

AltGr:
Just exporting `OPAM_USER_PATH_RO` seems to be enough as a workaround for now:
```bash
OPAM_USER_PATH_RO=/builds ./actions-runner/run
```

PR #268, opened 2022-05-25
Compiler: support cumulative exceptions

### LINKED ISSUE #208: Non-exclusive/cumulative exceptions
## The problem

In Catala, exceptions are exclusive by default. This means that if two exceptions are triggered at the same time, the program's execution is aborted with an error message. But consider this program:

```catala
scope TaxComputation:
  definition tax equals income *$ 20%

  exception definition tax under condition 
    income <$ $30,000
  consequence $0
 
  exception definition tax under condition  
    is_disabled
  consequence $0
```

In this example, it is possible that the two exceptions trigger at the same time, if the person is disabled and has an income less than $30,000. In that case, the program would abort with an error. But here we don't want the program to abort, since *the two exceptions return the same consequence*. The exceptions don't conflict, they are cumulative.

This pattern has happened a lot in eligibility computations for the French housing benefits and sounds quite general.

## Solution 1: implicit cumulativity

The correct semantics that we expect from this program is to merge the two exceptions into a single one trigger by the conjunction of both conditions: ` is_disabled or  income <$ $30,000`. This merge can be done during the desugaring phase or during the construction of the default tree in the translation from desugared to the scope language.

We could perform this merge automatically when detecting that multiple exceptions have *syntactically equal* consequences. Since syntactic equality implies semantic equality, this merge should be safe.

The downside of performing the exception merge automatically is to change quite dramatically the meaning of the `exception` surface syntax keyword. Is exception cumulativity in the presence of identical consequences intuitive or will it cause confusion to users?

## Solution 2: explicit cumulativity

If we judge that implicit cumulativity is a problem, then we should make an explicit syntax to declare that a set of exceptions at the same level of priority are cumulative. Declaring the exceptions as cumulative does not remove the check of syntactic equality for consequences. On the contrary, declaring exceptions with different consequences cumulative should yield a compiler error. The syntax for cumulative exceptions should look like that:

```catala
scope TaxComputation:
  label base_case definition tax equals income *$ 20%

  label no_tax exception base_case cumulative definition tax under condition 
    income <$ $30,000
  consequence $0
 
  label no_tax exception base_case cumulative definition tax under condition  
    is_disabled
  consequence $0
```

The `cumulative` keyword is associated to the two exceptions grouped together by sharing the same `no_tax` label.

## Discussion

Should we have solution n°1 or n°2?

AltGr:
Closes #208 (implementing Solution 1, without adding an explicit syntax)

Two exceptions or more, e.g. `(j1 |- c1)` and `(j2 |- c2)` such that `c1
= c2`, are collapsed by this transformation into `((j1 |- c1) | j2 |- c2)`,
introducing an arbitrary precedence that avoids the conflict.

The transformation is not applied if any exceptions apply to the subterms
themselves: while these exceptions could be merged, that would turn more
conflicts into arbitrary outcomes than wanted.

Probably easier to review patch by patch.


Makefile
```diff
@@ -78,7 +78,7 @@ compiler/surface/parser.messages: compiler/surface/tokens.mly compiler/surface/p
 parser-messages: compiler/surface/parser.messages
 
 format:
-	dune build @fmt --auto-promote >/dev/null || true
+	dune build @fmt --auto-promote 2>/dev/null || true
```
denismerigoux, line 81: Yes thanks!

compiler/scopelang/ast.ml
```diff
@@ -199,4 +328,19 @@ let make_let_in
        (Pos.get_position (Bindlib.unbox e2)))
     (Bindlib.box_list [e1])
 
+let make_default ?(pos = Pos.no_pos) exceptions just cons =
```
denismerigoux, line 331: Ok so you integrate some optimization in there already. Might be worth mentioning it in a comment saying that this is a semantics-preserving optimization, that duplicates what is done in `dcalc/optimizations.ml`

compiler/scopelang/ast.mli
```diff
@@ -164,3 +169,13 @@ val make_let_in :
...
+
+val make_default :
+  ?pos:Pos.t ->
+  expr Pos.marked list ->
+  expr Pos.marked ->
+  expr Pos.marked ->
+  expr Pos.marked
+(** [make_default ?pos exceptions just cons] builds the equivalent to an
+    [EDefault] term, while avoiding redundant nested constructions. The position
+    is extracted from [just] by default. *)
```
denismerigoux, line 181: The fact that the position is extracted from `just` is a printing trick that will play nicely with the error messages raised when two exceptions conflict with each other : the error message will print the positions of the conflicting justifications.

compiler/scopelang/scope_to_dcalc.ml
```diff
@@ -115,6 +115,45 @@ let tag_with_log_entry
         Pos.get_position e ))
     e
 
+(* In a list of exceptions, it is normally an error if more than a single one
+   apply at the same time. This relaxes this constraint slightly, allowing a
+   conflict if all the triggered conflicting exception yield syntactically equal
+   results (and as long as none of these exceptions have exceptions themselves)
```
denismerigoux, line 121: A description of the collapsed encoding in the comment here would be nice.

compiler/scopelang/scope_to_dcalc.ml
```diff
@@ -115,6 +115,45 @@ let tag_with_log_entry
...
+
+   NOTE: the choice of the exception that will be triggered and show in the
```
denismerigoux, line 123: I thought the encoding with a chain of exceptions made the choice of the triggered exception not abritrary any more ?

tests/test_bool/good/output/test_bool.catala_en.Scopelang
```diff
@@ -1,8 +1,4 @@
 let scope TestBool (foo: bool|context|output) (bar: integer|context|output) =
-  let bar : integer = reentrant or by default
-    ⟨⟨true ⊢ ⟨⟨true ⊢ 1⟩ | false ⊢ ∅ ⟩⟩ | true ⊢
-      ⟨⟨false ⊢ ∅ ⟩ | false ⊢ ∅ ⟩⟩;
+  let bar : integer = reentrant or by default ⟨1 | false ⊢ ∅ ⟩;
```
denismerigoux, line 2: Ok I agree that simplifies things A LOT.

tests/test_io/good/output/condition_only_input.catala_en.B.Dcalc
```diff
@@ -1,7 +1,6 @@
 let B =
   λ (B_in_31: B_in{}) →
-    let a.x_32 : bool = error_empty
-      ⟨true ⊢ ⟨⟨true ⊢ false⟩ | false ⊢ ∅ ⟩⟩ in
+    let a.x_32 : bool = error_empty false in
```
denismerigoux, line 3: `error_empty <a non-empty literal>` should be simplified in `<a non-empty literal>` in `dcalc/optimizations.ml`.

denismerigoux: Maybe it's already the case?

tests/test_proof/bad/enums_unit-overlap.catala_en
```diff
@@ -13,8 +13,8 @@ scope A:
   definition x equals Case1 content 2
   definition y under condition match x with pattern
     -- Case1 of i : true
-    -- Case2 : true consequence equals 2
+    -- Case2 : true consequence equals 3
```
denismerigoux, line 16: Interesting, that test was now passing :)

examples/allocations_logement/code_construction_legislatif.catala_fr
```diff
@@ -264,7 +264,7 @@ n'est pas prise en compte dans le calcul de l'aide personnelle au logement.
 
 ```catala
 champ d'application ÉligibilitéAidePersonnelleLogement:
-  étiquette l822_5_2 exception r822_22_3 définition patrimoine_pris_en_compte
+  exception r822_22_3 définition patrimoine_pris_en_compte
```
denismerigoux, line 267: Thanks, but I'll have to review the logic of this code to make sure we're doing the right thing so I would prefer if this PR did not touch anything in `examples/allocations_logement`.

denismerigoux (changes requested):
Thanks very much Louis for this feature addition that will be very useful. Apart from the comments I left, there's one more thing I'd like to ask from this PR. So far, I've tried to keep the English tutorial (in `examples/tutorial_en`)  in sync with the language features when I added them. Could you insert a paragraph describing this new behavior in the tutorial ? Otherwise it will be pretty confusing for programmers because it's quite a silent rewriting.

examples/allocations_logement/code_construction_legislatif.catala_fr
AltGr: Alright! I used this to check that the transformation was working on our real use-case, but was not sure about committing it either (and it was a good idea since at first the rewriting worked on trivial examples but not here)

compiler/scopelang/scope_to_dcalc.ml
AltGr: well, I think that if we have `(a |- c)` and `(b |- c)`, the trace will show that condition `a` was triggered if we chain as `<(a |- c) | (b |- c)>`, or `b` if we chain as `<(b |- c) | (a |- c)>` ; both ways of chaining seeming equally correct ?

AltGr:
@denismerigoux I'd be glad if you could review [the tutorial section](https://github.com/CatalaLang/catala/pull/268/commits/81fe7eba47c21e0b89e81cb4403abf8d79ce3502), in particular for correct wording. And if it's not too weird to include a strange natality-friendly policy in the example… `number_of_children` was the only variable already available :shrug:

compiler/scopelang/scope_to_dcalc.ml
denismerigoux: The log operators are setup this way (you can show them in Dcalc with the `-d` flag) : `<( (print_trace_if_true a ) |- c) | ((print_trace_if_true b) |- c)>`.  So yes if both `a` and `b` are true then it will show `a` or `b` arbitrarily but deterministically, which is fine. But if only `a` or `b` is true then it will show the right one.

examples/tutorial_en/tutorial_en.catala_en
```diff
@@ -526,6 +526,44 @@ patterns. Sometimes, you want to declare an exception to a group of
...
+    individual.number_of_children >= 7
+  consequence equals $0
+```
+
+The same problem as above would be triggered for families with an income below
+`$10,000` and 7 children or more. But here Catala can detect that it won't
+matter since the result in both cases is an exemption.
+
+```catala
+declaration scope Test4:
```
denismerigoux, line 554: Good example!

AltGr:
Good to merge ? Since you reviewed with "requesting changes" I can't merge anymore without your greenlight :)

denismerigoux (approved):
Thanks for the changes!

PR #270, opened 2022-05-30
Fix Nix expressions and convert them to flakes

W95Psp:
Hi,

We've looked into the Nix expressions with @lIlIlIlIIIIlIIIllIIlIllIIllIII.
This PR fixes the Nix expressions that build Catala, and converts them into [flakes](https://nixos.wiki/wiki/Flakes), which enhances reproducibility.


lIlIlIlIIIIlIIIllIIlIllIIllIII:
we still need to change a few things inside the github actions

PR #279, opened 2022-06-08
fix(compiler): fix line num for HTML generation

EmileRolley:
Fix line numbering for the Html backend.

Catala code block delimiters are removed after the pygmentization and this must be taken into account for the line numbering.

https://github.com/CatalaLang/catala/blob/1f4e869c33dfc1a8f0b7cf892b7a47d340429f30/compiler/literate/html.ml#L154-L162


denismerigoux (approved):
Thank you !

PR #285, opened 2022-07-08
Partially fix #192

### LINKED ISSUE #192: Specifying catala options in Clerk and backend testing
Some features are missing inside clerk 

* The possibility to write tests cases corresponding to specific catala's compiler arguments. For instance `--optimize` or `--avoid_exceptions` modify the compilation results, hence failing the test. Adding other outputs formats (for instance .catala_en.avoid_exception.ml) could solve the problem, but other solution might be better.
* To check if a compilation pass is correct, one way is to check whenever it type-check correctly inside ocaml. To do so, we need at least to import catala.runtime. One solution could be to generate dune files.

denismerigoux:


AltGr:
Remains to update `tests/README.md` and this should be good to go ?

PR #288, opened 2022-07-12
feat(backend): disambiguate OCaml record names derived from Catala structs

### LINKED ISSUE #191: Fix the OCaml backend's record field confusion
## The problem

Suppose you define two structs in your Catala program that yield input and output signatures that have the exact same names:

```catala
declaration scope A:
  context a content bool
 
declaration scope B:
  context a content bool
```

Then, the OCaml backend will generate structs like `A_in { a_in: bool }` and `B_in { a_in : bool }`. The problem is that when the OCaml generated code tries to build the input for either of the structs with things like `{ a_in = true }`, there is a confusion to whether it means `A_in` or `B_in`.

## The solution 

This is a known problem in OCaml and the solution is straightforward, as detailed in https://stackoverflow.com/questions/68178153/fixing-type-inference-when-record-types-have-same-labels. For the OCaml backend of Catala, we need to encapsulate the definitions of the structs in their own module, so that we can prefix the field names with the module name and resolve the confusion.

This solution should only modify `compiler/lcalc/to_ocaml.ml`, and provide a new test case showcasing the confusion as presented above.

EmileRolley:
Fixes #191.


denismerigoux (approved):
Thanks!

Makefile
```diff
@@ -272,10 +272,10 @@ CLERK=$(CLERK_BIN) --exe $(CATALA_BIN) \
 .FORCE:
 
 test_suite: .FORCE
-	$(CLERK) test tests
+	@$(CLERK) test tests
 
 test_examples: .FORCE
-	$(CLERK) test examples
+	@$(CLERK) test examples
```
denismerigoux, line 278: Not sure you want to do that, it's fairly useful for debug

denismerigoux (approved):

Issue #290, opened 2022-07-12
Feat: add a new plugin for wrapper generation from Lcalc


EmileRolley:
In order to be able to fully automate the web translation of Catala programs, such as automatic form generation, etc.. We need to provide first a plugin to generate a JSOO wrapper from the Lcalc representation.

**ADT encoding**
Catala ADT are encoded in tag unions where the following type

```ocaml
type b = A | B of int
```

is encoded into the following JS class:

```ocaml
class type b =
  object
    method kind : Js.js_string Js.readonly_prop (* Expects "A" or "B" *)
    method args: Js.Unsafe.any_js_array Js.t Js.readonly_prop
  end
```

PR #291, opened 2022-07-18
feat(plugins): add a new backend for generating JSOO wrapper

EmileRolley:
Implementation of the issue #290.

## Changelog

* New plugin `jsoo` generating a `js_of_ocaml` wrapper around a Catala program. 
* New runtimes:
   - The OCaml runtime has been moved in `./runtimes/ocaml` -- `Runtime_ocaml.Runtime`.
   - A new JSOO runtime is stored in `./runtimes/jsoo` -- `Runtime_jsoo.Runtime`.
* The `To_ocaml` module now exposed more functions used in `To_jsoo`.
* The OCaml generated variant types are now encapsulated inside their own module like struct in #288.
* New function `Cli.ocamlformat_file_opt` which calls `ocamlformat` to format generated ocaml files.
* The Makefile rule `generate_french_law_library_ocmal` now depends on the `plugins` one and generates the `law_source/allocations_famililales_api_web.ml` file. 
* New Makefile rule `%_api_web.ml` in `./examples/Makefile.common.mk` to use the `jsoo` plugin.


denismerigoux (approved):
Perfect! I'll merge that now.

PR #295, opened 2022-07-22
Expanding french_law with the French housing benefits

denismerigoux:
Generating OCaml, JS and Python artefacts from the French housing benefits example.


denismerigoux:
@EmileRolley if you invoke `make build_french_law_library_ocaml -B` you'll get an error message that corresponds to a bug in your JSOO wrapper plugin for the handling of `context` scope variables (both input and defined in the scope). Could you take a look at it next week ?

EmileRolley:
It seems related to the management both of the `unit` type and the `duration` which I missed beacause they doesn't appears in the `allocations_familiales` examples.

It should work now :)

PR #297, opened 2022-07-27
Change syntax for set membership predicate ('in' -> 'contains')

AltGr:
Also includes a small CLI patch


AltGr:
I am surprised that only 1 test and no examples are affected by the change :raised_eyebrow:

AltGr:
Update: replaced "comprend" by "contient" for the french version, and got a :heavy_check_mark: from the legal team

compiler/surface/lexer_pl.cppo.ml
```diff
@@ -29,6 +29,7 @@
 #define MS_INCREASING "rosnacy"
 #define MS_OF "z"
 #define MS_COLLECTION "kolekcja"
+#define MS_CONTAINS "contains"
```
denismerigoux, line 32: The translation of `contains` in Polish seems to be "[zawiera](https://context.reverso.net/translation/polish-english/zawiera)"

denismerigoux (approved):
All good, thanks ! Except the Polish translation, we should keep this frontend updated while we can.

compiler/surface/lexer_pl.cppo.ml
AltGr:
```suggestion
#define MS_CONTAINS "zawiera"
```

Issue #219, opened 2022-03-08
Portable executables for Catala


denismerigoux:
## Context 

For Catala to be usable by industrial users, it would be hugely beneficial if its distribution and installation would not entail installing the whole `opam` toolchain + the C/C++ dependencies : GMP, Z3.

First, as mentionned by https://github.com/CatalaLang/catala/issues/207, it would be good to be able to not include Z3 in the Catala compiler on demand with a custom build profile.

## Portable executables

Second, an easy way to distribute the Catala compiler would be as a static, portable executable. @AltGr wrote this excellent tutorial about static binaries compiled from OCaml code: https://www.ocamlpro.com/fr/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml. Following it, I was able to produce portable Linux/ELF x86-64 binaries for [Catala](https://github.com/CatalaLang/catala/releases/download/0.6.0/catala_v_0_6_0_Linux_x86_64.exe) and [Clerk](https://github.com/CatalaLang/catala/releases/download/0.6.0/clerk_v_0_6_0_Linux_x86_64.exe) included in the v0.6.0 release. 

## First issue: z3 static linking

When creating these binaries, I had to use `z3.4.8.11` instead of the latest `z3.4.8.14` because `z3.4.8.13` introduced a breaking change that forbids static linking: 

https://github.com/ocaml/opam-repository/blob/1fbd9337269c59f74b4046cd9a60efeaa9e52d02/packages/z3/z3.4.8.13/opam#L32-L37
 
Can we overcome this ? The OCaml API readme in Z3's folder still seems to say that you can statically link https://github.com/Z3Prover/z3/tree/master/src/api/ml

## Second issue: cross-compilation

Right now I managed only to compile for x86 ELF because that's what I got on my machine but we would need Windows and ARM binaries as well, preferably automatically generated through some CI that gets called at each release.

denismerigoux:
Alternative cheap solution: a node JS binary... https://github.com/CatalaLang/catala/releases/download/0.6.0/catala_v_0_6_0_node.js

PR #299, opened 2022-07-29
Build system rehaul, use of Docker for CI & artifact generation on Github

AltGr:


denismerigoux:
Wouahou ! So how does this work ? Will the generated files be updated nightly by a bot ? Or appear in a source code release attached to the GitHub page ? I put those generated assets under Git because I wanted for somebody that wanted to just use Catala to be able to use the assets directly without having to install the compiler to generate them.

AltGr:
Aah I didn't know!

But indeed, providing them as artifacts would be pretty easy: they are already generated by the CI so extracting them from the Docker image and feeding them to Github artifacts / release assets should be fairly straightforward.

Is there already a list of artifacts that we should extract in that way ?

denismerigoux:
So for me, ideally:
* All of `french_law` plus their `runtime` dependency as a source code zip ;
* Linux/Windows/Mac x86 executables of the Catala compiler (without Z3) + Clerk ;
* A Javascript binary of the Catala compiler (without JS)
You can take a look at the assets I put in the last release.

A push-button release process would be amazing, I would release more often that today with that.

AltGr:
Ok I'll handle that

denismerigoux:
See #219 where I talked a bit about this

AltGr:
Should now be OK; generation of the artifacts is only activated on the master branch but I tested it [here](https://github.com/CatalaLang/catala/actions/runs/2792168055). (I removed node's `node_modules` & python's `env` since then)

Compared to the first version, I leveraged dune's "promotion" feature which means I was able to keep the `french_law` artifacts in-tree as before while working around the issue of them getting stale or `dune` getting confused.

Makefile
```diff
@@ -357,14 +353,15 @@ all: \
 	tests_ocaml bench_ocaml \
 	build_french_law_library_js \
 	bench_js \
-	generate_french_law_library_python type_french_law_library_python\
+	generate_french_law_library_python type_french_law_library_python \
```
denismerigoux, line 356: Ok I see you've moved all the logic out of the Makefile and into `dune`. That's fine by me!

build_release.sh
```diff
@@ -0,0 +1,46 @@
...
+ -cclib -no-pie
+ -noautolink
+ -cclib -L/home/ocaml/.opam/z3/lib/stublibs
+ -cclib -lz3ml-static
+ -cclib -lz3
+ -cclib -lstdc++
+ -cclib -lthreadsnat
+ -cclib -lzarith
+ -cclib -lgmp
+ -cclib -lcamlstr
```
denismerigoux, line 20: So you've managed to do static linking with Z3 ? How painful was it ?

build_system/dune
```diff
@@ -16,9 +16,20 @@
   ANSITerminal)
  (modules clerk_driver))
 
+(rule
```
denismerigoux, line 19: What's that for ?

examples/aides_logement/dune
```diff
@@ -0,0 +1,33 @@
...
+   api_web
+   -t
+   --plugin-dir=../../compiler/plugins
+   aides_logement.catala_fr)))
+
+(rule
+ (targets aides_logement_schema.json)
+ (deps
+  (source_tree ..)
+  ../../compiler/plugins/json_schema%{ext_plugin})
```
denismerigoux, line 18: Ok we should leave some documentation about how to write those kind of rules in `examples/README.md`

french_law/ocaml/law_source/.ocamlformat_ignore
```diff
@@ -0,0 +1 @@
+*
```
denismerigoux, line 1: You did that to speed up the build, because we don't need to format generated files ?

french_law/ocaml/law_source/aides_logement.ml
```diff
@@ -4812,7 +4788,7 @@ let base_mensuelle_allocations_familiales
                handle_default
                  {
                    filename =
-                     "examples/aides_logement/../base_mensuelle_allocations_familiales/bmaf.catala_fr";
+                     "./../base_mensuelle_allocations_familiales/bmaf.catala_fr";
```
denismerigoux, line 4791: So this probably won't work. We can't have `../` in these paths, as they mess up the display of things in the website HTML-wise. That's why we made sure to call `catala` from the root directory to generate these OCaml files so that the paths all begin with `examples/`.

french_law/ocaml/law_source/dune
```diff
@@ -4,3 +4,27 @@
  (preprocess
   (pps js_of_ocaml-ppx))
  (libraries catala.runtime_ocaml catala.runtime_jsoo js_of_ocaml))
+
+(rule
+ (target allocations_familiales.ml)
+ (mode promote)
```
denismerigoux, line 10: Ok, neat!

generate_website_assets.sh
```diff
@@ -1,22 +1,29 @@
...
-scp french_law/js/french_law.js $1/french_law.js
-scp examples/allocations_familiales/allocations_familiales_schema.json $1/
-scp examples/aides_logement/aides_logement_schema.json $1/
+mkdir -p $1/playground
+
+BUILD=_build/default
+
+rsync -a $BUILD/_doc/_html/ $1/ocaml_docs/
+
+rsync $BUILD/examples/allocations_familiales/allocations_familiales.html $1/
```
denismerigoux, line 18: Thanks!

denismerigoux (approved):
That looks awesome, a well-deserved overhaul of the build system. I raised a few nitpicks, you can merge after dealing with them.

build_release.sh
AltGr: Actually not  harder than [passing the correct configure flag](https://gitlab.inria.fr/lgesbert/catala-ci-images/-/blob/z3static/Dockerfile#L9)

(and then tweaking the link to `-lz3` instead of `-lz3static`)

build_system/dune
AltGr: Avoid a failure when the `custom_linking.sexp` file doesn't exist (we default it to a file containing just `()`)

build_release.sh
denismerigoux: Ha ha ok this was beyond my abilities, I suffered for days on this  thing before giving up.

examples/aides_logement/dune
AltGr: Hm indeed! Note that we don't have dependency analysis here (shouldn't we extend Clerk ?) so I include the whole `example` subtree.

french_law/ocaml/law_source/.ocamlformat_ignore
AltGr: Ah indeed that's something I wanted to discuss with you!
It feels a bit strange to me to reformat just-generated files; that also adds a run-time dependency to `ocamlformat` for running these rules. And it complicated things (we don't want ocamlformat and dune to compete on the resulting file).

No big speedup here (but in the case of the Python files, `autopep8` is very slow)

But I guess there were reasons to do this in the first place so I am open to finding a workaround if needed!

examples/aides_logement/dune
denismerigoux: There is a `Makefile` backend to Catala that performs dependency analysis (inclusions) when called on a file, it produces a `.d` file.

french_law/ocaml/law_source/.ocamlformat_ignore
denismerigoux: You need to at least read the input type declarations to know what to feed your program. And since the pretty-printing in the backends is not ideal we run ocamlformat afterwards. @EmileRolley even calls directly ocamlformat in his new `api_web` backend plugin.

french_law/ocaml/law_source/aides_logement.ml
AltGr: Ah right! I had noticed this then forgot to check :/
I'll find a fix.

Makefile
AltGr: That was not strictly required for Python, but it was for js since we go through OCaml and use dune again further down the pipeline: so fo the sake of consistency…

build_system/dune
AltGr: (we could also statically add the file to the source tree; the upside of that would be that it is harder to get confused because you had a stale file sitting there; but it's not very pretty)

examples/aides_logement/dune
AltGr: I know ! But there is no straight-forward way to use those in dune (that I know about; support for `ocamldep` is built-in) — so I opted for the lazy solution for now.

french_law/ocaml/law_source/.ocamlformat_ignore
AltGr: Indeed! Just discussed the matter quickly with @EmileRolley and we agreed to disable it and focus on improving our pretty-printer on cases where readability is bad (from a quick check, it seems o.k. on type definitions). Otherwise, our pretty-printer will be completely untested: if we don't want to bother, and use OCamlformat, I'd say go all the way, simplify our pretty-printer into a bare-printer and call `ocamlformat` as a lib.

What do you think ?

Makefile
denismerigoux: Yep I agree, let's try and keep things consistent.

build_system/dune
denismerigoux: Ok and the `custom_linking.sexp` is produced by the `build_release.sh` script. Maybe that would be worth a comment in the dune file. I'm in favor of not adding `custom_linking.sexp` to the build tree.

examples/aides_logement/dune
denismerigoux: Ok, we can also add a new backend to the Catala compiler, who would share a lot of code with the Makefile backend, that would produce dune rules exactly as we expect.

french_law/ocaml/law_source/.ocamlformat_ignore
denismerigoux: I'm okay with relying on our pretty-printer and improving it as needed, dropping `ocamlformat` altogether. Using `ocamlformat` as a lib is tempting but would introduce a lot of complexity as we use only a small portion of the OCaml AST. A textual pretty-printer would be simpler for maintainers to come and just change a tidbit while not having to understand everything. Plus that way we don't have to keep in sync with `ocamlformat`'s AST changes, while OCaml textual syntax is somewhat backwards-compatible.

PR #300, opened 2022-07-29
Add source positions in all backends exceptions

denismerigoux:
With this PR, it is possible to spot which conflict happened easier when using the compiled versions of Catala programs.


EmileRolley (approved):
Looking good to me!

I'll improve the website to properly render error messages.

PR #303, opened 2022-08-02
Feat: improve the HTML generation with ToC and details tags

EmileRolley:
Related issue: https://github.com/CatalaLang/catala-website/issues/33

To facilitate navigation inside generated HTML pages, this PR adds the generation of ToC and adds `<details>` tags around `<h2>` sections.

New `Utils.String_common` module grouping helper functions about string manipulation based on `Ubase`, which allows to remove : `Lcalc.Backends` and `Lcalc.Print` modules.

## Demo _new_

https://user-images.githubusercontent.com/44124798/182911118-95810701-a908-4129-b27c-3a44f30d4611.mp4


compiler/literate/html.ml
```diff
@@ -158,6 +159,10 @@ let pygmentize_code (c : string Marked.pos) (language : C.backend_lang) : string
 
 (** {1 Weaving} *)
 
+let sanitize_html_href str =
+  str |> Ubase.from_utf8
```
denismerigoux, line 163: Wouahou I didn't know that `ubase` was a thing, actually it was released in January 2022, only after I already had coded up this piece:

https://github.com/CatalaLang/catala/blob/2da94b88c27d6017851bcd8fbffb4c7b116a5717/compiler/lcalc/backends.ml#L17-L42

Could you replace this piece of code in `backends.ml` by the use of `ubase` ? You could also get rid of the dependency on `camomile` once you do that if I remember correctly.

denismerigoux (approved):
That's great, thanks ! When you're done with the refactoring to remove some of my old code you can merge.

AltGr (approved):
LGTM!

compiler/utils/string_common.ml
```diff
@@ -0,0 +1,51 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+let to_ascii : string -> string = Ubase.from_utf8
+
+let is_uppercase (c : char) : bool =
+  let c = Char.code c in
+  0x41 <= c && c <= 0x5b
```
denismerigoux, line 21: This should be renamed `is_uppercase_ascii`, because it doesn't work with UTF8 chars.

compiler/literate/html.ml
```diff
@@ -75,6 +75,7 @@ let wrap_html
   Format.fprintf fmt
     "<head>\n\
      <style>\n\
+     summary { cursor: pointer; font-style: italic; }\n\
```
denismerigoux, line 78: Hum that styling should rather be put in the `catala-website/` folder, because by putting it here you're choosing this style for all reusers of the HTML produced. Generally CSS styles specified right in the HTML take precedence over styles specified in a separate CSS file.

compiler/literate/html.ml
```diff
@@ -185,45 +186,67 @@ let rec law_structure_to_html
...
     let h_name = Marked.unmark heading.law_heading_name in
+    let complete_headings = parents_headings @ [h_name] in
+    let id = complete_headings |> String.concat "-" |> sanitize_html_href in
     let fmt_details_open fmt () =
       if 2 = h_number then
         Format.fprintf fmt "<details><summary>%s</summary>" h_name
     in
     let fmt_details_close fmt () =
-      if 2 >= h_number then Format.fprintf fmt "</details>"
+      if 2 = h_number then Format.fprintf fmt "</details>"
```
denismerigoux, line 196: So with that, you only enclose items at level 2 of sectionning or below ? It sounds reasonable to me, but there should be a comment detailing the logic of what's being done here.

denismerigoux (approved):
A brilliant PR for a brilliant ending to your internship ! As always I've left a few nitpicks but I feel we should merge this soon.

compiler/literate/html.ml
EmileRolley: Yes, I think this is the best trade-off.

PR #315, opened 2022-08-12
Factorise ASTs (between dcalc and lcalc)

AltGr:


AltGr:
This is a first part but I am doing clean-up on top of it. The clean-up is the easy part :)

denismerigoux:
Thanks! Tell me when it's ready to review

AltGr:
Not sure if I should put all in the same PR, but the next stage of this refactoring can be seen here: https://github.com/CatalaLang/catala/compare/master...AltGr:ast-factorisation2?expand=1

denismerigoux:
Ok I'll review and merge this, then you can file a separate PR.

compiler/utils/var.mli
```diff
@@ -0,0 +1,73 @@
...
+val compare : 'e t -> 'e t -> int
+val eq : 'e t -> 'e t -> bool
+
+val translate : 'e1 t -> 'e2 t
+(** Needed when converting from one AST type to another. See the note of caution
+    on [Bindlib.copy_var]. *)
+
+(** Wrapper over [Set.S] but with a type variable for the AST type parameters.
+    Extend as needed *)
+module Set : sig
```
denismerigoux, line 42: Couldn't that be expressed with `Set.S with type elt = ...` ?

compiler/utils/var.mli
```diff
@@ -0,0 +1,73 @@
...
+  val remove : 'e var -> 'e t -> 'e t
+  val union : 'e t -> 'e t -> 'e t
+  val mem : 'e var -> 'e t -> bool
+  val of_list : 'e var list -> 'e t
+  val elements : 'e t -> 'e var list
+end
+
+(** Wrapper over [Map.S] but with a type variable for the AST type parameters.
+    Extend as needed *)
+module Map : sig
```
denismerigoux, line 57: Couldn't that be expressed with `Map.S with type key = ...` ?

denismerigoux (approved):
All good except perhaps the signature stuff. Thanks!

compiler/utils/var.mli
AltGr: The issue here is that `Set.elt` doesn't allow a type parameter.So we would have to go through a functor instanciation for each of the `'a` in `'a expr`.

All these shenanigans are just to instanciate a `Set.S` with `type 'a t` and matching `type 'a elt`. Unfortunately that forces us to rewrite the whole interface and use some tricks in the implementation…

An alternative, given we want to be able to write polymorphic functions that will handle `'a variables` and use sets or maps internally, would be to use first-class modules and instanciate them each time. But I think that would lead to much harder to read interfaces.

For example, in `dcalc/ast.ml`, we have
```ocaml
val free_vars_expr : 'm marked_expr -> 'm expr Var.Set.t
```
We could make this function return a first-class module instead, buuut:
```ocaml
module type SetValue = sig
  include Set.S
  val ret: elt
end
val free_vars_expr : 'm marked_expr -> (module SetValue with type elt = 'm expr)
```

Starts to feel like OOP at this point…

denismerigoux: Aaaah OK thanks for the explanation. Yeah let's things as they are now.

PR #318, opened 2022-08-16
Follow-up on the AST factoring

AltGr:
After the tedious part comes the cool stuff: reducing the number of code lines!


denismerigoux:
Can I review?

AltGr:
Please do! I'll hold on new commits for now ^^

compiler/dcalc/interpreter.ml
```diff
@@ -17,38 +17,39 @@
 (** Reference interpreter for the default calculus *)
 
 open Utils
+open Shared_ast
 module A = Ast
```
denismerigoux, line 21: We can remove that `A` alias no?

compiler/scopelang/ast.ml
```diff
@@ -114,7 +107,7 @@ and expr =
   | ELit of Dcalc.Ast.lit
   | EAbs of (expr, marked_expr) Bindlib.mbinder * typ Marked.pos list
```
denismerigoux, line 108: Are you planning to factorize Scopelang's exprs later?

compiler/shared_ast/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (name shared_ast)
+ (public_name catala.shared_ast)
+ (flags
+  (:standard -short-paths))
```
denismerigoux, line 5: What's that?

compiler/shared_ast/types.ml
```diff
@@ -146,10 +154,12 @@ type ('a, 't) marked_gexpr = (('a, 't) gexpr, 't) Marked.t
     library, based on higher-order abstract syntax *)
 and ('a, 't) gexpr =
   (* Constructors common to all ASTs *)
```
denismerigoux, line 156: Why is `gexpr` in `types.ml` and not in `expr.ml`?

compiler/shared_ast/types.ml
```diff
@@ -243,42 +257,47 @@ type scope_let_kind =
   | DestructuringSubScopeResults  (** [let s.x = result.x ]**)
   | Assertion  (** [let _ = assert e]*)
 
-type ('expr, 'm) scope_let = {
+type 'e scope_let = {
```
denismerigoux, line 260: Why is `scope_let` not in `scope.ml`?

denismerigoux (approved):
We can merge, but I would be curious to know more about what you envision as the final state of this big refactoring.

compiler/dcalc/interpreter.ml
AltGr: Probably, yes

compiler/shared_ast/dune
AltGr: It tells the OCaml compiler to choose the shortest string when printing types and values. Normally, I don't use it, but `dune` uses tricks to namespace libraries that are not so well hidden behind the scenes, and here with the "explicit toplevel module" library scheme (library `shared_ast` with a `shared_ast.ml` file that defines what to export), messages would become unreadable with `Shared_ast__.Types.` prefixes everywhere.

This is not a real fix (if only namespaces were well integrated into OCaml :roll_eyes: ) but it makes messages more readable. No effect on compilation otherwise.

denismerigoux: Wouahou that's very neat.

compiler/shared_ast/types.ml
AltGr: Ah actuallu `types.ml` is not _about_ Catala types at all, it's just a module _containing_ type definitions. It's a common pattern to define one module with most type definitions, it allows to skip a redundant ml or mli file (since type defs have to be duplicated between ml and mli), and is a more reasonable use of `open` to access types but without polluting the namespace for values.

I agree that in this context the naming confusion is quite unfortunate. Maybe call the module `typedefs`, or `definitions`, or just `t` ? (It's included anyway at the toplevel of `Shared_ast` so technically you shouldn't have to use its name. Except for the quirks on error messages, see about `--short-paths`)

denismerigoux: `Definitions` would be better.

AltGr:
Rebased + added some fixes

PR #320, opened 2022-08-23
Rationalise the tuple / enum types

AltGr:
This will allow to unify with types used earlier in the
pipeline (`Scopelang.Ast.typ`).

It seems cleaner! But some areas may warrant a later clean-up, in particular
handling of options and their types in the backends, or possible name conflicts
of structs/enums with built-in types when printing.

### DIFF
--- a/compiler/dcalc/typing.ml
+++ b/compiler/dcalc/typing.ml
@@ -38,8 +38,10 @@ type unionfind_typ = typ Marked.pos UnionFind.elem
 and typ =
   | TLit of A.typ_lit
   | TArrow of unionfind_typ * unionfind_typ
-  | TTuple of unionfind_typ list * A.StructName.t option
-  | TEnum of unionfind_typ list * A.EnumName.t
+  | TTuple of unionfind_typ list
+  | TStruct of A.StructName.t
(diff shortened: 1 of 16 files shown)


compiler/dcalc/typing.ml
```diff
@@ -490,16 +509,25 @@ and typecheck_expr_top_down
     Bindlib.box @@ unify_and_mark e1 (unionfind_make (TLit TUnit))
   | A.ELit LEmptyError as e1 ->
     Bindlib.box @@ unify_and_mark e1 (unionfind_make (TAny (Any.fresh ())))
-  | A.ETuple (es, s) ->
+  | A.ETuple (es, None) ->
```
denismerigoux, line 512: Why didn't you change to distinct `ETuple` and `EStruct` here too?

denismerigoux (approved):
This is fine to me, you can merge and continue from there.

compiler/dcalc/typing.ml
AltGr: Yes, that'll probably be better — but I'm trying to not do too much at once, these patches are already getting huge :)

PR #306, opened 2022-08-06
Use `dates_calc` library

rmonat:
This is a WIP. Current version 0.0.2 of `dates_calc` is in the process of being published, so I guess the CI should fail currently.

- [x] Test current implementation
- [x] I commented out the `duration / duration` operator, can we remove it?
- [x] Need to add support for first and last day of month
- [x] Finish porting the Z3 backend to dates_calc

PR #328, opened 2022-09-05
Fix nix build for #306

TWal:


denismerigoux:
Cherry-picked into #306

PR #334, opened 2022-09-13
Generalise the typer for the different ASTs

AltGr:
Should now be ready; includes
- generalisation of the typer for expressions, as advertised
- add typing at the scope level in scopelang
- running the typing pass as early as scopelang (and stop there with `catala Typecheck`)
- tweak of the typer to re-use existing type annotations, if any, failing if they don't match
- fixes in various passes to avoid introducing invalid type annotations (using the "typing hole" `TAny` instead)
- run the typer a second time after dcalc, leveraging the change above
- several fixes and re-inforcements to the typer. In particular (see discussion below), fixes an issue due to misguided use of bindlib, and introduced with the type annotations everywhere in the code, and that caused typing checks to be delayed until the expression was deconstructed.

NOTE: this introduces many assertions, that I fixed for all existing tests. If some passes are not properly tested (e.g. some optimisations, compilation without exceptions, etc.) they will be likely to break these — please report them and I'll gladly add the missing tests and fix them.


AltGr:
@denismerigoux I think you will be interested in https://github.com/CatalaLang/catala/pull/334/commits/279303c5aba8d2403164981ace6c1c968441d2df : one mystery solved :)

compiler/shared_ast/typing.ml
```diff
@@ -698,9 +694,9 @@ let infer_types_program prg =
       let rec process_scope_body_expr env = function
         | A.Result e ->
           let e' = wrap ctx (typecheck_expr_bottom_up ctx env) e in
+          wrap ctx (unify ctx e (ty (Bindlib.unbox e'))) ty_out;
```
AltGr, line 697: I am not really sure this is correct though: if some unionfinds ever get generated at `unbox` time, they will be duplicated and we unify on an instance that will get dropped, leaving another instance in the marked AST.

This would also mean that part of the typing work gets done twice

denismerigoux: Remember me why we don't simply let a `StructuredError` generated at unification failure go up to get caught in `driver.ml` ?

AltGr: Hm initially there were difficulties with reporting the proper expressions where the error occured (and not a sub/surrounding one) with `top_down` and `bottom_up` wrapping each other's calls. But indeed that's no longer the case so that could probably be simplified!

This wouldn't, however, solve the issue of typing error getting delayed more than we'd like :thinking:

AltGr:
(Just updated the PR description with up-to-date info, please refer to the top of the page)

compiler/scopelang/scope_to_dcalc.ml
```diff
@@ -655,11 +678,11 @@ let translate_rules
   let return_exp =
     Bindlib.box_apply
       (fun args ->
-        ETuple (args, Some sigma_return_struct_name), pos_mark pos_sigma)
+        ETuple (args, Some sigma_return_struct_name), mark_tany mark pos_sigma)
```
denismerigoux, line 681: So here the Scope -> Dcalc translation does not try to fill in all the types it introduces; rather it relies on a re-typing pass. Right ?

compiler/shared_ast/expr.ml
```diff
@@ -756,3 +718,99 @@ let rec size : type a. (a, 't) gexpr -> int =
...
+
+let make_multiple_let_in xs taus e1s e2 mpos =
+  make_app (make_abs xs e2 taus mpos) e1s (pos (Bindlib.unbox e2))
+
+let make_default exceptions just cons mark =
+  let rec bool_value = function
+    | ELit (LBool b), _ -> Some b
+    | EApp ((EOp (Unop (Log (l, _))), _), [e]), _
+      when l <> PosRecordIfTrueBool
+           (* we don't remove the log calls corresponding to source code
```
denismerigoux, line 789: This is important for logging but is placed at a rather odd position. Could we relocate this bit of logic in the Scopelang -> Dcalc translation?

compiler/shared_ast/print.mli
```diff
@@ -43,9 +43,19 @@ val unop : Format.formatter -> unop -> unit
...
-val naked_expr :
+val expr :
   ?debug:bool (** [true] for debug printing *) ->
   decl_ctx ->
   Format.formatter ->
-  ('a, 't) gexpr ->
+  ('a, 'm mark) gexpr ->
   unit
+
+(** {1 Debugging versions that don't require a context} *)
```
denismerigoux, line 53: I see you got lazy :)

compiler/shared_ast/typing.ml
```diff
@@ -293,55 +333,121 @@ let bmap2 (f : 'a -> 'b -> 'c Bindlib.box) (es : 'a list) (xs : 'b list) :
...
+        Env.get_subscope_var env scope_name (Marked.unmark v)
+    in
+    match ty with
+    | Some ty -> Bindlib.box (mark e1 (ast_to_typ ty))
+    | None ->
+      Errors.raise_spanned_error pos_e "Reference to %a not found"
+        (Expr.format ctx) e)
+  | A.EStruct (s_name, fmap) ->
+    let+ fmap' =
+      (* This assumes that the fields in fmap and the struct type are already
```
denismerigoux, line 374: This could deserve an `assert` ?

compiler/shared_ast/typing.ml
```diff
@@ -456,177 +562,244 @@ let rec typecheck_expr_bottom_up
...
-    Bindlib.box @@ unify_and_mark e1 (unionfind_make (TAny (Any.fresh ())))
-  | A.ETuple (es, None) ->
-    let+ es' = bmap (typecheck_expr_bottom_up ctx env) es in
+      Errors.raise_spanned_error pos_e "Reference to %a not found"
+        (Expr.format ctx) e)
+  | A.EStruct (s_name, fmap) ->
+    unify_and_mark (unionfind_make (TStruct s_name))
+    @@ fun () ->
+    let+ fmap' =
+      (* This assumes that the fields in fmap and the struct type are already
```
denismerigoux, line 607: Ditto above

compiler/surface/parser.messages
```diff
@@ -1741,7 +1741,7 @@ source_file: BEGIN_CODE SCOPE CONSTRUCTOR UNDER_CONDITION TRUE YEAR
 ## small_expression
 ##
 
-expected an enum constructor to test if the expression on the left
+expected the function application operator
```
denismerigoux, line 1744: So the syntax has changed a lot and the update feature from menhir does not create the new error messages that would be needed with the new syntactic constructions. So from time to time I just re-create a new `parser.messages` from scratch with Menhir and git-merge it with the old one ; that way we keep the existing messages and add new ones for the new syntactic constructs. Maybe now is a good time to do that.

french_law/python/src/allocations_familiales.py
```diff
@@ -545,19 +545,19 @@ def smic(smic_in:SmicIn):
     date_courante = smic_in.date_courante_in
     residence = smic_in.residence_in
     try:
-        def temp_brut_horaire(_:Any):
+        def temp_brut_horaire(_:Unit):
             raise EmptyError
-        def temp_brut_horaire_1(_:Any):
+        def temp_brut_horaire_1(_:Unit):
```
denismerigoux, line 550: What we see in those diffs is that the types that get no constraints attached to them during unification (e.g. arguments of thunks) now get resolved to `Any`,whereas before I used to pick `Unit` for those since we can pick any type. I would prefer the old behavior to be reverted (as well as for the OCaml backend but it's the same problem).

tests/README.md
```diff
@@ -49,3 +49,11 @@ for instance), you can mass-reset the expected outputs with
 `make test_suite CLERK_OPTS=--reset`.
 `git diff` will then allow to check all the changes at once.
 **Caution**: It's your responsability to check all the changes before committing them.
+
+## Tips
+
+* Running a single test-file just to check changes when tweaking either the compiler or the test file itself, but without updating or diffing with the reference can be useful when debugging. The following command outputs the result to `stdout` and can be used from within text editors:
```
denismerigoux, line 55: Thanks!

tests/test_io/bad/using_non_output.catala_en
```diff
@@ -14,7 +14,7 @@ scope B:
   assertion a.a = 0
 ```
 ```catala-test-inline
-$ catala Typecheck 
+$ catala Dcalc
```
denismerigoux, line 17: Aaaaaah interesting, now you have to go to `Dcalc` for this sort of errors to appear. For me the `Typechecking` mode of the Catala compiler should run all the checks that would appear if you were to compile it to at least Dcalc and typecheck. So two solutions : either you remap internally the `Typechecking` backend to go to at `Dcalc`, or we add a linting pass at `Scopelang` that checks the invariants needed at later compilation passes.

tests/test_typing/bad/err3.catala_en
```diff
@@ -17,17 +17,17 @@ $ catala Typecheck
...
+10 |   definition a equals number of (z ++ z) / 2
+   |                                          ^
    +
 
 Type integer coming from expression:
   --> tests/test_typing/bad/err3.catala_en
    |
 10 |   definition a equals number of (z ++ z) / 2
-   |                       ^^^^^^
+   |                                          ^
```
denismerigoux, line 30: So generally the type indicating the operator is not the most obvious hint as to what's going wrong in your program. But I guess we can live with it, until we get our OCaml version of https://docs.rs/ariadne/latest/ariadne/index.html and are able to craft magnificent error messages with notes, hints and stuff.

denismerigoux (approved):
Thanks for this massive and excellent PR. I've let a lot of comments, but feel free to delay some of the changes requested to a later PR.

compiler/scopelang/scope_to_dcalc.ml
AltGr: Yes, it's a best-effort: at first I tried to introduce more "manually-computed" types but it was too tedious and time-consuming.

The good approach would be to use "smart constructors" to generate the new terms (the `Expr.make_*` functions). For this specific case I actually did add `Expr.make_tuple` later on, so that could be used here.

But in general, having the new terms with `TAny` and then re-typing to infer their type and checking consistency with both their parents and sub-terms doesn't seem like a bad approach, it has a small speed impact (re-typing is required) but offers the same guarantees. One limitation at the moment is that we don't distinguish between `typed expr` with full typing and `typed expr` that may contain holes ; this could be added at some point but for now it remains manageable.

build_system/clerk_driver.ml
```diff
@@ -314,6 +314,32 @@ let search_for_expected_outputs (file : string) : expected_output_descr list =
...
+end
+
+let pipe_diff_cmd =
+  let open Nj.Expr in
+  let has_patdiff = Sys.command "type patdiff >/dev/null 2>&1" = 0 in
+  if has_patdiff then
+    Seq
+      [
+        Lit "|";
+        Lit "patdiff";
```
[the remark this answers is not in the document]
AltGr: Better/shorter output but it needs to be installed through `opam` and has quite a lot of dependencies so I preferred to keep it optional.
In particular, it highlights just the part of the lines that was modified so it makes it much faster to spot actual changes.
## colordiff
image
## patdiff
image

compiler/dcalc/optimizations.ml
```diff
@@ -248,9 +248,8 @@ let program_map
     (fun new_scopes -> { p with scopes = new_scopes })
     (scopes_map t ctx p.scopes)
 
-let optimize_program (p : 'm program) : untyped program =
+let optimize_program (p : 'm program) : 'm program =
   Bindlib.unbox
     (program_map partial_evaluation
        { var_values = Var.Map.empty; decl_ctx = p.decl_ctx }
        p)
-  |> Program.untype
```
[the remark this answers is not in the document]
AltGr: This one, yes: it only does reductions so that is easy to check. The one in `lcalc`, which I haven't merged yet with this one, I haven't checked yet, but it seems straightforward as well.

compiler/driver.ml
```diff
@@ -171,18 +171,29 @@ let driver source_file (options : Cli.options) : int =
...
+        let _prgm = Scopelang.Ast.type_program prgm in
+        (* That's it! *)
+        Cli.result_print "Typechecking successful!"
+      | ( `Interpret | `OCaml | `Python | `Scalc | `Lcalc | `Dcalc | `Proof
+        | `Plugin _ ) as backend -> (
+        let type_ordering =
+          Scopelang.Dependency.check_type_cycles prgm.program_ctx.ctx_structs
+            prgm.program_ctx.ctx_enums
         in
+        let prgm = Scopelang.Ast.type_program prgm in
```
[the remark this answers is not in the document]
AltGr: Indeed, well spotted

french_law/python/src/allocations_familiales.py
AltGr: The change is in the other way, a quirk made it remain `Any` and I fixed it to be restricted to `Unit` :)

tests/test_io/bad/using_non_output.catala_en
AltGr: Duplicating the check seems fragile (at the moment its code is tightly bound to the →dcalc transformation)… but that could make sense if we have a whole family of sanity-checks to run at the same time. I already displaced the type cycle detection because it's not reasonable to run typing before that.

I don't see a good reason not to have `catala Typecheck` run the dcalc conversion just for validation: it feels a bit weird, but probably only to the compiler implementers ;)

tests/test_typing/bad/err3.catala_en
AltGr: I agree. I must admit I have been turning some knobs to tweak the error messages to better positions, but we're not completely there yet. It's also a notoriously difficult problem where you have to account for how people think :)

Here the type mismatch comes from the return type of the operator: the bug is that we should actually mark the whole application and not just the operator (the `/` expression is *not* actually of type int)! I didn't yet understand exactly why, but I'd argue that marking `/` already makes more sense than marking `number`.

Still, I am glad that the expression marked as the source of the error is now the actual expression and not a part of the declaration in the metadata section!

compiler/surface/parser.messages
AltGr: Ah, I understand! Thanks for the hint.

compiler/lcalc/compile_with_exceptions.ml
```diff
@@ -23,10 +23,11 @@ type 'm ctx = ('m D.expr, 'm A.expr Var.t) Var.Map.t
 (** This environment contains a mapping between the variables in Dcalc and their
     correspondance in Lcalc. *)
 
-let thunk_expr (e : 'm A.expr Bindlib.box) (mark : 'm mark) :
-    'm A.expr Bindlib.box =
+let thunk_expr (type m) (e : m A.expr Bindlib.box) : m A.expr Bindlib.box =
   let dummy_var = Var.make "_" in
-  Expr.make_abs [| dummy_var |] e [TAny, Expr.mark_pos mark] mark
+  let pos = Expr.pos (Bindlib.unbox e) in
```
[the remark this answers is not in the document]
AltGr: You are right to raise a flag about that.

I haven't witnessed the perf change myself, but I am not surprised by this… and I have already been looking for a better solution to these multiple `unbox`.

Basically, bindlib stores boxed terms as a base definition plus a stack of closures added every time you run `Bindlib.apply_box` or its variants. `Bindlib.unbox` is the `force` function that will compute fresh variables and apply all of these, rebuilding the whole AST. This is also what caused the unfortunate "delayed type error" bug.

Now, it's clear that if you run `unbox` to throw away the term during a recursive traversal, you are actually doing a useless quadratic traversal of the term, possibly re-doing stacked computations as you go, like type inference :scream:.

This is all the more frustrating as the position information is completely independent from variable bindings ; this specific case is not in a recursive function, but remains distasteful.

tests/test_typing/bad/err3.catala_en
AltGr: After more detailed investigation, this comes from the fact that unification is done on the arrow type: we can build that type from the outer information and the arguments, then do a single unification. It makes sense then that this points to the operator. This would likely be non-trivial to change…

compiler/lcalc/compile_with_exceptions.ml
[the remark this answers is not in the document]
AltGr: This can be merged as is I think; I am studying a more general solution to this issue.

PR #335, opened 2022-09-17
gedit syntax highlighting and Chinese surface language

YueLiPicasso:
Hello,

 This PR contains two parts:
-  gedit syntax highlighting support for Chinese/English Catala, and 
-  Extension of the lexer to support Chinese surface language.

The performance of the syntax highlighter can be assessed by opening any existing `.catala_en` or `.catala_cn` (in Chinese) file with gedit (a Chinese Catala file can be found in `catala/examples/tutorial_cn`). The `INSTALL.md` is updated with instructions for enabling gedit syntax highlighting.

Regarding the lexer extension to support Chinese surface language,  I have followed all the instructions and hints in `CONTRIBUTION.md` and produced a compilable `compiler/surface/lexer_cn.cppo.ml `.  The current state of the work is:
- The design of the Chinese Catala surface syntax is completed with the help of the gedit syntax-highlighter. The Chinese keywords and their correspondence with the English keywords and tokens are made clear in `catala/syntax_highlighting/en/gedit/catala.lang`.  
- Strings of UTF-8 encoding of the Chinese keywords and their Sedlex regex counterparts have been entered into a dedicated file `compiler/surface/lexer_cn.cppo.ml`
- Relevant modifications of other parts of the Catala source as suggested by "non-exhaustive patterns" warnings.  

 However, a puzzling parsing error occcured during interpreting the Chinese Catala file and I do not know how to resolve that. Reproduce the error by:
 ```
 cd catala
 make dependencies
 make build
 opam install ./
 catala Interpret --scope=测试3 examples/tutorial_cn/tutorial_cn.catala_cn
 ```
I tried to follow the hint and run the `ocamlc` command but it did not help.  
 
 ```
ocamlc _build/default/compiler/surface/lexer_cn.ml
File "lexer.cppo.ml", line 15, characters 5-11:
Error: Unbound module Tokens
```

### CODE REVIEW

**Comment by denismerigoux** [comment]
Thanks @YueLiPicasso! Concerning your error it should be caused by the fact that doing `opam install ./` installs the official latest version of Catala (0.7.0) from the git tag it is associated with, and not from the file tree in the repository containing your latest changes. 

To execute the Catala compiler with your changes, you should try:

```
dune exec catala -- Interpret --scope=测试3 examples/tutorial_cn/tutorial_cn.catala_cn
```

which currently yields the following syntax error:

```
[ERROR] Parsing error after token " ": what comes after is unknown

 --> examples/tutorial_cn/tutorial_cn.catala_cn
  | 
8 | 声明 结构 个人：
  |      ^
  + 
```

**Comment by YueLiPicasso** [comment]
Hello Denis,

I checked the auto-generated `lexer_cn.ml` file, and identified the source of problem. Below is a discussion.

> The lexer cannot recognize arbitrary (non-keyword) Chinese character strings in Catala source, which are supposed to be identifiers or constructors. 

The relevant lines from `lexer_cn.ml` are copied below.
```
| uppercase, Star (uppercase | lowercase | digit | '_' | '\'') ->
      (* Name of constructor *)
      L.update_acc lexbuf;
      CONSTRUCTOR (Utf8.lexeme lexbuf)
  | lowercase, Star (lowercase | uppercase | digit | '_' | '\'') ->
      (* Name of variable *)
      L.update_acc lexbuf;
      IDENT (Utf8.lexeme lexbuf)
``` 
# Analysis

For languages like English, French and Polish, there is a notion of "uppercase" and "lowercase" characters. But for Chinese (and for Japanese and Korean as well,  known as the CJK languages) there is _not_ a notion of cases. Then code above has no way of recognising CJK identifiers and constructors.  To further demonstrate the issue, I modified the line quoted by the error message
```
<modify from:> 声明 结构 个人 
<to:> 声明 结构 U个人

/* Here I use C-style meta-comments to indicate the nature of the Chinese characters */
/*声明DECLARATION   结构STRUCT 个人CONSTRUCTOR/Individual */
```
That is, I prefixed the Chinese constructor with an English uppercase U. This time the error message clearly pointed out:
```
[ERROR] Parsing error after token "U": what comes after is unknown
--> examples/tutorial_cn/tutorial_cn.catala_cn
  | 
8 | 声明 结构 U个人：
  |       ^
  + 
``` 

# Solution Ideas

The first attempt would be modifying the `lexer_cn.ml` lines by adding some CJK character classes `cjk`
```
| (cjk | uppercase), Star (cjk | uppercase | lowercase | digit | '_' | '\'') ->
      (* Name of constructor *)
      L.update_acc lexbuf;
      CONSTRUCTOR (Utf8.lexeme lexbuf)
  | ( cjk | lowercase), Star ( cjk | lowercase | uppercase | digit | '_' | '\'') ->
      (* Name of variable *)
      L.update_acc lexbuf;
      IDENT (Utf8.lexeme lexbuf)
``` 
Two further problems arise:
* `lexer_cn.ml` shall not be hand-coded. Rather, the code that generates `lexer_cn.ml` shall be extended with a rule catering to the the need of CJK identifiers. 
* There is no longer a notion of upper/lower-cases to help distinguishing IDENT from CONSTRUCTOR. Rather they should all be regarded as of the same token type IDENTIFIER, and context information is needed to further distinguish them.

YueLiPicasso:
# Closing Remarks

There is not a straight forward solution regarding parsing Chinese (and the like, caseless) identifiers. At least the following things need to be done:
- Add a Sedlex predefined regexp class that matches all Chinese unicode codepoints. 
- Modify the lexer and perhaps the parser to accept arbitrary Chinese strings as identifiers and distinguish IDENT from CONSTRUCTOR.

Otherwise mixing Chinese keywords with English/Latin identifiers would be a cheap solution not complying with the spirit of literate programming. 


For now I would close this PR, and open a new PR instead only for the gedit syntax highlighting plug-in.

AltGr:
Hm you have a point here, indeed we took the same convention as OCaml for distinguishing uppercase/lowercase idents because it's convenient, but it doesn't scale to non-latin languages!

Although a little more verbose, there is an easy solution: use some prefix char to distinguish one from the other, like most script languages do.
E.g. instead of `Foo` and `bar`, we would write `#Foo` and `bar`.
That seems cleaner to me than trying to match character classes among languages — you may have an OK hack for Chinese, but what about Japanese, or Arabic… ?

Now if we decide on a convention here, we may only use it for non-latin character sets and keep it optional for latin, or for consistency use the same for every language. This is at least worth discussing… the bigger downside is that these prefix chars may make the language look more arcane to non-developers.

PR #336, opened 2022-09-20
gnu gedit syntax highlighting

YueLiPicasso:
... for English and (hypothetically) Chinese.


denismerigoux:
Thanks Yue for splitting your PR. I've removed the Chinese annotations from the gedit plugin ; we can put them back once we have figured out a way to support Chinese in the parser with this `IDENT`/`CONSTRUCTOR` distinction (#335).

PR #337, opened 2022-09-20
Clerk: don't default to infinite ninja jobs

AltGr:
Hopefully this may fix the CI OOM issues


AltGr:
The fix seems to work!

PR #340, opened 2022-09-23
Change the clerk tests to use inline output

AltGr:
NOTE: this keeps the functionality for tests in separate output/... files, but
it's not used anymore.


tests/README.md
```diff
@@ -6,50 +6,46 @@ language.
...
 ## Workflow for adding new tests
 
 1. Create a new test file in `foo/{good,bad}/bar.catala_<language>` (pick the right directory and
    an informative name for your test)
 2. Write your test, and pick a toplevel scope `A` to run.
 3. Add the following section to your file:
       ~~~markdown
-      ```catala-test {id="A.Interpret"}
+      ```catala-test-inline
       catala Interpret -s A
```
denismerigoux, line 23: Missing `$`

tests/README.md
```diff
@@ -6,50 +6,46 @@ language.
...
       ```
       ~~~
-4. Run `CLERK_OPTS=--reset make tests/foo/{good,bad}/bar.catala_<language>`
-   from the root of the Catala repository. This
-   will record the content of the output of your test into the target file.
-5. Check that the expected output is present in `tests/foo/{good,bad}/output/bar.catala_<language>.A.Interpret`
-6. Don't forget to `git add` both the test file and the output file.
+4. Run `make tests/foo/{good,bad}/bar.catala_<language> CLERK_OPTS=--reset`
+   from the root of the Catala repository. This will update the test with the
+   actual output of the catala command.
```
denismerigoux, line 28: Did you add that reset feature also ? It's VERY useful and I don't want us to lose it.

denismerigoux (approved):
Thanks very much! Before merging:

* could you reformat the test files to get rid of trailing whitespaces?

tests/README.md
AltGr: yup of course! :)
To ba fair I most often do it this way and then use my git tools to review the diffs

tests/README.md
AltGr: Oops well spotted!

AltGr:
> could you reformat the test files to get rid of trailing whitespaces?

Hm I could trim it from the command outputs… but shouldn't we rather tweak the printer so that it doesn't generate it in the first place ? :grimacing:

denismerigoux:
I don't know where they come from, only that if I open your new test files with my vscode with auto trailing whitespaces removal it puts a lot of changes everywhere. But yeah we should rather tweak th printer (although I don't know why the printer would print trailing whitespaces).

PR #341, opened 2022-09-23
Fix & improve typing messages

AltGr:
This attempts to improve a few error messages, and fixes a bug where typing was
done unexpectedly lazily.

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -885,7 +885,7 @@ let driver
                 (add_root_test_build ninja ctx.all_file_names
                    ctx.all_test_builds));
           let ninja_cmd =
-            "ninja -f " ^ ninja_output ^ " " ^ ninja_flags ^ " test"
+            "ninja -k 0 -f " ^ ninja_output ^ " " ^ ninja_flags ^ " test"
           in
           Cli.debug_print "executing '%s'..." ninja_cmd;
           let return = Sys.command ninja_cmd in
--- a/compiler/desugared/ast.ml
+++ b/compiler/desugared/ast.ml
@@ -40,7 +40,7 @@ module LabelSet : Set.S with type elt = LabelName.t = Set.Make (LabelName)
 module ScopeDef = struct
   type t =
     | Var of ScopeVar.t * StateName.t option
-    | SubScopeVar of SubScopeName.t * ScopeVar.t
+    | SubScopeVar of SubScopeName.t * ScopeVar.t * Pos.t
         (** In this case, the [ScopeVar.t] lives inside the context of the
             subscope's original declaration *)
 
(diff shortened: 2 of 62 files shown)


AltGr:
NOTE: is on top of #340 and should be rebased once that is merged.

compiler/desugared/ast.ml
```diff
@@ -40,7 +40,7 @@ module LabelSet : Set.S with type elt = LabelName.t = Set.Make (LabelName)
 module ScopeDef = struct
   type t =
     | Var of ScopeVar.t * StateName.t option
-    | SubScopeVar of SubScopeName.t * ScopeVar.t
+    | SubScopeVar of SubScopeName.t * ScopeVar.t * Pos.t
```
denismerigoux, line 43: To what does the `Pos.t` points to ? And why only for `SubScopeVar` and not for `Var` ?

compiler/desugared/ast.ml
```diff
@@ -49,35 +49,36 @@ module ScopeDef = struct
...
+    | SubScopeVar (x', x, _), SubScopeVar (y', y, _) ->
       let cmp = SubScopeName.compare x' y' in
       if cmp = 0 then ScopeVar.compare x y else cmp
 
   let get_position x =
     match x with
     | Var (x, None) -> Marked.get_mark (ScopeVar.get_info x)
     | Var (_, Some sx) -> Marked.get_mark (StateName.get_info sx)
-    | SubScopeVar (x, _) -> Marked.get_mark (SubScopeName.get_info x)
+    | SubScopeVar (_, _, pos) -> pos
```
denismerigoux, line 66: Ok I see it's for error messages.

tests/test_typing/bad/err2.catala_en
```diff
@@ -23,7 +23,7 @@ Error coming from typechecking the following expression:
    |                                  ^^^^^^^^
    + 
 
-Type Structure array coming from expression:
+Type any[10] array coming from expression:
```
denismerigoux, line 26: So this is not an improvement. The union find representative chosen should not be a type variable if there is a non-type-variable representative.

tests/test_typing/bad/err3.catala_en
```diff
@@ -12,5 +12,61 @@ scope S:
...
+Type decimal coming from expression:
+  --> tests/test_typing/bad/common.catala_en
+   | 
+15 |   output a content decimal
+   |                    ^^^^^^^
+   + 
+#return code 255#
+```
+
+A bug with delayed typechecking can make the above accept the error while it is
```
denismerigoux, line 42: What is that ?

denismerigoux (commented):
Some things have improved indeed but type variables appearing in error messages is a no-go for me

compiler/desugared/ast.ml
AltGr: `ScopeVar.t` / `StateName.t` already contain a position; `SubScopeName.t` as well. But in this case, the position of the dot `subscope.var` can be different from both the above.

tests/test_typing/bad/err3.catala_en
AltGr: Just a comment on why the test is needed. Should it rather be in a ``catala block ?

AltGr:
@denismerigoux a41de29293b should make the message better, ready for review.

build_system/clerk_driver.ml
```diff
@@ -885,7 +885,7 @@ let driver
                 (add_root_test_build ninja ctx.all_file_names
                    ctx.all_test_builds));
           let ninja_cmd =
-            "ninja -f " ^ ninja_output ^ " " ^ ninja_flags ^ " test"
+            "ninja -k 0 -f " ^ ninja_output ^ " " ^ ninja_flags ^ " test"
```
denismerigoux, line 888: Doesn't this reintroduced what was fixed by #337  ?

denismerigoux (approved):
Thanks! Sounds good except for this `err3` delayed thing which looks bad to me.

tests/test_typing/bad/err3.catala_en
AltGr: Here we get the same error twice, which is correct behaviour. If you look at the sate before the fix the test above answers "typing successful": https://github.com/CatalaLang/catala/pull/341/commits/498429e4b7765634c05e538dc96e28524e6e1410#diff-bde130b0c71ac3cb3a70fd86df22d09dc60f2b704391cf2c7bf2fb09c5613dfbL15

AltGr:
> Thanks! Sounds good except for this err3 delayed thing which looks bad to me.

Actually this fixes it, I introduced that regression when I added the type annotations in the AST I guess.

PR #346, opened 2022-10-07
Swap boxing and annotations in expressions

AltGr:
This was the only reasonable solution I found to the issue raised
[here](https://github.com/CatalaLang/catala/pull/334#discussion_r987175884).

This was a pretty tedious rewrite, but it should now ensure we are doing
things correctly. As a bonus, the "smart" expression constructors are now
used everywhere to build expressions (so another refactoring like this one
should be much easier) and this makes the code overall feel more
straightforward (`Bindlib.box_apply` or `let+` no longer need to be
visible!)

---

Basically, we were using values of type `gexpr box = naked_gexpr marked
box` throughout when (re-)building expressions. This was done 99% of the
time by using `Bindlib.box_apply add_mark naked_e` right after building
`naked_e`. In lots of places, we needed to recover the annotation of this
expression later on, typically to build its parent term (to inherit the
position, or build the type).

Since it wasn't always possible to wrap these uses within `box_apply` (esp.
as bindlib boxes aren't a monad), here and there we had to call
`Bindlib.unbox`, just to recover the position or type. This had the very
unpleasant effect of forcing the resolution of the whole box (including
applying any stored closures) to reach the top-level annotation which isn't
even dependant on specific variable bindings. Then, generally, throwing
away the result.

Therefore, the change proposed here transforms
- `naked_gexpr marked Bindlib.box` into
- `naked_gexpr Bindlib.box marked` (aliased to `boxed_gexpr` or `gexpr
boxed` for convenience)

This means only 1. not fitting the mark into the box right away when
building, and 2. accessing the top-level mark directly without unboxing

The functions for building terms from module `Shared_ast.Expr` could be
changed easily. But then they needed to be consistently used throughout,
without manually building terms through `Bindlib.apply_box` -- which covers
most of the changes in this patch.

`Expr.Box.inj` is provided to swap back to a box, before binding for
example.

Additionally, this gives a 40% speedup on `make -C examples
pass_all_tests`, which hints at the amount of unnecessary work we were
doing --'


denismerigoux (approved):
Amazing, thanks ! I have no comments, everything is perfect :)

PR #351, opened 2022-10-27
Fancier error messages

AltGr:
(this is low priority, but since I have it laying around better post it)

* Adds parseable positions to error messages (handy to make them clickable in emacs¹)
* Avoids trailing whitespaces
* Uses some fancy unicode chars for outlines

### DIFF
--- a/compiler/utils/pos.ml
+++ b/compiler/utils/pos.ml
@@ -179,18 +179,18 @@ let retrieve_loc_text (pos : t) : string =
                && cur_line <= sline + (2 * (eline - sline))
                && cur_line mod 2 = sline mod 2
              then
-               Cli.with_style blue_style "%*d | " spaces
+               Cli.with_style blue_style "%*d |" spaces
                  (sline + ((cur_line - sline) / 2))
              else if cur_line >= sline - include_extra_count && cur_line < sline
-             then Cli.with_style blue_style "%*d | " spaces cur_line
+             then Cli.with_style blue_style "%*d |" spaces (cur_line + 1)
              else if
                cur_line
                <= sline + (2 * (eline - sline)) + 1 + include_extra_count
                && cur_line > sline + (2 * (eline - sline)) + 1
              then
-               Cli.with_style blue_style "%*d | " spaces
+               Cli.with_style blue_style "%*d |" spaces
                  (cur_line - (eline - sline + 1))
-             else Cli.with_style blue_style "%*s | " spaces ""))
+             else Cli.with_style blue_style "%*s |" spaces ""))
         (Cli.add_prefix_to_each_line
            (Printf.sprintf "%s"
               (String.concat "\n"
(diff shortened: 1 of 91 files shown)


compiler/utils/pos.ml
```diff
@@ -125,24 +133,26 @@ let retrieve_loc_text (pos : t) : string =
...
-            Cli.with_style error_indicator_style "%*s"
-              (String.length line - 1)
-              (String.make
+            Cli.with_style error_indicator_style "%*s%s"
+              (get_start_column pos - 1)
+              ""
+              (string_repeat
                  (max (String.length line - get_start_column pos) 0)
-                 '^')
+                 "‾")
```
denismerigoux, line 148: Ah! I didn't know this style of error messages. What compiler does it try to mimic?

tests/test_arithmetic/bad/division_by_zero.catala_en
```diff
@@ -80,19 +80,19 @@ $ catala Interpret -s Money
...
-   |
-30 |   definition i equals $10.0 /$ $0.0
-   |                       ^^^^^^^^^^^^^
-   + `Division_by_zero` exception management
-   +-+ with money
+  --> tests/test_arithmetic/bad/division_by_zero.catala_en:30:22-35
+   │
+30 │   definition i equals $10.0 /$ $0.0
+   │                       ‾‾‾‾‾‾‾‾‾‾‾‾‾
+   └┬ `Division_by_zero` exception management
```
denismerigoux, line 87: Wouahou ! I like it <3

examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
```diff
@@ -33,18 +33,18 @@ Message: unexpected token
 Autosuggestion: did you mean "content", or maybe "condition"?
 
 Error token:
-  --> examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
```
denismerigoux, line 36: Can you also stylise this arrow ?

denismerigoux (approved):
Thanks ! I really wish we could extract Catala's error messages into a proper library and publish it on opam. But I fear it might turn into a time sink hole... As for style purposes I would suggest we follow https://docs.rs/ariadne/0.1.3/ariadne/, what do you think?

AltGr:
> As for style purposes I would suggest we follow https://docs.rs/ariadne/0.1.3/ariadne/, what do you think?

Yeah these are beautiful! This PR is just a quick replace of a few chars, this will be much more work obviously :)

examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
AltGr: I had trouble finding a similar unicode long arrow (or a hline that is always aligned with the normal → arrow). But I'll copy what ariadne does ^^

compiler/utils/pos.ml
AltGr: Nothing specific, but this is relatively standard

This is as much as I could find for standard: https://github.com/typester/emacs/blob/master/etc/compilation.txt#L203

So maybe a better choice could be a `.` instead of the `.`:
- single line `file:line.col-col:`
- multiple lines `file:line.col-line.col:

this might give us a better chance for editors to automatically highlight them

denismerigoux: Ah no I was referring to the `‾` instead of `^`

AltGr: Ah sorry! It doesn't mimic anything specific, it's just an unicode underline :shrug:

PR #353, opened 2022-11-08
Add support for assertions in verification backend

### LINKED ISSUE #344: Add assertions in the context of VC generation for Z3
## The problem

Right now we have a lot of false positives raised by the Z3 verification backend, typically for `EmptyError` verification conditions. Consider the following program:

~~~markdown
```catala
declaration scope Foo:
   input current_date content date
   output x content integer

scope Foo:
   definition x under condition
      current_date >=@ |2020-01-01| and current_date <@ |2020-12-31|
   consequence equals
      1

   definition x under condition
      current_date >=@ |2021-01-01| and current_date <@ |2021-12-31|
   consequence equals
      2

   definition x under condition
      current_date >=@ |2022-01-01| and current_date <@ |2022-12-31|
   consequence equals
      3
```
~~~

Then we get:

```
[ERROR] [Foo.x] This variable might return an empty error:
 --> test.catala_en
  | 
4 |    output x content integer
  |           ^
  + 
The solver generated the following counterexample to explain the faulty behavior:
--> current_date : 1900-01-01
```

But of course we don't care about dates from 1900 ; this is a false positive from the proof backend that we want to remove by adding user-specified constraints to the value `current_date`.

## A solution 

We could add those constraints as runtime assertions :

~~~markdown
```catala
declaration scope Foo:
   input current_date content date
   output x content integer

scope Foo:
   assertion current_date >=@ |2020-01-01| and current_date <@ |2022-12-31|

   definition x under condition
      current_date >=@ |2020-01-01| and current_date <@ |2020-12-31|
   consequence equals
      1

   definition x under condition
      current_date >=@ |2021-01-01| and current_date <@ |2021-12-31|
   consequence equals
      2

   definition x under condition
      current_date >=@ |2022-01-01| and current_date <@ |2022-12-31|
   consequence equals
      3
```
~~~

With this information, the verification backend should not report that `x` can trigger an `EmptyError`. Hence, assertions on scope variables should be encoded in the context of the solver.

R1kM:
This PR adds support for top-level assertions in the Catala proof backend, and thus fixes #344 .

To do so, it does the following changes:
* It extends the verification_condition type with a field called vc_asserts, capturing all asserts in the scope of this specific VC
* When creating VCs for a scope, it collects all top-level assertions in this scope and adds them to the vc_asserts field of the corresponding VCs.
* Leveraging existing support for hypotheses in the Z3 backend, it then folds vc_asserts in one expression, and encodes it in the Z3 VC.
* It also adds a few small examples of this behavior in test_proof/


compiler/verification/conditions.ml
```diff
@@ -407,10 +439,15 @@ let rec generate_verification_conditions_scopes
...
         in
-        let _, vcs =
+        let _, vcs, asserts =
           generate_verification_conditions_scope_body_expr ctx scope_body_expr
         in
-        vcs
+        let combined_assert =
+          conjunction_exprs asserts
+            (Typed
+               { pos = Pos.no_pos; ty = Marked.mark Pos.no_pos (TLit TBool) })
```
R1kM, line 448: There is probably a better way of doing this, but I'm not sure how

denismerigoux: Looks nice to me actually. The only thing I might be afraid of is that if this big assert expr is duplicated in all Z3 queries then it will also carry asserts with variables that have nothing to do with the current query. Not sure if it's a big problem.

denismerigoux (commented):
That looks great ! However I have a `Not_found` when running

```
$ dune exec catala -- Proof examples/aides_logement/aides_logement.catala_fr 
catala: internal error, uncaught exception:
        Not_found
        Raised at Stdlib__Map.Make.find in file "map.ml", line 137, characters 10-25
        Called from Verification__Z3backend.translate_expr in file "compiler/verification/z3backend.real.ml", line 653, characters 14-40
        Called from Verification__Z3backend.translate_op in file "compiler/verification/z3backend.real.ml", line 516, characters 24-45
        Called from Verification__Z3backend.translate_op in file "compiler/verification/z3backend.real.ml", line 516, characters 24-45
        Called from Verification__Z3backend.Backend.translate_expr in file "compiler/verification/z3backend.real.ml" (inlined), line 814, characters 4-24
        Called from Verification__Z3backend.Backend.encode_asserts in file "compiler/verification/z3backend.real.ml", line 817, characters 18-38
        Called from Verification__Solver.solve_vc.(fun) in file "compiler/verification/solver.ml", line 35, characters 14-70
```

compiler/verification/conditions.ml
R1kM: Right, I think there is room for improvement and further optimization of the queries. However, in the current state, the asserts should remain small terms, and only be added in the scopes corresponding to the definition. I think this is fine for now, happy to revisit if we see any issue

R1kM:
Thanks @denismerigoux for the heads-up. The issue was that the types of free variables appearing only in the assertions were not collected, leading to the failure you observed.
Since type information is now available directly in the AST, I refactored the verification backend in #355, which now avoids this issue.
Running the verification backend on the aides_logement now "succeeds", with many failures due to some language features not supported.

PR #355, opened 2022-11-16
Do not duplicate the tracking of types of free variables in verification backend

R1kM:
To encode verification conditions to Z3, the verification backend requires the type of free variables in the VCs.
These types were previously collected during VC generation in a map stored as a field in the VC.
However, since #272, all sub-expressions in the AST contain type information, which make this map redundant.

This PR cleans up the verification backend to remove the need for this map and the corresponding ctx_var map in the backend context.


denismerigoux (approved):
Yeah, code simplification! Thanks!

PR #356, opened 2022-11-17
AST refactoring

### LINKED ISSUE #330: Discussion for AST simplification
The summer has seen a thourough code reorganisation, with the unification of shared code and types for different passes under the `shared_ast` sub-library.

However, the AST structures themselves have been left unchanged, if defined in a different way (one thing at a time!) . For example, while the type for tuples and structures is now separated, but the underlying expressions still fall into the same `EStruct` case with an option for the structure name.

This is to open a thread for discussion on the changes that we want or not before actually implementing the changes. I'll write different aspects in different comments.

AltGr:
Many changes got bundled in here and would be too tedious to separate.

Closes #330

See changes in `shared_ast/definitions.ml` to check the main point.

- the biggest change is a modification of the struct and enum types in
 expressions: they are now stored as `Map`s throughout passes, and no
longer
 converted to indexed lists after scopelang. Their accessors are also
changed,
 and tuples only exist in Lcalc (they're used for closure conversion).

  This implied adding some more information in the contexts, to keep the
mapping
 between struct fields and scope output variables. It should also be much
more
 robust (no longer relying on assumptions upon different orderings).

- another very pervasive change is more cosmetic: the rewrite of the main
AST to
 use inline records, labelling individual subfields.

- moved the checks for correct definitions and accesses of structures from
 `Scope_to_dcalc` to `Typing`

- defining some new shallow iterators in module `Shared_ast.Expr`, and
 factorising a few same-pass rewriting functions accordingly (closure
 conversion, optimisations, etc.)

- some smaller style improvements (ensuring we use the proper compare/equal
 functions instead of `=` in a few `when` closes, for example)

### DIFF
--- a/compiler/shared_ast/expr.mli
+++ b/compiler/shared_ast/expr.mli
@@ -204,7 +204,18 @@ val shallow_fold :
   (('a, 't) gexpr -> 'acc -> 'acc) -> ('a, 't) gexpr -> 'acc -> 'acc
 (** Applies a function on all sub-terms of the given expression. Does not
     recurse, and doesn't open binders. Useful as helper for recursive calls
-    within traversal functions *)
+    within traversal functions. This can be used to compute free variables with
+    e.g.:
+
+    {[
+      let rec free_vars = function
+        | EVar v, _ -> Var.Set.singleton v
+        | EAbs { binder; _ }, _ ->
+          let vs, body = Bindlib.unmbind binder in
+          Array.fold_right Var.Set.remove vs (free_vars body)
+        | e ->
+          shallow_fold (fun e -> Var.Set.union (free_vars e)) e Var.Set.empty
+    ]} *)
 
 val map_gather :
   acc:'acc ->
@@ -215,7 +226,20 @@ val map_gather :
 (** Shallow mapping similar to [map], but additionally allows to gather an
     accumulator bottom-up. [acc] is the accumulator value returned on terminal
     nodes, and [join] is used to merge accumulators from the different sub-terms
-    of an expression. [acc] is assumed to be a neutral element for [join] *)
+    of an expression. [acc] is assumed to be a neutral element for [join].
+    Typically used with a set of variables used in the rewrite:
+
+    {[
+      let rec rewrite e =
+        match Marked.unmark e with
+        | Specific_case ->
+          Var.Set.singleton x, some_rewrite_fun e
+        | _ ->
+          Expr.map_gather ~acc:Var.Set.empty ~join:Var.Set.union ~f:rewrite e
+    }]
+
+
+    See [Lcalc.closure_conversion] for a real-world example. *)
 
 (** {2 Expression building helpers} *)
 


compiler/scalc/to_python.ml
```diff
@@ -461,40 +463,41 @@ let format_ctx
       struct_name
       (Format.pp_print_list
          ~pp_sep:(fun fmt () -> Format.fprintf fmt ", ")
-         (fun _fmt (struct_field, struct_field_type) ->
+         (fun fmt (struct_field, struct_field_type) ->
```
denismerigoux, line 466: wouahou good catch

compiler/scopelang/scope_to_dcalc.ml
```diff
@@ -531,25 +500,31 @@ let translate_rule
...
+                 subscope_sig.scope_sig_in_fields)
+          in
+          StructFieldMap.add field e acc)
+        StructFieldMap.empty all_subscope_input_vars
     in
     let subscope_struct_arg =
-      (* FIXME: this is very fragile: we assume that the ordering of the scope
-         variables is the same as the ordering of the struct fields. *)
-      Expr.etuple subscope_args (Some called_scope_input_struct)
+      Expr.estruct called_scope_input_struct subscope_args
```
denismerigoux, line 527: YEAH

compiler/verification/z3backend.real.ml
```diff
@@ -313,11 +314,17 @@ and find_or_create_enum (ctx : context) (enum : EnumName.t) :
...
+        (fun ctr ty (ctx, ctrs) ->
+          let ctx, ctr = create_constructor ctr ty ctx in
+          ctx, ctr :: ctrs)
+        ctrs (ctx, [])
+    in
     let z3_enum =
       Datatype.mk_sort_s ctx.ctx_z3
         (Marked.unmark (EnumName.get_info enum))
-        z3_ctrs
+        (List.rev z3_ctrs)
```
denismerigoux, line 327: `fold_left` vs `fold_right` semantics for `EnumConstructorMap.fold` ?

tests/test_array/good/aggregation_2.catala_en
```diff
@@ -29,13 +29,13 @@ scope B:
 $ catala Interpret -s A
 [RESULT] Computation successful! Results:
 [RESULT] x =
-           [S {"id"= 0; "income"= $0.00}; S {"id"= 1; "income"= $9.00};
-             S {"id"= 2; "income"= $5.20}]
+           [ S { "id"= 0; "income"= $0.00 };  S { "id"= 1; "income"= $9.00 };
```
denismerigoux, line 32: Why the extra space here ?

denismerigoux: especially before the struct name

denismerigoux (approved):
Thanks again for this massive PR!

denismerigoux:
Before merging, can you `make format` the generated OCaml and Python files in `french_law` ?

compiler/verification/z3backend.real.ml
AltGr: Yeah this one actually took me a little extra time to debug :/
`Set`/`Map` always fold from the left, and don't even provide reverse fold functions (although they wouldn't be any more expensive...) ; so the fold actually returns a reversed list.

Best solution would probably be to have a `Map.fold_map` function that already preserves order...

tests/test_array/good/aggregation_2.catala_en
AltGr: Good remark, that's suspicious indeed, I'll see how the printer got changed.

AltGr:
> Before merging, can you make format the generated OCaml and Python files in french_law ?

hm I am not sure what you mean, we disabled reformatting of generated files a while ago :grimacing:

PR #358, opened 2022-11-18
Uniformise naming of conversion modules across compilation passes

AltGr:
This is just a proposal for an administrative renaming across libraries encoding the passes of the compiler. There should be no functional changes.

Basically, each pass defines its library with an `Ast` module, but translations used to be defined either as `from_*` for some (`Lcalc.From_dcalc`) and `to_*` for some others (`Scopelang.To_dcalc`). Besides, naming was diverse (`Desugared.Desugaring.desugar_program`).

Now that we are less constrained by mutual dependencies (common definitions being gatherd in the `Shared_ast` lib), this patch proposes to uniformise all passes with `PassB.From_passA.translate_program` modules and functions. It also implies that every pass only depends on the previous one, which results in a more pleasant dependency order.

~~Note: based on top of #356 for now, see [the actual diff](https://github.com/AltGr/catala/compare/struct-maps...CatalaLang:catala:uniform-pass?expand=1)~~

### DIFF
--- a/compiler/dcalc/dune
+++ b/compiler/dcalc/dune
@@ -1,7 +1,15 @@
 (library
  (name dcalc)
  (public_name catala.dcalc)
- (libraries bindlib unionFind utils re ubase catala.runtime_ocaml shared_ast)
+ (libraries
+  bindlib
+  unionFind
+  utils
+  re
+  ubase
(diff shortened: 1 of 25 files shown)


denismerigoux (approved):
Perfect, thank you!

Issue #47, opened 2021-01-03
Improve name resolution for top-level scope, structs and enums


denismerigoux:
Right now, name resolution is done in two passes (in file `src/catala/catala_surface/name_resolution.ml`):

1. Look at scope, struct and enum declarations in order and create identifiers for them and their fields/cases/context
2. Look at scope uses and link the identifiers to the ones created in 1.

This is nice but lacks flexibility for scope, structs and enums. Especially, we want users to be able to declare their scope anywhere in the code, and not following a dependency order. For instance, if scope `A` has a subscope of type `B`, we don't want to force the user to declare `B` before `A`.

This problem can be solved by doing a three-pass name resolution with the following passes:

1. Look at scope, struct and enum declarations and create identifiers for them
2. Look at fields/cases/context are create identifiers for them, also possibly linking to the scope/struct/enum identifiers of 1.
3. Look at scope uses and link the identifiers to the ones created in 1. and 2.

PR #361, opened 2022-11-28
Add a disambiguation pass to resolve structure field accesses

AltGr:


compiler/desugared/disambiguate.ml
```diff
@@ -0,0 +1,78 @@
...
+open Shared_ast
+open Ast
+
+let expr ctx env e =
+  (* The typer takes care of disambiguating: this consists in: - ensuring
+     [EAbs.tys] doesn't contain any [TAny] - [EDStructAccess.name_opt] is always
+     [Some] *)
+  (* Intermediate unboxings are fine since the last [untype] will rebox in
+     depth *)
+  Typing.check_expr ctx ~env (Expr.unbox e)
```
denismerigoux, line 26: Where is the logic that throws an error if you're trying to type `x.f1` but `x` is of type struct `A` and `f1` belongs to struct `B`?

compiler/desugared/from_surface.ml
```diff
@@ -269,53 +266,29 @@ let rec translate_expr
...
-        with Not_found ->
-          Errors.raise_spanned_error (Marked.get_mark x)
-            "Unknown subscope or struct field name"
+      let str =
+        Option.map
+          (fun c ->
+            try Name_resolution.get_struct ctxt c
+            with Not_found ->
+              Errors.raise_spanned_error (Marked.get_mark c)
+                "Struct %s has not been defined before" (Marked.unmark c))
```
denismerigoux, line 278: The error message suggest you must declare (not define) structs before using them. But actually this is not the case since the closing of https://github.com/CatalaLang/catala/issues/47.

compiler/scopelang/from_desugared.ml
```diff
@@ -65,36 +66,53 @@ let rec translate_expr (ctx : ctx) (e : Desugared.Ast.expr) :
...
-    Expr.estruct name (StructFieldMap.map (translate_expr ctx) fields) m
-  | EStructAccess { e; field; name } ->
-    Expr.estructaccess (translate_expr ctx e) field name m
+    Expr.estruct name (StructField.Map.map (translate_expr ctx) fields) m
+  | EDStructAccess { name_opt = None; _ } ->
+    (* Note: this could only happen if disambiguation was disabled. If we want
+       to support it, we should still allow this case when the field has only
+       one possible matching structure *)
+    Errors.raise_spanned_error (Expr.mark_pos m)
+      "Ambiguous structure field access"
```
denismerigoux, line 82: The error message should contain a suggestion about how to disambiguate. Ideally with the names of all the structs corresponding to the field access.

compiler/shared_ast/definitions.ml
```diff
@@ -259,9 +222,23 @@ and ('a, 't) naked_gexpr =
...
     }
       -> (([< desugared | scopelang ] as 'a), 't) naked_gexpr
+  (* [desugared] has ambiguous struct fields *)
+  | EDStructAccess : {
+      name_opt : StructName.t option;
+      e : ('a, 't) gexpr;
+      field : IdentName.t;
+    }
+      -> ((desugared as 'a), 't) naked_gexpr
+  (* Resolved struct/enums, after [desugared] *)
```
denismerigoux, line 235: Should be ocamldoc comments

compiler/shared_ast/expr.ml
```diff
@@ -931,3 +959,33 @@ let make_tuple el m0 =
...
     etuple el m
+
+let translate_op_kind : type a. a op_kind -> 'b op_kind = function
+  | KInt -> KInt
+  | KRat -> KRat
+  | KMoney -> KMoney
+  | KDate -> KDate
+  | KDuration -> KDuration
+
+let translate_op : type a. a operator -> 'b operator = function
```
denismerigoux, line 970: wouahou that's a weird functions but yeah I get why you need it (GADTs...)

compiler/shared_ast/typing.ml
```diff
@@ -373,56 +396,95 @@ and typecheck_expr_top_down :
...
+    in
+    let e_struct' =
+      typecheck_expr_top_down ctx env (unionfind t_struct) e_struct
+    in
+    let name =
+      match UnionFind.get (ty e_struct') with
+      | TStruct name, _ -> name
+      | TAny _, _ ->
+        Printf.ksprintf failwith
+          "Disambiguation failed before reaching field %s" field
```
denismerigoux, line 453: OK so you put the code for disambiguation here. I would have preferred to see it in [compiler/desugared/disambiguate.ml](https://github.com/CatalaLang/catala/pull/361/files/4ee4a96ac7e9f12fe5cf1e7a4be8b78ad624d472#diff-3920d1db78c17509179cc641fe562122da0172f6bcd9ae0b2c827dfa65470f33). Do you think you can switch it?

denismerigoux (approved):
Thanks so much ! I left a couple improvements comments before merging. The question about switching struct disambiguation from the typer to the disambiguation module is more tricky though, not sure if you can actually do it.

compiler/shared_ast/typing.ml
AltGr: The two are pretty inter-mingled: the thing is that for typing an expression like `struct.field1.field2`, you will actually need to disambiguate and resolve `field1` to get the structure type for `field2`.

That's why my first draft, which would only locally call the typer when a disambiguation was required, would end up duplicating most of the logic and re-typing the same sub expressions multiple times.

Another approach could be to expose much more of `Typing` and find a way for external disambiguation code to be called from within the typer, but I fear the complexity of that would make things worse overall.

denismerigoux: yeah after thinking more about it I agree with you

compiler/desugared/from_surface.ml
AltGr: Ah indeed, I should have taken the opportunity to fix it!

compiler/scopelang/from_desugared.ml
AltGr: This case will never be reached unless disambiguation is disabled though (is that something we might want to support ? I left that door open but don't know what the use-cases would be)
Normally, after disambiguation, `name_opt` will always be `Some`, or an error would already have been raised if the struct type couldn't be resolved (that should never happen on a well-typed program).

compiler/shared_ast/expr.ml
AltGr: Hm yeah :roll_eyes: it should be removed soon actually.

compiler/desugared/disambiguate.ml
AltGr: You found it already ;) but precisely and for reference, here:
https://github.com/CatalaLang/catala/blob/4ee4a96ac7e9f12fe5cf1e7a4be8b78ad624d472/compiler/shared_ast/typing.ml#L467-L473

Ideally, we would suggest existing fields that are close syntactically, but that would be useful in many place, so could be a feature for the upcoming error message handling ? We could also list the structures that the specified field actually belongs to, if any :thinking:

AltGr: Actually, if the fields is uniquely resolved to a different struct, as in your example, we could trigger the "normal" type error message with the field requiring type `B` and the struct being type `A`. Do you think that would be more helpful in general ?

denismerigoux: Ideally the error message is customized saying something like "here you want to access a field of struct `B` but the value underlined below has the type of struct `A`"

compiler/scopelang/from_desugared.ml
denismerigoux: Aren't there cases where disambiguation can fail ? We might have cases later when the language evolves where we can't guarantee desambiguation will always work

compiler/desugared/disambiguate.ml
AltGr: indeed that'll be nicer, I'll add a specific message for this case!

compiler/scopelang/from_desugared.ml
AltGr: I... don't think there should be, but that warrants some more formalization :innocent: 

At the moment, anyway, there are none — and by "leaving the door open" I meant that I always kept the ways to manually disambiguate programs.

I guess the future extensions you have in mind would include the ability to define polymorphic functions. If it's real polymorphism, you won't be able to use field accesses on that type anyway so we're fine (the error would be triggered earlier than here, during the disambiguation pass) ;

The only case I could see where this would become problematic is if we start to want type inference on variables, e.g. expecting `fun x -> x.foo` to infer that `x` is `some_struct`. Even if we allow such functions, I would be in favour of asking the user to disambiguate at the point of introduction of the variable when needed.

AltGr: A more straightforward answer: yes, it can fail, but that would happen in the disambiguation pass, not here.

Currently, failure would only happen on badly typed programs. But indeed, with language extensions, we can imagine cases where it would fail on a a correct program and require manual disambiguation.

denismerigoux: Ok some we can leave the message as it is now, and make it better if ever it's really used later.

denismerigoux:
@AltGr can I merge this?

Issue #204, opened 2022-02-18
C backend for Catala


denismerigoux:
Once #158, a C backend can easily be implemented for Catala, starting from the Scalc intermediate representation. The code should roughly follow the structure of the Python backend.

## Enum translation

They should be translated as tagged unions.

## Closures translation

They should be translated to function pointers after a closure conversion pass is done (closure conversion and hoisting can maybe be implemented in Lcalc).

AltGr:
Milestone achieved: I just successfully compiled and ran `tests/arithmetic/good/rounding.catala_en`.

I diverged significantly from the first draft, heading towards a simplified solution that could be improved later on:

1. maybe the biggest change is that I disabled monomorphisation: we'll need it at some point, in particular for equality testing, but it makes everything quite a bit more complex than just handling arrays and options as structs over `void*` pointers, in a polymorphic way.

1. I used GMP and implemented all of our operators over it (with a "functional" layer, more on this below)

1. everything is a pointer: integers, structures, etc. We need them in quite a few places, the base types are pointers (allocated by GMP) anyway, and this simplifies things a lot. This makes point 1. pretty reasonable, call-by-reference is obvious and return values of functions can be handled normally. Of course, there is an indirection cost, and it'll probably be a good idea to carefully unbox specific cases at some point; but I think this is the best starting point. Mixing pointers with unboxed values will probably require extra care that is best done independently.

    Note that almost everything is tagged `const`, and that the C compiler might be able to do some magic.

1. I am using a very dumb allocator that works like an eternally expanding small heap. Since we are very unlikely to be bound by memory, and are in a functional-like context where lots of very small allocations are done (gmp allocates 16 bytes for every integer). This should be extremely fast, and balance out some of the costs from having pointers everywhere. The allocated space is freed in bulk at the end of the computation.

1. This is all tested with C89 and `--pedantic` ; it turns out structure literals don't really exist in C89 (only at declaration, only with constants, and without field names). All declaration also need to be lifted to the top of the block; so we proceed with `some_struct* foo; [...]; foo = catala_malloc(sizeof(some_struct)); foo->field1 = ..., etc.`


The base types are implemented as such:
- options have a tag and a void* pointer (with the correct typecast done at construction and deconstruction time, which is validated by te catala type system)
- arrays similarly have a size field, and a `void**` pointer that is a C array of pointers to the elements
- tuples will be a fixed-size array of `void*` pointers, with the proper casts introduced like for option types


[...]

PR #364, opened 2022-12-06
C backend for Catala : the closure conversion strikes back

denismerigoux:
Follow up on #204.

## Changelog

* [x] Have a very simple example working
* [ ] Debug re-typechecking of Lcalc
* [ ] Debug compiling without exception pass
* [ ] Hoist closures


denismerigoux:
Merging this now because @adelaett depends on some of this code.

denismerigoux:
@AltGr do you think you can review tomorrow to unblock Alain?

compiler/shared_ast/typing.ml
```diff
@@ -330,24 +343,26 @@ let ty (_, { uf; _ }) = uf
...
 and typecheck_expr_top_down :
     type a m.
+    leave_unresolved:bool ->
     A.decl_ctx ->
     (a, m A.mark) A.gexpr Env.t ->
     unionfind_typ ->
     (a, m A.mark) A.gexpr ->
     (a, mark) A.boxed_gexpr =
- fun ctx env tau e ->
+ fun ~leave_unresolved ctx env tau e ->
```
AltGr, line 365: You could have saved some verbosity by adding e.g. `let typecheck_expr_top_down = typecheck_expr_top_down ~leave_unresolved ctx in` here; the verbose version might be clearer though.

compiler/shared_ast/scope.mli
```diff
@@ -44,6 +44,7 @@ val fold_right_lets :
     scope lets to be examined (which are before in the program order). *)
 
 val map_exprs_in_lets :
+  ?reset_types:bool ->
```
AltGr, line 47: An explanation of this arg maybe ? Esp. since it clears the `ScopeLet` types, but doesn't untype the expressions.

compiler/driver.ml
```diff
@@ -309,9 +309,24 @@ let driver source_file (options : Cli.options) : int =
             in
             let prgm =
               if options.closure_conversion then (
+                if not options.avoid_exceptions then
+                  Errors.raise_error
+                    "Option --avoid_exceptions must be enabled for \
+                     --closure_conversion";
```
AltGr, line 315: Might be worth having `--avoid_exceptions` automatically be turned on in this case then ?

compiler/driver.ml
```diff
@@ -213,7 +213,7 @@ let driver source_file (options : Cli.options) : int =
         | `Typecheck ->
           Cli.debug_print "Typechecking again...";
           let _ =
-            try Shared_ast.Typing.program prgm
+            try Shared_ast.Typing.program prgm ~leave_unresolved:true
```
AltGr, line 216: I'm curious why the flag is set here ?

AltGr (approved):
LGTM! I haven't reviewed the new `Closure_conversion` code itself thoroughly, but this all looks pretty nice.

compiler/driver.ml
denismerigoux: Since `--avoid_exceptions` and `--closure_conversion` are experimental these days I want the CLI to force users to actively activate both and not have them activated silently.

compiler/driver.ml
denismerigoux: You're right we're before the closure conversion so this should be set to `false`.

PR #369, opened 2022-12-16
Updates the specified repo for pandoc

zachwick:
The `pandoc` package can no longer be found in the `testing` repo and is now located in the `community` repo.

Omitting the `--repository` option throws an error that the `pandoc` package cannot be found, so this option appears to still be required.

Signed-off-by: zach wick <zach@zachwick.com>


denismerigoux (approved):
Thanks!

PR #370, opened 2022-12-18
Fix typo in tutorial_en.catala_en

MGerrior:
When working through the tutorial, I was thrown off a bit by this minor typo. The typo made it look like the tax calculation was subtracting the $10,000 exception from above, as opposed to the $100,000 from the bracket.


AltGr:
Thanks a lot for your interest and taking the time to fix this!

PR #374, opened 2022-12-20
Force parens in compound logic formulas

### LINKED ISSUE #373: Priority for logical operations
See https://github.com/CatalaLang/catala/pull/368#discussion_r1052289677 :

The usual convention is to have AND bind tighter than OR. At the moment, in Catala, OR is higher priority.

It means that `a and b or c` is understood as `a and (b or c)`, while computer programmers might assume it to be `(a and b) or c`.

I'd be in favour of it to actually be an error and forcing the user to write one of the forms with parens (while still, of course, allowing `a and b and c` etc. without parens). I'll see how that could be done.

AltGr:
Closes #373

This forbids expressions such as `a and b or c`, avoiding the need to set an
implicit priority between `and`, `or` and `xor`, which I find error-prone.

Instead, when that appears, a message asking for explicit parentheses will be
shown to the user.

Implementation note: since that would be extremely tedious to do in the parser
directly, the parser is set to allow right-associativity without discrimination
for the logical operators, and the check is done during desugaring. This
required to explicit parentheses in the surface AST to discriminate the case
where the priority was explicit.


AltGr:
Hm I included XOR in here without thinking about it, but on second thought that should'nt even be transitive at all!

AltGr:
(I mean, obviously _mathematically_ that would be fine but I'd bet nobody will ever be interested in using `a xor b xor c` for what it actually means ; parens don't change anything in that case but forcing them might nudge people into realising it's not doing what they would expect)
(it would also be possible to desugar it into `1 = number of x in [a; b; c] such that x`, which I think is the more useful meaning one may want to give to it — but that would probably be too far-fetched, unless it happens to be a very common pattern in the law ?)

denismerigoux:
I like your approach ! Reviewing this now.

compiler/desugared/from_surface.ml
```diff
@@ -170,6 +170,21 @@ let disambiguate_constructor
...
+    without parens, so that is handled here. *)
+let rec check_formula (op, pos_op) e =
+  match Marked.unmark e with
+  | S.Binop ((((S.And | S.Or | S.Xor) as op1), pos_op1), e1, e2) ->
+    if op <> op1 then
+      Errors.raise_multispanned_error
+        [None, pos_op; None, pos_op1]
+        "Please add parentheses to explicit which of these operators should be \
+         applied first";
+    check_formula (op, pos_op) e1;
```
denismerigoux, line 184: Shouldn't `op` be `op1` in the recursive call?

denismerigoux (approved):
Thanks! Hopefully you're stopping the humongous PRs and getting back to human-sized ones :)

compiler/desugared/from_surface.ml
AltGr: From the `if` above the case where `op <> op1` was excluded, so it's equivalent. For error reporting though, maybe reporting the closest operator makes more sense so you're right, I'm changing it.

PR #377, opened 2023-01-04
Update aides_logement and fix bugs to produce an updated working simulator

denismerigoux:
## Changelog

* [x] Update the housing benefits with last executive order
* [ ] Fix log parsing bug

### DIFF
--- a/compiler/surface/parser.mly
+++ b/compiler/surface/parser.mly
@@ -662,7 +662,7 @@ let source_file_item :=
   END_DIRECTIVE ; {
   let filename = String.trim (String.concat "" args) in
   let pos = Pos.from_lpos $sloc in
-  let jorftext = Re.Pcre.regexp "JORFTEXT\\d{12}" in
+  let jorftext = Re.Pcre.regexp "(JORFARTI\\d{12}|LEGIARTI\\d{12}|CETATEXT\\d{12})" in
   if Re.Pcre.pmatch ~rex:jorftext filename && page = None then
     LawInclude (Ast.LegislativeText (filename, pos))
   else if Filename.extension filename = ".pdf" || page <> None then
--- a/compiler/literate/latex.ml
+++ b/compiler/literate/latex.ml
@@ -151,7 +151,8 @@ codes={\catcode`\$=3\catcode`\^=7}
 \tableofcontents
 
 \[\star\star\star\]
-\clearpage|latex}
+\clearpage
+|latex}
     (match language with Fr -> "french" | En -> "english" | Pl -> "polish")
     (match language with Fr -> "\\setmainfont{Marianne}" | _ -> "")
     (* for France, we use the official font of the French state design system
--- a/examples/aides_logement/archives.catala_fr
+++ b/examples/aides_logement/archives.catala_fr
@@ -1,6 +1,59 @@
-# Archives législatives et réglementaires
+# Archives de l'arrêté du 27 septembre 2019 relatif au calcul des aides personnelles au logement et de la prime de déménagement
 
-## Articles valables du 1er octobre 2021 au 1er juillet 2022
+## Articles valables du 1er juillet 2022 au 31 décembre 2022
+
+### Article 15 | LEGIARTI000046126962 [archive]
+
+Pour l'application du 5° de l'article D. 823-17 du même code, le forfait " R0 " est fixé selon
+le tableau suivant (en euros) :
+
+Composition du foyer	                     MONTANT  (en euros)
+-----------------------------------------  -------------------
+Personne seule sans personne à charge	     4 870
+Couple sans personne à charge	             6 977
+Personne seule ou couple ayant :
+    -une personne à charge	               8 322
+    -deux personnes à charge	             8 509
+    -trois personnes à charge	             8 834
+    -quatre personnes à charge	           9 163
+    -cinq personnes à charge	             9 488
+    -six personnes à charge                9 816
+    -par personne à charge supplémentaire  323
+
+```catala
+champ d'application CalculAidePersonnaliséeLogementLocatif
+  sous condition date_courante >= |2022-07-01| et
+    date_courante < |2022-12-31|:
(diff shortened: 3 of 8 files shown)


tests/test_typing/bad/err6.ml
```diff
@@ -5,124 +5,213 @@ open Runtime_ocaml.Runtime
 [@@@ocaml.warning "-4-26-27-32-41-42"]
 
 module Enum = struct
- type t =
-    | Int of integer
-    | Dec of decimal
-    | Dat of date
-  end
+  type t = Int of integer | Dec of decimal | Dat of date
```
AltGr, line 8: Huh, is this file expected to be committed in ?
`err6.catala_en` is supposed to just have a type error and no output file (which should be inlined anyway)

compiler/dcalc/from_scopelang.ml
```diff
@@ -307,9 +307,121 @@ let rec translate_expr (ctx : 'm ctx) (e : 'm Scopelang.Ast.expr) :
...
+       [tests/test_scope/scope_call3.catala_en], when a scope returns a
+       function, because we insert loggins calls at the call site of the
+       function and not during its definition, then we're missing the call log
+       instructions of the function returned. To avoid this trap, we need to
+       rebind the resulting scope output struct by eta-expanding the functions
+       to insert logging instructions*)
+    let result_var = Var.make "result" in
+    let result_eta_expanded_var = Var.make "result" in
+    let result_eta_expanded =
+      Expr.estruct sc_sig.scope_sig_output_struct
```
AltGr, line 352: When going through the tedious task of manually building AST chunks, I find it helps a lot to write them in comments above as (pseudo-)syntax, both to understand what I am doing and to help reading the code.

compiler/dcalc/from_scopelang.ml
```diff
@@ -307,9 +307,121 @@ let rec translate_expr (ctx : 'm ctx) (e : 'm Scopelang.Ast.expr) :
...
+         (TStruct sc_sig.scope_sig_output_struct, Expr.pos e)
+         result_eta_expanded
+         (tag_with_log_entry
+            (tag_with_log_entry if_then_else_returned
+               (VarDef (TStruct sc_sig.scope_sig_output_struct))
+               direct_output_info)
+            EndCall
+            [ScopeName.get_info scope; Marked.mark (Expr.pos e) "direct"])
+         (Expr.pos e))
+      (Expr.pos e)
```
AltGr, line 424: Good call on the workaround! My bad on completely overlooking the logging issue when I added the scope calls. The solution doesn't feel very satisfying, but it should be solid enough, and hopefully we can have some simplification when we move the function call logs to the function themselves, as you are suggesting.

compiler/dcalc/from_scopelang.ml
```diff
@@ -307,9 +307,121 @@ let rec translate_expr (ctx : 'm ctx) (e : 'm Scopelang.Ast.expr) :
...
+                          (Expr.with_ty m t_out))
+                       (VarDef (Marked.unmark t_out))
+                       (f_markings @ [Marked.mark (Expr.pos e) "output"]))
+                    EndCall f_markings)
+                 [t_in] (Expr.pos e)
+             | _ -> original_field_expr)
+           (StructName.Map.find sc_sig.scope_sig_output_struct ctx.structs))
+        (Expr.with_ty m (TStruct sc_sig.scope_sig_output_struct, Expr.pos e))
+    in
+    let if_then_else_returned =
```
AltGr, line 394: I am not sure I fully understand the rewrite: why is this an `IfThenElse` and not just a `let_in` ?

AltGr (commented):
I have some questions, but otherwise looks good!

compiler/driver.ml
```diff
@@ -202,8 +202,21 @@ let driver source_file (options : Cli.options) : int =
           end
           else prgm
         in
+        (* Cli.debug_print (Format.asprintf "Typechecking results :@\n%a"
+           (Print.typ prgm.decl_ctx) typ); *)
         match backend with
         | `Typecheck ->
+          Cli.debug_print "Typechecking again...";
```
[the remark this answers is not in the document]
denismerigoux: If there's no bug in the compiler this re-typechecking should be transparent to the user. And `Typecheck`still fails as soon as possible : if there's a typing error it will be caught in `Scopelang` and abort. But when deving on the compiler the fact that `Typecheck` does `Dcalc` typechecking is really helpful. Otherwise I found myself to use `Interpret` instead of `Typecheck` just to have that double typechecking in my dev workflow.

compiler/surface/parser.mly
```diff
@@ -662,7 +662,7 @@ let source_file_item :=
   END_DIRECTIVE ; {
   let filename = String.trim (String.concat "" args) in
   let pos = Pos.from_lpos $sloc in
-  let jorftext = Re.Pcre.regexp "JORFTEXT\\d{12}" in
+  let jorftext = Re.Pcre.regexp "(JORFARTI\\d{12}|LEGIARTI\\d{12}|CETATEXT\\d{12})" in
```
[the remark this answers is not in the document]
denismerigoux: I changed the syntax cheet

compiler/dcalc/from_scopelang.ml
denismerigoux: Yes

compiler/dcalc/from_scopelang.ml
denismerigoux: Because I was worried that the `let_in` with a unused value (`true`) could be optimized later by e.g. `js_of_ocaml` or a JS bundler, leading to an incomplete trace formation.

compiler/driver.ml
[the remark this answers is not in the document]
AltGr: Fair enough !
At some point we should have options to toggle these checks on or off, but I agree we're not quite there yet :)

AltGr (approved):

PR #380, opened 2023-01-09
Add support for paths in the parser

AltGr:
This is in preparation for module handling. Lookup isn't handled yet so paths in
the syntax will lead to a "not yet implemented" error (instead of a syntax
error).


denismerigoux (approved):
OK looks good to me!

AltGr:
(rebased)

PR #381, opened 2023-01-10
Fix code in LaTeX literate output

AltGr:
pfffff


AltGr:
Fixing syntax coloring ended up breaking the code output. The LaTex `minted` package is really very fragile…

denismerigoux:
Why do you delete `syntax.tex` in this PR ?

PR #386, opened 2023-01-17
Install Catala plugins

### LINKED ISSUE #378: Package plugin files into opam
Right now the plugin files in `compiler/plugins` are not packaged in the opam packet, which means that someone that wants to use a plugin like the JS of OCaml API generation has to clone the dev repo. @protz suggested to add these files into the `lib/catala` folder of the opam install.

AltGr:
Fixes #378

- the plugins are compiled as libraries rather than with `executable`, so
  that dune is able to install them
- they get installed to
  `lib/catala/plugins/<plugin-name>/<plugin-name>.cmxs`
- the lookup for plugins is now recursive to cope with the plugin
  subdirectories in the point above


AltGr:
This makes e.g. `catala api_web` directly available after an `opam install catala`.

AltGr:
That was fast, thanks :)

PR #391, opened 2023-01-31
Add top-level definitions

AltGr:
This adds top-level definitions to the AST.

Still needs some more work:
- cleanup of the git history
- cleanup of some parts of the code (mostly make some namings more homogeneous,
now there is "topdefs" or "globals", etc.)

- The scalc compilation of topdefs is still todo, we need to discuss the best way
  to express them.
- The web_api plugin certainly needs proof-reading.
- The rewrite of Lcalc.from_dcalc could be separated (use of Var.translate seems
  to remove the need for keeping a variable correspondance map)


AltGr:
Now almost ready: still needs a git history cleanup and maybe some variable renamings, but should otherwise be good.

The actual compilation to python and web_api need proof-reading and tests.

AltGr:
The branch accidentally includes a few patches that aren't really related but were more convenient to add here (feel free to ask if you feel they need a separate review):
- expand tuples to be available in the AST at all passes
- review all parser error messages
- some small fixes in the cheat-sheet

Additional note: a syntax to *call* multi-argument functions was added so that I could write tests for those, but is to be considered experimental. Furthermore, while it's now possible to define multi-argument **top-level** functions, the syntax for local or scope-local functions hasn't been extended: those remain single-argument for the time being even though there is no technical reason to keep this limitation.

compiler/scalc/print.ml
```diff
@@ -165,20 +169,35 @@ and format_block
...
+let format_item decl_ctx ?debug ppf def =
+  Format.pp_open_hvbox ppf 2;
+  Format.pp_open_hovbox ppf 4;
+  Print.keyword ppf "let ";
+  let () =
+    match def with
+    | SVar { var; expr } ->
+      format_var_name ppf var;
+      Print.punctuation ppf " =";
+      Format.pp_close_box ppf ();
```
denismerigoux, line 181: Why don't you use the `Format` specification syntax with `@[` ?

compiler/scopelang/dependency.ml
```diff
@@ -19,7 +19,46 @@
 
 open Catala_utils
 open Shared_ast
-module SVertex = ScopeName
+
+type vertex = Scope of ScopeName.t | Topdef of TopdefName.t
+
+module SVertex = struct
+  type t = vertex
+  (* While we enforce that globals don't depend on scopes, and could therefore
```
denismerigoux, line 27: Well do we really enforce this invariant right now? Direct scope calling is an expression so syntactically it could fit in a toplevel def right?

compiler/scopelang/print.ml
```diff
@@ -98,6 +98,24 @@ let scope ?(debug = false) ctx fmt (name, decl) =
              SubScopeName.format_t subscope_name Print.punctuation "]"))
     decl.scope_decl_rules
 
+let print_topdef ctx ppf name (e, ty) =
+  Format.pp_open_vbox ppf 2;
+  let () =
+    Format.pp_open_hovbox ppf 2;
+    Print.keyword ppf "let";
+    Format.pp_print_space ppf ();
+    TopdefName.format_t ppf name;
```
denismerigoux, line 107: Why not use regular `@[` `Format` syntax?

compiler/surface/lexer.cppo.ml
```diff
@@ -601,14 +601,16 @@ let rec lex_code (lexbuf : lexbuf) : token =
       Buffer.add_string cents (String.make (2 - Buffer.length cents) '0');
       L.update_acc lexbuf;
       MONEY_AMOUNT (Buffer.contents units, Buffer.contents cents)
-  | Rep (digit, 4), '-', Rep (digit, 2), '-', Rep (digit, 2) ->
+  | '|', Rep (digit, 4), '-', Rep (digit, 2), '-', Rep (digit, 2), '|' ->
```
denismerigoux, line 604: Yes good decision

compiler/surface/parser.mly
```diff
@@ -596,6 +601,16 @@ let enum_decl_line :=
   }
 }
 
+(* let def_depends ==
```
denismerigoux, line 604: Stale comment?

doc/syntax/syntax_en.tex
```diff
@@ -444,9 +445,9 @@ \section{Metadata declaration}
   \\
   State transitions declaration & \begin{catala}
     ```catala
-    internal var1 content ...
-      state before
-      state after
+      internal var1 content ...
+        state before
+        state after
```
denismerigoux, line 450: Don't you put the new syntax for toplevel definitions here ?

tests/test_name_resolution/bad/toplevel_defs.catala_en
```diff
@@ -0,0 +1,24 @@
...
+  equals (output of S1).a
+```
+
+```catala-test-inline
+$ catala typecheck
+[ERROR] Scope calls are not allowed outside of a scope
+
+┌─⯈ tests/test_name_resolution/bad/toplevel_defs.catala_en:11.10-22:
+└──┐
+11 │   equals (output of S1).a
```
denismerigoux, line 20: Yes!

denismerigoux (approved):
What a beast! Very glad to have that landed in. There's some cleanup to do before merging but thanks for the work!

compiler/scopelang/print.ml
AltGr: I actually find this style more readable when the format string is mostly a concatenation: mentally matching the `%a` with the correct parameters is a pain. Things are also made worse with `ocamlformat` which doesn't allow us to keep a clear pairing between the printers and their arguments.

The `let ()` is here just to visualise the boundaries of the box (I'd have used `begin/end` but `ocamlformat` scraps them).

And it's not that much more verbose when even symbols need a printer for color.

doc/syntax/syntax_en.tex
AltGr: I will, but I'd prefer to wait until we agree on a stable version — I am assuming undocumenting the syntax implies it's experimental for now :)

compiler/dcalc/interpreter.ml
```diff
@@ -364,6 +364,12 @@ and evaluate_expr (ctx : decl_ctx) (e : 'm Ast.expr) : 'm Ast.expr =
          if the term was well-typed)"
         (Expr.format ctx ~debug:true)
         e StructName.format_t s)
+  | ETuple es ->
+    Marked.same_mark_as (ETuple (List.map (evaluate_expr ctx) es)) e
+  | ETupleAccess { e = e1; index; size } ->
+    Marked.same_mark_as
+      (ETupleAccess { e = evaluate_expr ctx e1; index; size })
```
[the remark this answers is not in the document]
AltGr: Oops indeed, thanks for catching this!

compiler/desugared/from_surface.ml
```diff
@@ -1238,7 +1303,8 @@ let init_scope_defs
               scope_def_map)
         sub_scope_def.Name_resolution.var_idmap scope_def_map
   in
-  IdentName.Map.fold add_def scope_idmap Ast.ScopeDefMap.empty
+  IdentName.Map.fold add_def scope_idmap
+    Ast.ScopeDefMap.empty (* TODO: add topdefs too *)
```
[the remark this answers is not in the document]
AltGr: Ah that's a stale comment that I forgot to remove: registering the toplevel declaration infos is already done at the `Name_resolution` step. This function registers the internal scope vars / subscope structures, which doesn't apply to toplevel definitions.

compiler/desugared/name_resolution.mli
```diff
@@ -149,6 +150,10 @@ val get_scope : context -> IdentName.t Marked.pos -> ScopeName.t
 (** Find a scope definition from the typedefs, failing if there is none or it
     has a different kind *)
 
+val process_base_typ : context -> Surface.Ast.base_typ Marked.pos -> typ
+(** Convert a surface base type to an AST type *)
+(* Note: should probably be moved to a different module *)
```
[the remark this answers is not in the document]
AltGr: No since it's used a lot in `Name_resolution`, on which `From_surface` depends. Maybe it would make sense directly in `Ast` ?

The reason I asked is that this was internal to `Name_resolution`, but I now needed to expose the function to `From_surface` that uses it when translating toplevel defs and didn't want to duplicate the code. It's not that bad though, maybe we can wait and see if further evolutions warrant a new module for these conversions ?

compiler/lcalc/compile_without_exceptions.ml
```diff
@@ -498,39 +511,34 @@ let translate_scope_body
         })
       (Bindlib.bind_var v' (translate_scope_let ctx' lets))
 
-let rec translate_scopes (ctx : 'm ctx) (scopes : 'm D.expr scopes) :
-    'm A.expr scopes Bindlib.box =
-  match scopes with
-  | Nil -> Bindlib.box Nil
-  | ScopeDef { scope_name; scope_body; scope_next } ->
```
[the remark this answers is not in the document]
AltGr: Be sure to check the functions to fold over this in module `Scope`, that should save you re-writing the matches and bind-unbind  every time (and extend those as needed).

compiler/scalc/print.ml
AltGr: See my answer [below](https://github.com/CatalaLang/catala/pull/391#discussion_r1106123928)

compiler/scopelang/dependency.ml
AltGr: Syntactically, it could, but wereject it explicitely [during desugaring](https://github.com/CatalaLang/catala/pull/391/files/c3af0b409735d4fff803b40fc9a19f37809641f1#diff-185db1a11ce406fa3ae293f0de69e5252be36e4f312c5195ad6216c93c73d43bR403-R406)

compiler/surface/parser.mly
AltGr: Indeed, thanks!

doc/syntax/syntax_en.tex
denismerigoux: Ok, I'll try to not forget this :)

compiler/desugared/name_resolution.mli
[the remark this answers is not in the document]
denismerigoux: Yes I agree.

PR #397, opened 2023-02-15
Add date rounding option in scope

### LINKED ISSUE #293: Specify date rounding mode in scope declaration
## The problem

Until now, Catala uses the [calendar](https://v3.ocaml.org/p/calendar/2.04) OCaml library for its dates computations. However, the level of precision offered by this library is insufficient to handle the ambiguities of dates arithmetic in the Gregorian calendar. For instance : what is `March 31 st + 1 month` ? Is is `May 1st` or `April 30th` ?

To give a precise semantics to this kind of dates conundrum, we've developped our own dates computation library : https://github.com/CatalaLang/dates-calc. This library aims to offer a correctness-focused, minimal backend for Catala's dates computation that can be ported in other languages. 

During the development of the library, it quickly appeared that date arithmetic is not exact and has to incorporate a kind of rounding for it to work. For instance,  `March 31 st + 1 month` is rounded up `May 1st` and rounded down `April 30th`. By default, ambiguous computations that need rounding yield a runtime exception. But the law might set different meta-rule for dates computation. So, we need 
a way to specify the preferred rounding modes of dates in Catala.

## Proposed solution

The naïve solution would be to set the dates rounding mode via an imperative flag, or a CLI option globally for the whole program. However, this is not functional enough for us, and might pose problem when two programs modelling two different portions of the law interact while having two different date rounding conventions.

So what we propose, is to attach the date rounding mode as a declaration belonging to a scope. All operations on dates that belong to this scope will inherit the declared date rounding mode. Here's a proposal for the syntax:

```
declaration scope Foo:
  input x content integer
  ...
  output y content boolean

# Later, attached to some article of law
scope Foo:
  declaration date rounding mode decreasing # or "increasing"
``` 

This syntax reuses part of the `MetaAssertion` surface syntax that's currently unsused and should be cleaned/refactored.

rmonat:
This MR fixes #293. Example usecase:
```catala_en
declaration scope AgeIsLessThan:
  input birthday content date
  input currentday content date
  input years content duration
  output r content boolean

scope AgeIsLessThan:
  definition r equals birthday + years <= currentday
  date round increasing 

declaration scope Test:
  age scope AgeIsLessThan
  internal bday content date
  internal today content date

scope Test:
  definition bday equals |2000-02-29|
  definition today equals |2018-03-01|
  definition age.birthday equals bday
  definition age.currentday equals today
  definition age.years equals 18 year
  assertion age.r
#  assertion bday + 18 year <= today # would fail 
```

Note that this option does not leak in subscope calls. The `from_scopelang` pass translates this option into the corresponding rounding argument of the addition operator between dates and durations. 

Current status:
- [x] Update parser
- [x] Update parser.messages
- [x] Handle translation in dcalc
- [x] Fix lcalc translation
- [x] Port French examples
- [x] Documentation
- [x] Clean commit history
- [x] Add specific tests

Questions:
1) in `compiler/surface.ml:590...`, are visitors update necessary?
2) regarding documentation, shall I update the date section of the tutorials?


examples/prologue_france/prologue.catala_fr
```diff
@@ -9,4 +9,24 @@ déclaration énumération Collectivité :
...
+
+déclaration champ d'application CalculAgeSupStrict:
+  entrée date_naissance contenu date
+  entrée date_courante contenu date
+  entrée années contenu durée
+  résultat r contenu booléen
+
+champ d'application CalculAgeSupStrict:
+  définition r égal à date_naissance + années > date_courante
+  date arrondi croissant
```
denismerigoux, line 31: Why did you create a separate scope to do this ? The code for the date checking is in `prologue.catala_fr` so not justified by any legal rule. We had said that we wanted `date rounding increasing` inside `scope` blocks and not `declaration scope` blocks in order to justify them by legal rules. 

In this example I would not have created these two extra scope and just would have added `date arrondi croissant` to the scope `ÉligibilitéAidesPersonnelleLogement`.

denismerigoux (approved):
The syntax is great for me but the way you implemented the housing benefits computation is a little bit weird to me, I've left some comments :)

examples/prologue_france/prologue.catala_fr
rmonat: Here is my rationale for this approach: I don't want all date computations within `ÉligibilitéAidesPersonnelleLogement` to use the same rounding mode. For age computations I'm pretty sure rounding up is the correct way to do it (although I don't know the exact law defining this, hence the prologue definition), but what about other cases?

denismerigoux: Ok I understand better. We'll ask them officials the question about date rounding to know more :)

denismerigoux:
@rmonat what's the status of this?

rmonat:
@denismerigoux if you're fine with the current definition I just have to add a bit of doc and it'll be ready for review.

denismerigoux:
Yes I'm fine with it, go ahead :) Thanks for the work!

compiler/surface/parser.mly
```diff
@@ -496,6 +496,17 @@ let scope_item :=
...
+| DATE ; i = LIDENT ; v = addpos(variation_type) ;
+  {
+    (* Round is a builtin, we need to check which one it is *)
+    match Localisation.lex_builtin i with
+    | Some Round ->
+       DateRounding(v), Marked.get_mark v
+    | _ -> 
+         Errors.raise_spanned_error
+           (Pos.from_lpos $sloc)
+           "Expected round"
```
AltGr, line 508:
```suggestion
           (Pos.from_lpos $loc(i))
           "Expected the form 'date round increasing' or 'date round decreasing'"
```
The message would still be in english only, but with more context it's easier to understand

compiler/desugared/from_surface.ml
```diff
@@ -1142,6 +1142,21 @@ let process_scope_use_item
...
+  | Surface.Ast.DateRounding (r, _) ->
+    let scope_uid = scope in
+    let scope : Ast.scope = ScopeName.Map.find scope_uid prgm.program_scopes in
+    let r =
+      match r with
+      | Surface.Ast.Increasing -> Ast.Increasing
+      | Surface.Ast.Decreasing -> Ast.Decreasing
+    in
+    let new_scope =
+      { scope with scope_options = Ast.DateRounding r :: scope.scope_options }
```
AltGr, line 1154: What happens with several (possibly conflicting) rounding modes in the same scope ? We may want it to be an error.

compiler/dcalc/from_scopelang.ml
```diff
@@ -914,6 +916,17 @@ let translate_scope_decl
         | _ -> ctx)
       ctx scope_variables
   in
+  let date_rounding =
+    List.find_opt
+      (function Desugared.Ast.DateRounding _ -> true)
+      sigma.scope_options
+    |> Option.map (function
+         | Desugared.Ast.DateRounding Desugared.Ast.Increasing ->
+           (RoundUp : Runtime_ocaml.Runtime.date_rounding)
```
AltGr, line 925:
```suggestion
           Runtime_ocaml.Runtime.RoundUp
```
is the more classic syntax, more concise and doesn't rely on type-based disambiguation

AltGr (approved):
Very nice, thanks!

Summarising my review:
- a little bit of (advisory) style nitpicking (I hope I'm not being obnoxious :grimacing: )
- a few questions about some edge cases and documentation
- a couple stale files ?

Also I am not absolutely convinced by the syntax (I guess `increasing` and `decreasing` got used because they already were keywords, albeit unused ?)
I don't mind it either, though.

compiler/scalc/to_python.ml
```diff
@@ -70,8 +70,8 @@ let format_op
   | LastDayOfMonth -> Format.pp_print_string fmt "last_day_of_month"
   | Round_mon -> Format.pp_print_string fmt "money_round"
   | Round_rat -> Format.pp_print_string fmt "decimal_round"
-  | Add_int_int | Add_rat_rat | Add_mon_mon | Add_dat_dur | Add_dur_dur | Concat
-    ->
+  | Add_int_int | Add_rat_rat | Add_mon_mon | Add_dat_dur _ | Add_dur_dur
+  | Concat ->
```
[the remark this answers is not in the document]
rmonat: Yes I guess this is related to #398. I added it there, but let me know if you want me to write it somewhere else.

rmonat:
Thank you very much Louis! I'm still working on the rebase

rmonat:
@AltGr done! (I think). I'm really unsure about the `rebase` though. Let me know.

examples/prologue_france/prologue.catala_fr
rmonat: @AltGr there is currently no way to do it. I think it should at least be added when functions of multiple arguments can be defined and called in toplevel definitions, what do you think?

compiler/desugared/from_surface.ml
```diff
@@ -1142,6 +1142,32 @@ let process_scope_use_item
...
+      match
+        List.find_opt
+          (fun scope_opt ->
+            scope_opt = Ast.DateRounding Ast.Increasing
+            || scope_opt = Ast.DateRounding Ast.Decreasing)
+          scope.scope_options
+      with
+      | Some _ ->
+        Errors.raise_spanned_error (Marked.get_mark item)
+          "A date rounding mode has already been specified"
```
AltGr, line 1163: I know I am being annoying :grimacing: but it would be very nice for the user to show the two positions of the two date rounding specifications here :pray:

examples/prologue_france/prologue.catala_fr
rmonat: Regarding documentation, where is it best to explain these behaviors?

compiler/desugared/from_surface.ml
rmonat: You're absolutely right. This is done now.

examples/prologue_france/prologue.catala_fr
rmonat: Sorry, I missed the support for that. I opened #431.

denismerigoux:
@AltGr can we merge now?

AltGr:
Yup! I was just waiting for CI :)

PR #400, opened 2023-02-17
Fix python output of toplevel defs

AltGr:


compiler/scalc/to_python.ml
```diff
@@ -601,7 +601,7 @@ let format_program
         (Format.pp_print_list ~pp_sep:Format.pp_print_newline (fun fmt ->
            function
            | SVar { var; expr } ->
-             Format.fprintf fmt "@[<hv 4>%a = (@,%a@,@])@," format_var_name var
+             Format.fprintf fmt "@[<hv 4>%a = (@,%a@,@])@," format_var var
```
denismerigoux, line 604: Ah well caught!

tests/test_name_resolution/good/toplevel_defs.catala_en
```diff
@@ -192,3 +191,284 @@ let S_9 (S_in_32: S_in {}) =
...
+    return ((decimal_of_money(x_1) * y) + decimal_of_string("10."))
+
+def glob5_aux():
+    def glob5(x_2:Decimal):
+        y_1 = decimal_of_string("1000.")
+        return (x_2 * y_1)
+    return glob5((decimal_of_integer(integer_of_string("2")) *
+                 decimal_of_string("3.")))
+
+glob5_1 = (glob5_aux())
```
denismerigoux, line 366: Is this the wanted behavior? Why pass through this auxiliary variable ?

denismerigoux (approved):
Thanks!

tests/test_name_resolution/good/toplevel_defs.catala_en
AltGr: I have been conservative in the compilation to `scalc`, but believe it can be simplified indeed (I've started some work in that direction, will be happy to discuss it at the next occasion), here and in a few other cases.

The scheme I've chosen at the moment, whenever the computation of a toplevel variable spills out statements, is to wrap its computation in an auxiliary function with no arguments, and apply that just once. The output seems consistent with that, but it could probably be simplified and flattened.

PR #401, opened 2023-02-20
Python runtime: Fix source positions linked to run-time errors

rohanpadhye:
Fixes a typo-like bug where the `source_position` field of error objects was assigned the class `SourcePosition` instead of the actual position passed in as an argument.


AltGr (approved):
Well spotted, thanks!

PR #408, opened 2023-02-24
Fix flake update mechanism

adelaett:
The update mecanism was broken from a few weeks. Even if the flake.nix still build with respect to the flake.lock in the repo, it was lagging being the lastest nixpkgs updates. This fixes the issue. 

Moreother, i took the liberty to check why nix build don't catala's test suite. The reason was to run tests, buildDunePackages runs `dune build -p catala`, and this do nothing because there is no test registered with catala, everything is registered with clerk. Hence I've fixed clerk's nix package to run tests and added to both regular CI and flake-update CI.

### DIFF
--- a/.github/workflows/run-builds.yml
+++ b/.github/workflows/run-builds.yml
@@ -11,5 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
       - uses: cachix/install-nix-action@v15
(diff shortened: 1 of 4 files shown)


adelaett:
I've modified the nix build so it uses github's cache for dependencies. This reduces the time it takes to compile and run tests.

PR #409, opened 2023-02-24
Adelaett fix arrowlist

adelaett:
Follow-up of #404.

Last pull request enabled multiple argument functions within the compiler (even if it is not possible to create those in the surface language). This broke the handmade event parsers in the runtime. This change correct the handcrafted parser by permitting multiple VarDefintions. It have been tested on french housing and family benefits.

Left to do:
- include this part of the compiler to the tests to not get in the same kind of issues later on.

### DIFF
--- a/french_law/ocaml/bench.ml
+++ b/french_law/ocaml/bench.ml
@@ -224,6 +224,16 @@ let run_test_aides_logement () =
     exit (-1)
   | Runtime.AssertionFailed _ -> ()
 
+let _test =
+  let _ = run_test_aides_logement () in
+  let raw_events = Runtime.retrieve_log () in
+  Runtime.EventParser.parse_raw_events raw_events
+
+let _test =
+  let _ = run_test_allocations_familiales () in
+  let raw_events = Runtime.retrieve_log () in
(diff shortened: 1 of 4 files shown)


denismerigoux:
Is it ready? Should we merge before #412?

adelaett:
yes it is ready. I am waiting for the fix in catala-website

adelaett:
The parsing code is now in the CI, and the PR of the change on catala's website is made. Ready to merge.

denismerigoux (approved):
Looks good to me! @EmileRolley care to review?

french_law/ocaml/bench.ml
```diff
@@ -224,6 +224,16 @@ let run_test_aides_logement () =
     exit (-1)
   | Runtime.AssertionFailed _ -> ()
 
+let _test =
+  let _ = run_test_aides_logement () in
+  let raw_events = Runtime.retrieve_log () in
+  Runtime.EventParser.parse_raw_events raw_events
```
EmileRolley, line 230: question: we assume here the test pass if there is no triggered exceptions?

EmileRolley (approved):
It looks great to me :)

PR #413, opened 2023-02-28
Require argument name declarations, support multiple function arguments everywhere

AltGr:


AltGr:
I'll now be updating the syntax cheat-sheet, but this should be good to go

examples/aides_logement/code_construction_reglementaire.catala_fr
```diff
@@ -4645,10 +4646,10 @@ constatée.
 
 ```catala
 champ d'application CalculAllocationLogementAccessionPropriété:
-  définition abattement_dépense_nette_minimale de allocation_mensuelle
+  définition abattement_dépense_nette_minimale de aide_calculée_1_8_d832_24
```
denismerigoux, line 4649: That's a weird name no ?

denismerigoux: Why didn't you keep `allocation_mensuelle` everywhere ?

tests/test_exception/bad/exceptions_cycle.catala_en
```diff
@@ -20,60 +20,36 @@ scope A:
 
 ```catala-test-inline
 $ catala Interpret -s A
-[ERROR] Cyclic dependency detected between exceptions!
+[ERROR] Exception cycle detected when defining x: each of these 3 exceptions applies over the previous one, and the first applies over the last
```
denismerigoux, line 23: Very nice error message!

tests/test_func/bad/param_inconsistency.catala_en
```diff
@@ -0,0 +1,33 @@
...
+  definition f1 of x under condition cond
+    consequence equals decimal of (x * 2)
+  definition f1 of y under condition not cond
+    consequence equals y / 2
+  definition out equals f1 of 10
+```
+
+```catala-test-inline
+$ catala typecheck
+[ERROR] The arguments of this definition don't match the declaration.
```
denismerigoux, line 17: can you add to the message `("x" is different from "y")` ?

denismerigoux (approved):
Thanks! Looks good to me apart from these few fixes.

examples/aides_logement/code_construction_reglementaire.catala_fr
AltGr: Hum I had to choose between the two, and kept the one that felt more precise, but I can switch.

denismerigoux: Yeah in this case `allocation_mensuelle` is preferable.

tests/test_func/bad/param_inconsistency.catala_en
AltGr: It's more complicated than that because we're comparing lists, but done.
(didn't go as far as detecting permutations though :sweat_smile: )

PR #422, opened 2023-03-04
Fix non-transitive comparison function on desugared vars

AltGr:


denismerigoux:
Wouhaou that was causing the bug?

AltGr:
Thanks for the reformatting!

PR #423, opened 2023-03-07
Solidity backend

JMLX42:
Ongoing pull request to add support for a Soildity backend.

This is a very (very) early implementation. Please do not merge until all tasks are done.

- [x] Enum type declaration
- [x] Struct  type declaration
- [ ] Function pointer type declaration
- [ ] Optional type declaration
- [ ] Expression support
- [ ] Statement support

PR #424, opened 2023-03-08
Some small fixes, error message improvements, CI artifacts cleanup

AltGr:
this is stuff that turned up during pair-programming sessions: cycles should now be much easier to understand and debug ; a _pro rata temporis_ was required ; the syntax for rules with states was inconsistent.

The part on the artifacts is independent: it affects https://catalalang.github.io/catala/ by
- having the french_law artifacts as separate .tar.gz files (rather than a hard-to-download scattered tree layout)
- adding the git revision tag to these artifacts (but including stable links for the latest release)
- skipping the api doc tree (the top-level html index is enough for browsing)


.github/workflows/run-make-all.yml
```diff
@@ -31,16 +31,23 @@ jobs:
         if: ${{ github.ref == 'refs/heads/master' }}
         run: |
           RELEASE_TAG=$(git describe --tags)
-          mkdir -p artifacts
           docker run --rm catalalang/catala-build:${IMAGE_TAG} sh -uexc '
```
denismerigoux, line 34: Maybe you could make that line more robust see https://github.com/CatalaLang/catala/pull/423#issuecomment-1458634308

tests/test_scope/bad/cyclic_scope_calls.catala_en
```diff
@@ -0,0 +1,55 @@
...
+
+```catala-test-inline
+$ catala typecheck
+[ERROR] Cyclic dependency detected between the following scopes:
+  S4 → S3 → S2 → S4
+
+S4 is used here in the definition of S3:
+┌─⯈ tests/test_scope/bad/cyclic_scope_calls.catala_en:21.23-35:
+└──┐
+21 │   definition o equals (output of S4).o
```
denismerigoux, line 37: Amazing error message!

denismerigoux (approved):
Thanks!

PR #427, opened 2023-03-11
fix(runtimes/ocaml): rename `fun_call.inputs` into `fun_call.fun_inputs`

EmileRolley:
Needed by https://github.com/CatalaLang/catala-website/pull/46


denismerigoux (approved):
Ah yes decco needed the field names to be exactly the same I guess?

PR #428, opened 2023-03-13
Embed the Pygments lexers in Catala

AltGr:
Instead of requiring that our custom lexers for Catala be installed on the
system (or the proper venv loaded), this passes them to the `pygmentize` command
on the fly.

This makes the HTML literate output usable on any stock installation of
Pygments.

In addition, this also works for LaTeX output by removing the need for `minted`,
and generating the latex for colored code output by calling `pygmentize`
directly from Catala, using the built-in lexers.

### DIFF
--- a/compiler/literate/html.ml
+++ b/compiler/literate/html.ml
@@ -30,13 +30,6 @@ module C = Cli
 (** Converts double lines into HTML newlines. *)
 let pre_html (s : string) = String.trim (run_pandoc s `Html)
 
-(** Raise an error if pygments cannot be found *)
-let raise_failed_pygments (command : string) (error_code : int) : 'a =
-  Errors.raise_error
-    "Weaving to HTML failed: pygmentize command \"%s\" returned with error \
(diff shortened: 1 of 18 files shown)


compiler/literate/html.ml
```diff
@@ -57,28 +50,19 @@ let wrap_html
...
-    "<head>\n\
+    "<!DOCTYPE html>\n\
+     <html>\n\
+     <head>\n\
      <style>\n\
      %s\n\
      </style>\n\
      <meta http-equiv='Content-Type' content='text/html; charset=utf-8'/>\n\
      </head>\n\
+     <body>\n\
```
denismerigoux, line 65: When is this closed?

denismerigoux (approved):
That's so neat, thanks!

compiler/literate/html.ml
denismerigoux: Ok I saw it below.

PR #430, opened 2023-03-14
Compile LaTeX code using minted without the Python venv

AltGr:
This is a hack, but not a dirty one: a new command `catala pygmentize` is added,
which is just a wrapper around `pygmentize` that calls it with the proper lexers
defined.

The point is that this needs no installation, just a stock `pygmentize`
installation and the `catala` binary.


doc/syntax/syntax_en.tex
```diff
@@ -26,6 +26,8 @@
 
 \setlist[itemize]{noitemsep, topsep=0pt}
 
+\renewcommand{\MintedPygmentize}{../../_build/default/compiler/catala.exe pygmentize}
```
denismerigoux, line 29: Could you put somewhere in the README files or doc what `catala pygmentize` is for and how to use it for instance for powering LaTeX's `minted` ?

denismerigoux (approved):
Nice hack, I support it.

doc/syntax/syntax_en.tex
rmonat: Let's say I have some catala code in my latex and some c code somewhere else. Will `catala pygmentize` also support the c code?

PR #433, opened 2023-03-24
Rework the AST Gadt to allow merging of different ASTs

AltGr:
The phantom polymorphic variant qualifying AST nodes is reversed:
- previously, we were explicitely restricting each AST node to the passes where it belonged using a closed type (e.g. `[< dcalc | lcalc]`)
- now, each node instead declares the "feature" it provides using an open type (e.g. `[> 'Exceptions ]`)
- then the AST for a specific pass limits the features it allows with a closed type

The result is that you can mix and match all features if you wish,
even if the result is not a valid AST for any given pass. More
interestingly, it's now easier to write a function that works on
different ASTs at once (it's the inferred default if you don't write a
type restriction).

The opportunity was also taken to simplify the encoding of the
operators, which don't need a second type parameter anymore.


compiler/dcalc/interpreter.ml
```diff
@@ -157,159 +153,182 @@ and evaluate_operator :
...
+        | Mult_rat_rat | Mult_mon_rat | Mult_dur_int | Div_int_int | Div_rat_rat
+        | Div_mon_mon | Div_mon_rat | Div_dur_dur | Lt_int_int | Lt_rat_rat
+        | Lt_mon_mon | Lt_dat_dat | Lt_dur_dur | Lte_int_int | Lte_rat_rat
+        | Lte_mon_mon | Lte_dat_dat | Lte_dur_dur | Gt_int_int | Gt_rat_rat
+        | Gt_mon_mon | Gt_dat_dat | Gt_dur_dur | Gte_int_int | Gte_rat_rat
+        | Gte_mon_mon | Gte_dat_dat | Gte_dur_dur | Eq_int_int | Eq_rat_rat
+        | Eq_mon_mon | Eq_dat_dat | Eq_dur_dur ),
+        _ ) ->
+      err ()
+    | _ -> .
```
denismerigoux, line 331: What does this `.` mean?

compiler/shared_ast/definitions.ml
```diff
@@ -95,134 +141,132 @@ type log_entry =
...
+  (** Explicit monomorphic versions of the overloaded operators *)
 
   (** Classification of operators. This could be inlined in the definition of
       [t] but is more concise this way *)
-  type (_, _) kind =
-    | Monomorphic : ('a any, monomorphic) kind
-    | Polymorphic : ('a any, polymorphic) kind
-    | Overloaded : ([< desugared ], overloaded) kind
-    | Resolved : ([< scopelang | dcalc | lcalc ], resolved) kind
+  (* type (_, _) kind =
```
denismerigoux, line 164: This comment should be removed?

compiler/shared_ast/operator.ml
```diff
@@ -332,88 +332,18 @@ let kind_dispatch :
     | Eq_mon_mon | Eq_dat_dat | Eq_dur_dur ) as op ->
     resolved op
 
+let translate (r : date_rounding option) t =
```
denismerigoux, line 335: `r` should be a optional argument with an explicit name, I read a call to `Operator.translate` earlier with a `None` and I had no idea what it meant.

compiler/shared_ast/var.ml
```diff
@@ -21,14 +21,15 @@ open Definitions
...
-type 'e t = ('a, 't) naked_gexpr Bindlib.var constraint 'e = ('a any, 't) gexpr
+type 'e t = ('a any, 't) naked_gexpr Bindlib.var constraint 'e = ('a, 't) gexpr
 
 type 'e vars = ('a, 't) naked_gexpr Bindlib.mvar
   constraint 'e = ('a any, 't) gexpr
 
 let make (name : string) : 'e t = Bindlib.new_var (fun x -> EVar x) name
 let compare = Bindlib.compare_vars
 let eq = Bindlib.eq_vars
+let translate_mark v = Bindlib.copy_var v (fun x -> EVar x) (Bindlib.name_of v)
```
denismerigoux, line 32: What is the difference with `translate` just below?

denismerigoux (approved):
Thanks @AltGr !

compiler/dcalc/interpreter.ml
denismerigoux: Wouahou it's the "turfu"

compiler/shared_ast/operator.ml
AltGr: Indeed; actually I don't even think it's worth having it as a parameter to this function, it's unrelated and used only once.

compiler/shared_ast/var.ml
AltGr: The implementation happens to be the same, but the difference is in the  `mli`: this one has a more restricted type.

But you're right that just adds noise, I'll remove it.

PR #436, opened 2023-03-28
Update INSTALL.md

rprimet:
It seems that building with ocaml < 5.0.0 will not work due to  `UChar.utf_decode_length` not being available in the stdlib prior to that version


AltGr:
Hum I'm using 4.14 :)

rprimet:
Yep, you're right, it seems that `UChar.utf_decode_length` was added in 4.14

Issue #392, opened 2023-02-13
An attribute language for Catala code items


denismerigoux:
## The problem

While discussing #293, we were wondering whether we should add new syntax to Catala to be able to express the date rounding mode. In this case, the result was yes because we deemed it significant enough. However, there will be many cases when we want to attach information to a point of the code (a scope, a variable, a definition, etc.). 

One use case for such additional information might be for a custom Catala backend that translates Catala programs to specific language in a specific legacy environment where the new Catala program connects to an older application. You might want to specify at the level of a scope signature which variables of the Catala scope connect to which variables of a function in the legacy application. 

An example of a fully-fledged attribute language is the one of [Rust](https://doc.rust-lang.org/reference/attributes.html). 

## Discussion

By designing and implementing a fully-fledged attribute language ahead, we are future-proof and offer a flexible system for custom needs in attaching information to the AST. As for Rust, this could be used for storing comments and documentation in the AST.

But by offering such a mechanism, we are allowing users to clutter the code with metadata that is not related to the text of the law. The syntax of the attribute language might also confuse lawyers and obscure their ability to read the codebase. 

A compromise could be to still design this attribute language, but implement it in an extrinsic way. Instead of putting the attributes directly in the source code, you could have an external `.metadata` file containing the association of attributes to code items :

```
TaxModule.IncomeTaxScope.some_var : attribute1, attribute2, etc.
```

The compiler should statically check whether the items referenced in the metadata file exist and attach the info at the AST level.

What do you think?

AltGr:
I need to examinate the options more thoroughly, but I would tend to prefer inline attributes, which are generally easier to maintain.

Another use I'd be very happy about could be attaching labels and specifications to the user-directed inputs, to help generate forms ot interactive  applications.

denismerigoux:
Closed because it's implemented now!

PR #438, opened 2023-03-30
Add common linting passes to Catala

denismerigoux:
All lints created by this PR generate warnings. Maybe we should at this point give numbers to warnings and the ability for the user to suppress them, maybe in conjunction with #392. 

Lints in this PR :
* [x] Non-defined scope variables
* [x] Unused user-defined struct fields
* [x] Unused user-defined enum constructors
* [x] Unused scope variables


AltGr (approved):
This is pretty neat!

Some thoughts for later:
- I think it may soon become useful to add numbers to the warnings and a way to selectively enable or disable them
- At the moment each warning does a full AST traversal, this is clearly easier to read and manage, but if at some point we start having a large number of them it will be more efficient to wrap them into a single traversal. Probably over-engineering at this point to define a specific folding data-structure and package them all in it though.

PR #439, opened 2023-03-30
Generic mapping function across different ASTs

AltGr:
Used in lcalc/compile_with_exceptions only at the moment

This adds an inner type parameter to the AST nodes, but it's hidden from the
outside (`('a, 't) naked_gexpr` is now actually defined from `('a, 'a, 't)
base_gexpr`) and no concern, unless in the case where you actually want do do
type-transforming recursion — which becomes possible.

And this allows to write extremely light-weight transformations in the compiler
passes, where the terms that actually get rewritten are no longer mixed in a sea
of trivial rewritings :)

### DIFF
--- a/compiler/shared_ast/expr.mli
+++ b/compiler/shared_ast/expr.mli
@@ -199,6 +199,30 @@ val map :
         f e
     ]} *)
 
+val map_raw :
+  f:(('a, 'm1) gexpr -> ('b, 'm2) boxed_gexpr) ->
+  fop:('a Op.t -> 'b Op.t) ->
+  floc:('a glocation -> 'b glocation) ->
+  (('a, 'b, 'm1) base_gexpr, 'm2) Marked.t ->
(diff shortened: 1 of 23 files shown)


compiler/shared_ast/expr.mli
```diff
@@ -197,6 +199,30 @@ val map :
...
+  floc:('a glocation -> 'b glocation) ->
+  (('a, 'b, 'm1) base_gexpr, 'm2) Marked.t ->
+  ('b, 'm2) boxed_gexpr
+(** Lower-level version of shallow [map] that can be used for transforming the
+    type of the AST. See [Lcalc.Compile_without_exceptions] for an example. The
+    structure is like this:
+
+    {[
+      let rec translate = function
+        | SpecificCase e -> TargetCase (translate e)
```
denismerigoux, line 214: Very neat description

denismerigoux (approved):
Thanks @AltGr, that's good for me!

PR #446, opened 2023-04-06
Overseas housing benefits : last fixes and tests

denismerigoux:
Yet another batch of fixes to the code of the housing benefits example after new legal information, plus minor changes to the compiler.


denismerigoux:
@AltGr you can review this one now for the parts that touch the compiler :)

compiler/catala_utils/pos.ml
```diff
@@ -96,15 +96,10 @@ let to_string (pos : t) : string =
...
-      (e.Lexing.pos_cnum - e.Lexing.pos_bol)
-  else
-    Printf.sprintf "%s:%d.%d-%d.%d:" s.Lexing.pos_fname s.Lexing.pos_lnum
-      (s.Lexing.pos_cnum - s.Lexing.pos_bol)
-      e.Lexing.pos_lnum
-      (e.Lexing.pos_cnum - e.Lexing.pos_bol)
+  Printf.sprintf "%s:%d.%d-%d.%d:" s.Lexing.pos_fname s.Lexing.pos_lnum
+    (s.Lexing.pos_cnum - s.Lexing.pos_bol + 1)
+    e.Lexing.pos_lnum
+    (e.Lexing.pos_cnum - e.Lexing.pos_bol + 1)
```
AltGr, line 102: The same-line format was too tedious to parse ?
(this is a subset of the [GNU format](https://www.gnu.org/prep/standards/standards.html#Formatting-Error-Messages) so I was assuming it would be straight-forward everywhere ; might have been worth adding this pointer in a comment, on second thought)

Well-spotted for the off-by-one, indeed the standard is to start from col. 1 and I was starting from 0. :+1:

compiler/desugared/dependency.ml
```diff
@@ -58,19 +65,22 @@ module Vertex = struct
     | Var (x, sx), Var (y, sy) ->
       ScopeVar.equal x y && Option.equal StateName.equal sx sy
     | SubScope x, SubScope y -> SubScopeName.equal x y
-    | (Var _ | SubScope _), _ -> false
+    | Assertion a, Assertion b -> Ast.AssertionName.equal a b
+    | _, _ -> false
```
AltGr, line 69:
```suggestion
    | (Var _ | SubScope _ | Assertion _), _ -> false
```
I prefer to be conservative for these, although it's a bit redudant, because otherwise you won't get warned when a new case is added.

compiler/desugared/linting.ml
```diff
@@ -44,6 +44,62 @@ let detect_empty_definitions (p : program) : unit =
...
 
+(* To detect rules that have the same justification and conclusion, we create a
+   set data structure with an appropriate comparison function *)
+module RuleExpressionsMap = Map.Make (struct
+  type t = rule
+
+  let compare x y =
+    let xj, xj_mark = x.rule_just in
+    let yj, yj_mark = y.rule_just in
+    let just =
```
AltGr, line 55: Note that in `ast.ml` the unbox is done before the comparison ; the two approaches should be equivalent.
A shorter version would cheat by using `Expr.compare` on `EDefault {exceptions=[]; just; cons}`, but meh ;)

compiler/driver.ml
```diff
@@ -567,11 +567,13 @@ let driver source_file (options : Cli.options) : int =
...
   with
-  | Errors.StructuredError (msg, pos) ->
+  | Errors.StructuredError (msg, pos) -> (
     let bt = Printexc.get_raw_backtrace () in
     Cli.error_print "%s" (Errors.print_structured_error msg pos);
     if Printexc.backtrace_status () then Printexc.print_raw_backtrace stderr bt;
-    -1
+    match options.message_format with
+    | Human -> -1
+    | EditorParsable -> 0 (* editors don't suffer a non-zero return code *))
```
AltGr, line 576: That seems like a pretty surprising behaviour to me (and undocumented in the CLI option)
My editor reports if the compilation succeeded or failed depending on the return code, and I would get puzzled by this.

compiler/shared_ast/expr.mli
```diff
@@ -341,6 +341,8 @@ val remove_logging_calls :
     their argument. *)
 
 val format :
+  ?hide_function_body:bool
+    (** if [true], prints "<function>" for [EAbs] nodes *) ->
```
AltGr, line 345: Ah that will conflict with my PR that simplified the expr printing interface.
To be consistent we probably want to keep this in `Print.expr` (which has an added `unit` arg to make the optional arguments usable) and discard them in the simpler `Expr.format` (which fits more nicely in "%a" fmts), if that sound OK ?

AltGr (commented):
Nice! Just a few notes in this review, except from the return value behaviour which could be confusing.

I am wondering if your editor reacts specifically to the high return code, in general "normal" failures should be reported by values <100 while high values are meant for serious system errors or signals. Maybe worth trying returning e.g. 10 in our normal error case.

compiler/shared_ast/print.ml
```diff
@@ -457,7 +460,8 @@ let rec expr_aux :
       Cli.format_with_style [List.hd colors] fmt ")")
     else expr colors fmt e1
   in
-  let lhs ex = paren ~rhs:false ex in
+  let default_punct color fmt s = Cli.format_with_style [color] fmt s in
```
AltGr, line 463: This is lacking a `Format.pp_print_as`  (like `punctuation`) which leads to Format column calculation errors when used with unicode chars.

compiler/catala_utils/pos.ml
denismerigoux: I didn't know there was a GNU standard for error messages, we should follow it and it should definitely in a comment! The same-line format that was here before made it slightly more complex to parse error messages (because you had to check whether the position was on the same line or not), so I just made it more simple to parse.

compiler/desugared/linting.ml
denismerigoux: Here we're doing only one `unbox` instead of two, which seems more optimal to me because `unbox` appears to be costly.

denismerigoux: I applied the same "one-unboxing-only" in `ast.ml`.

compiler/catala_utils/cli.ml
```diff
@@ -122,6 +126,20 @@ let color =
...
+  Arg.(
+    value
+    & opt ~vopt:Human message_format_opt Human
+    & info ["message_format"]
+        ~doc:
+          "Selects the format of error and warning messages emitted by the \
+           compiler. If set to $(i,human), the messages will be nicely \
+           displayed and meant to be read by a human. If set to $(i, editor), \
+           the messages will be rendered in a way that is easily parsable by \
+           IDEs.")
```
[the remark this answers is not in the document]
denismerigoux: So at this point I think we should really refactor our CLI and bundle all outgoing messages in a library like https://gitlab.com/alistair.obrien/grace. Let's talks about it tomorrow with the potential intern.

Issue #194, opened 2022-02-09
Document structure information missing in some positions


denismerigoux:
## The problem

In the Catala compiler, all source code position information contains a list of strings that represents the headings that you have to traverse in the literate document structure to reach the position :

https://github.com/CatalaLang/catala/blob/36777a06f54539c6731af4ebec22b23bed07f8ec/compiler/utils/pos.ml#L15

It allows us to build error messages like this:

```
[ERROR] "requrements_ownership_met": unknown identifier for a variable of scope Section121SinglePerson

   --> examples/us_tax_code/tests/../section_121.catala_en
    | 
275 |   (section121Person1.requrements_ownership_met or
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^
    + Section 121
    +-+ (b) Limitations
      +-+ (2) Special rules for joint returns
        +-+ (A) $500,000 Limitation for certain joint returns
```

Currently, this information about the structural position is filled after parsing using a visitor: 

https://github.com/CatalaLang/catala/blob/36777a06f54539c6731af4ebec22b23bed07f8ec/compiler/surface/fill_positions.ml#L17-L33

However, some position information produced by the parser are currently not visited by this function because they're marked `@opaque`. For instance, in 

https://github.com/CatalaLang/catala/blob/36777a06f54539c6731af4ebec22b23bed07f8ec/compiler/surface/ast.ml#L351

`RuleName.t` has been built by the parser from a `string * Pos.t` and this `Pos.t` is not visited. When an error message uses this `Pos.t` is does not display the nice hierarchical structure position as it should.

## The solution

This `@opaque` qualifier should be removed and a special visitor defined for the types defined by invocations to the `Uid` functor like `RuleName.t`, etc. This visitor should be integrated with the visitor deriving system in place for `surface.Ast` that uses François Pottier's visitor library.

PR #447, opened 2023-04-07
Print exception tree for fun and profit

denismerigoux:
This features comes from our experience of pair programming. Often we had to go to the drawing board to draw the expression tree and figure out what was going on. With this PR, it's possible to check whether what is actually in the code maps our understanding of how the code should be.

### DIFF
--- a/compiler/desugared/dependency.ml
+++ b/compiler/desugared/dependency.ml
@@ -173,11 +173,11 @@ let build_scope_dependencies (scope : Ast.scope) : ScopeDependencies.t =
       scope.scope_sub_scopes g
   in
   let g =
-    Ast.ScopeDefMap.fold
+    Ast.ScopeDef.Map.fold
       (fun def_key scope_def g ->
         let def = scope_def.Ast.scope_def_rules in
         let fv = Ast.free_variables def in
-        Ast.ScopeDefMap.fold
+        Ast.ScopeDef.Map.fold
           (fun fv_def fv_def_pos g ->
             match def_key, fv_def with
             | ( Ast.ScopeDef.Var (v_defined, s_defined),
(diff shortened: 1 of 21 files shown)


compiler/scopelang/from_desugared.mli
```diff
@@ -16,4 +16,10 @@
 
 (** Translation from {!module: Desugared.Ast} to {!module: Scopelang.Ast} *)
 
-val translate_program : Desugared.Ast.program -> Shared_ast.untyped Ast.program
+val translate_program :
```
AltGr, line 19: Not at all sure about this, but if it can be decoupled in two steps, would it be feasible to have a first function that generates the exception graph, then the `translate_program` taking that graph as parameter ?

This would be more consistent with the way `Desugared.from_surface` works, taking as parameter a name resolution context that has been previously computed. Not sure it's better in general though ?

compiler/shared_ast/definitions.ml
```diff
@@ -45,6 +45,67 @@ module SubScopeName = Uid.Gen ()
 module StateName = Uid.Gen ()
 
 (** {1 Abstract syntax tree} *)
+module DesugaredVarName : sig
```
AltGr, line 48: The `Definitions` module is expected to contain only definitions, which morally allows us to ① have no `.mli` and ② to `open` it throughout.

I know in this case it's harmless since you took care to provide a signature — and also that we're already cheating with this rule a bit with the `Uid.Gen ()` applications — but since the type is not used below in the module, there should be no trouble in putting it in its own file ?

Also:
- why not just `DesugaredVar` ?
- shouldn't this replace `Desugared.Ast.ScopeDef` which is made redundant and could probably be removed ?

AltGr (approved):
This is very nice! I just have some code structure and nitpicking comments.

compiler/desugared/dependency.ml
```diff
@@ -353,17 +357,27 @@ let build_exceptions_graph
         in
         LabelName.Map.update label_of_rule
           (fun rule_set ->
+            let pos =
+              (* We have to overwrite the law info on tis position because the
+                 pass at the surface AST level that fills the law info on
+                 positions only does it for positions inside expressions, the
+                 visitor in [surface/fill_positions.ml] does not go into the
+                 info of [RuleName.t], etc.*)
+              Pos.overwrite_law_info
```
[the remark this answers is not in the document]
denismerigoux: The corresponding issue is: https://github.com/CatalaLang/catala/issues/194.

Issue #443, opened 2023-04-04
Bug in formalizing US Section 121: Five-year periods


rohanpadhye:
Hi folks,

I'm working on some experimental tooling for testing of Catala programs (more on that later), and I encountered what I think is a bug in the [formalization of Section 121 of the US Tax Code](https://github.com/CatalaLang/catala/blob/master/examples/us_tax_code/section_121.catala_en).  Here's a test that demonstrates the bug:

````
> Include: examples/us_tax_code/section_121.catala_en 

The point of this test is to show a usage period that continues AFTER the sale date
for a personal property (presumably a seller renting the property they just sold) and
using it to satisfy the requirements of Section 121.

```catala
declaration scope FiveYearBug:
  scope121a scope Section121SinglePerson

scope FiveYearBug:
  definition scope121a.date_of_sale_or_exchange equals |2021-01-01|
  definition scope121a.gain_from_sale_or_exchange_of_property equals $500,000
  definition scope121a.property_ownage equals [Period {
    # Owned for 2 years until sale date; this looks good
    -- begin: |2019-01-01|
    -- end: |2021-01-01|
  }]
  definition scope121a.property_usage_as_principal_residence equals [Period {
    # Used for 1.5 years until sale date; not enough to meet requirements
    -- begin: |2019-06-01|
    -- end: |2021-01-01|
  }; Period {
    # Rented the property for six months AFTER selling it; this should not count!
    -- begin: |2021-01-01|
    -- end: |2021-06-01|
  }]
  definition scope121a.other_section_121a_sale equals NoOtherSaleWhereSection121aApplied

  assertion not scope121a.requirements_met   # Fail! The requirements appear to be met
```
````

We should expect the requirements to not be met because there wasn't a total of 2-year usage during the 5-year period ***ending on the date of sale/exchange***:

```
### (a) Exclusion

Gross income shall not include gain from the sale or exchange of property if,
during the 5-year period ending on the date of the sale or exchange, such
property has been owned and used by the taxpayer as the taxpayer’s principal
residence for periods aggregating 2 years or more.
```

However, the assertion above fails:

```
$ catala Interpret -s FiveYearBug bug121.catala_en --disable_warnings
[ERROR] Assertion failed

┌─⯈ bug121.catala_en:19.12-42:
└──┐
19 │   assertion not scope121a.requirements_met
[...]

PR #448, opened 2023-04-10
[RFC] Concolic testing for Catala

rohanpadhye:
Hi folks,

First of all, let me say that Catala is a really awesome project. I've been following it for a few months, even though I have no background in computational law.  I mainly do research in automated testing and program analysis. This is a draft PR of what for me has been a fun side project: an automatic test-input generation system for Catala using the Python backend. 

Summary of Concolic plugin:

- Please see the [**README for the Concolic Plugin**](https://github.com/rohanpadhye/catala/blob/concolic/doc/concolic/README.md) which should explain what it does. I will keep updating this file as new questions arise.
- **tl;dr**: The plugin builds on the Python backend to perform simultaneous concrete + symbolic ("concolic") execution using the [DART algorithm](https://doi.org/10.1145/1065010.1065036) and the [Z3 constraint solver](https://pypi.org/project/z3-solver/), which allows systematic generation of test inputs covering all possible execution paths starting from some initial input. A quick look at the [runtime file](https://github.com/rohanpadhye/catala/blob/concolic/runtimes/python/catala/src/catala/concolic.py) should explain how it's implemented.
- The goal of the concolic plugin is twofold: (1) to support automatic bug finding via systematic property-based testing (think: Quickcheck for Catala), and (2) to enable automatic generation of *readable* tests which can help both in debugging Catala code and in adding tricky "examples" to the associated legal text. To support readability, the concolic API allows providing soft constraints which the engine incorporates into a MAX-SAT solver.

Request for Comments:

- This is a ***DRAFT*** PR. I'm mainly opening it for feedback. What do you think? Is this something that could be useful?
- For maintainers: If you would like to merge this in the future, please help me figure out the right way to use the build system to add the plugin. 
  - I am not familiar with Dune, so I am sure that my current approach of copying `concolic.py` to the Python runtime is flawed. Code review and comments are appreciated. 
  - Similarly, the `concolic.ml` file is simply a fork of the `to_python.ml` file, with some changes to function names and binary operators in the emitted Python code. Perhaps there is an easier way to store just the diff instead of having to keep the plugin in-sync with the Python backend.
- Any thoughts on the API or manual effort required in setting up parametric test drivers (e.g., `FuzzSection121SinglePerson` in the README)?

FAQ: How does this compare to the existing Proof backend?
- Honestly, I don't know for sure. I found out about the proof backend after having a prototype of the concolic engine, so I went ahead with it anyway. In the long run, I assume the proof backend would be stronger, but perhaps testing can help bridge some completeness gaps. 
- For example, I am guessing there are cases that are not always easy to support for verification (e.g., date/duration arithmetic, arbitrary sized lists, and recursive scope invocations, if any). Concolic testing, being dynamic, allows for "concretization", where a symbolic expression for a variable can simply be replaced with a concrete value from an execution when dealing with unsupported features. This results in incompleteness, but still allows making progress (which can be further enhanced using random testing).
- The testing approach also has the benefit of generating readable example inputs for various code paths, which can be used for documentation or debugging. This is probably harder to do with static verification conditions.

What else can be added to the concolic plugin?

- [ ] Symbolic arithmetic with dates and durations in months (currently only days and years is supported with any degree of completeness, using months causes concretization)
- [ ] Support for lists and enums (currently you have to write a test driver that takes only primitive inputs and builds the structures in Catala)
- [ ] Random restarts. Currently, when concretization occurs we just get fewer test inputs. As per the [DART algorithm](https://doi.org/10.1145/1065010.1065036), we should be able to restart concolic execution with random inputs to try and increase coverage. This is also the spirit behind whitebox fuzzing.
- [ ] Adding more parametrized tests with assertions in Catala files to demonstrate property-based testing. 
- [ ] Providing a cleaner run API (e.g. `catala Concolic -s TestScope ...`) and pretty-printed colorful output.
- [ ] Minimize the generated test input corpus. Since concolic testing targets path coverage, we produce a large number of similarly-differing test inputs for programs that have a lot of diamonds in the control-flow graph. Some users may prefer to just get a set of test inputs with high line coverage or branch coverage. It should be possible to minimize the test inputs produced by the concolic engine with such coarser coverage criteria using a greedy solution to the [set cover problem](https://en.wikipedia.org/wiki/Set_cover_problem).


rohanpadhye:
Thanks for your comments @denismerigoux. I have enabled editing by maintainers, so feel free to push your commits enabling the proper formatting to the concolic branch.

It's great to know that test generation is something that seems like it could be useful. I agree that the interface as it stands is not very lawyer friendly, but for now I was just developing a proof of concept to see if it could be useful on any of the existing formalizations. It would certainly be much more user-friendly to have the concolic engine run as a Catala Interpreter, bypassing the Python driver entirely. I wonder if this would need extensions to the Catala syntax to specify the parametrized tests with constraints/preferences on the entry points (I can imagine some combination of `assume` and `prefer` statements similar to the current `assert`). I don't think I will have the time/resources to figure out how to port everything to OCaml and operate on the SCalc or DCalc IRs (one reason I chose Python is the relative ease of performing instrumentation on the generated code; the DART algorithm itself I had already implemented for an assignment in a class that I teach). However, if this is something that anyone else wants to pursue, I am happy to collaborate in the development or assist in design decisions.

For property-based testing, the README currently does not have enough examples to demonstrate what I had in mind. I was interested in exploring whether it makes sense to specify some high-level consistency properties (e.g., same income dollar is not taxed twice, same expense dollar is not allowed to be deducted twice) or correctness properties for a given scope (e.g., I identified #443 when trying to express properties about usage after sale not affecting the requirements being met). All of these can be expressed as assertions over a collection of executions of the same scope. Perhaps the proof backend is easier to use for something like this. Let me work on coming up with some more concrete examples and I'll add to the docs.

denismerigoux:
I'm closing this after our conversation during the last summer. Thanks again for pointing us the usefulness of concolic testing ! We'll likely implement it directly in OCaml alongside the compiler.

PR #453, opened 2023-04-14
Add experimental lazy interpreter as a plugin

AltGr:
To try it (without installing Catala):
```shell-session
$ make plugins
$ export CATALA_PLUGINS=_build/default/compiler/plugins
$ dune exec -- catala lazy
examples/aides_logement/tests/tests_calcul_apl_locatif.catala_fr -s
Exemple2
```

Keep in mind that this is a work-in-progress prototype :)

### DIFF
--- a/compiler/plugins/lazy_interp.ml
+++ b/compiler/plugins/lazy_interp.ml
@@ -1,0 +1,273 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2023 Inria, contributor:
+   Louis Gesbert <louis.gesbert@inria.fr>.
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you may not
+   use this file except in compliance with the License. You may obtain a copy of
+   the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+open Catala_utils
+open Shared_ast
+
+(* -- Definition of the lazy interpreter -- *)
+
+let log fmt = Format.ifprintf Format.err_formatter (fmt ^^ "@\n")
+let error e = Errors.raise_spanned_error (Expr.pos e)
+let noassert = true
+
+type laziness_level = {
+  eval_struct : bool;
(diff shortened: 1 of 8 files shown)


AltGr:
Example output (very redundant since this is with lazy evaluation actually turned off and re-played each time):
image

AltGr:
I had marked this as draft, but while this correctly describes the state of the interpreter, it's a plugin so it may be OK to merge as-is so that everyone can experiment with it. The changes to the main tree are minimal:

- small changes to the plugin API
- added support for `dcalc` plugins (this is the first patch)
- exposed a couple functions in `.mli`s

compiler/plugins/lazy_interp.ml
```diff
@@ -0,0 +1,273 @@
...
+      (* if false, stop evaluating as soon as you can discriminate with
+         `EEmptyError` *)
+}
+
+let value_level = { eval_struct = false; eval_op = true; eval_default = true }
+
+module Env = struct
+  type 'm t =
+    | Env of
+        ((dcalc, 'm mark) gexpr, ((dcalc, 'm mark) gexpr * 'm t) ref) Var.Map.t
```
denismerigoux, line 41: I'm curious, could you put a comment describing what you're putting in the `Env.t` ?

compiler/plugins/lazy_interp.ml
```diff
@@ -0,0 +1,273 @@
...
+      let v_env =
+        try Env.find v env
+        with Not_found ->
+          error e0 "Variable %a undefined [@[<hv>%a@]]" Print.var_debug v
+            Env.print env
+      in
+      let e, env1 = !v_env in
+      let r, env1 = lazy_eval ctx env1 llevel e in
+      if not (Expr.equal e r) then (
+        log "@[<hv 2>{{%a =@ [%a]@ ==> [%a]}}@]" Print.var_debug v
```
denismerigoux, line 86: When will this `log` be shown? With the `-d` flag?

denismerigoux (approved):
I'm okay with merging this! This is so cool, I hope you'll have time to get back to this and comment your code to show why it works from a more theoretical perspective.

compiler/plugins/lazy_interp.ml
AltGr: I will! Until then: an `env` maps a variable to an expression _and_ the corresponding captured env.
I don't want to think about how much more complicated that would have been in an imperative setting :D

The `ref` part is only used for laziness: the raw expression is first added to the environment, then we can replace it once computed. We don't use `lazy` because we want more control (in particular, we can reduce the same expression several times with different levels)

PR #462, opened 2023-05-02
Some more fixes and improvements to the printer

AltGr:
- continue the printer improvements from previous PR
- fixes some issues that remained with it
- clean up `Print.expr` interface (context no longer needed)
- some prettification of matches (internal lambdas are inlined)


compiler/lcalc/to_ocaml.ml
```diff
@@ -342,7 +342,8 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
     Format.fprintf fmt
       "@[<hov 2>%a@ @[<hov 2>{filename = \"%s\";@ start_line=%d;@ \
        start_column=%d;@ end_line=%d; end_column=%d;@ law_headings=%a}@]@ %a@]"
-      Print.operator op
+      (Print.operator ~debug:true)
```
denismerigoux, line 345: Why `debug:true` here?

compiler/plugins/lazy_interp.ml
```diff
@@ -84,9 +84,9 @@ let rec lazy_eval :
       let r, env1 = lazy_eval ctx env1 llevel e in
       if not (Expr.equal e r) then (
         log "@[<hv 2>{{%a =@ [%a]@ ==> [%a]}}@]" Print.var_debug v
-          (Print.expr ~debug:true ctx)
+          (Print.expr' ~debug:true ())
```
denismerigoux, line 87: Why `expr'` here? I find this `expr/expr'` quite confusing.

compiler/shared_ast/expr.ml
```diff
@@ -536,11 +536,7 @@ and equal : type a. (a, 't) gexpr -> (a, 't) gexpr -> bool =
   | EArray es1, EArray es2 -> equal_list es1 es2
   | ELit l1, ELit l2 -> l1 = l2
   | EAbs { binder = b1; tys = tys1 }, EAbs { binder = b2; tys = tys2 } ->
-    Type.equal_list tys1 tys2
-    &&
-    let vars1, body1 = Bindlib.unmbind b1 in
-    let body2 = Bindlib.msubst b2 (Array.map (fun x -> EVar x) vars1) in
-    equal body1 body2
+    Type.equal_list tys1 tys2 && Bindlib.eq_mbinder equal b1 b2
```
denismerigoux, line 539: Good catch

compiler/shared_ast/expr.ml
```diff
@@ -702,18 +698,27 @@ let rec free_vars : ('a, 't) gexpr -> ('a, 't) gexpr Var.Set.t = function
 
 let rec skip_wrappers : type a. (a, 'm) gexpr -> (a, 'm) gexpr = function
   | EApp { f = EOp { op = Log _; _ }, _; args = [e] }, _ -> skip_wrappers e
+  | EApp { f = EApp { f = EOp { op = Log _; _ }, _; args = [f] }, _; args }, m
+    ->
+    skip_wrappers (EApp { f; args }, m)
   | EErrorOnEmpty e, _ -> skip_wrappers e
+  | EDefault { excepts = []; just = ELit (LBool true), _; cons = e }, _ ->
+    skip_wrappers e
```
denismerigoux, line 706: I would put that as an optimization in `shared_ast/optimizations.ml` rather than just a `skip_wrapper` thing.

tests/test_io/good/all_io.catala_en
```diff
@@ -19,18 +19,31 @@ scope A:
...
+let scope A (A_in: A_in
+                     {c_in:
+                     integer;
+                     d_in:
+                     integer;
+                     e_in:
+                     unit → integer;
+                     f_in:
+                     unit → integer}): A
+                                           {b:
```
denismerigoux, line 31: This is still weird!

denismerigoux (changes requested):
Thanks @AltGr I'm not a fan of the `expr/expr'` interface, can you give these two functions more meaningful names?

compiler/shared_ast/expr.ml
AltGr: Well the previous version was OK, but I had'nt discovered this function from `Bindlib` yet :)

compiler/shared_ast/expr.ml
AltGr: It's useful besides optimisations: typically, you want to use it before matching on some constructs to ignore intermediate wrappers. Or sometimes when you want a concise printing of an expression.

Note also that it's shallow, as mentionned in the mli: it will remove any number of nested wrappers, but only "on top" of the expression.

compiler/plugins/lazy_interp.ml
AltGr: They're the same function but with different arguments: the trouble is that you can't have automatic elimination of the optional `?debug` argument when you use the `Format.printf "%a" Print.expr e` form.

- The `expr` form skips the debug arguments, and so can be used directly in this form
- The `expr'` form has an added unit argument, allowing `Format.printf "%a" (Print.expr' ()) e`

I have seen this `f` / `f'` in a few projects, but indeed it's not very nice ; I'll try to find better naming.

tests/test_io/good/all_io.catala_en
AltGr: Indeed! I focused on the expression printer for now, but the types printer also needs more love!

compiler/lcalc/to_ocaml.ml
AltGr: Ah this is correct but indeed a bit unclean ;
- `Print.operator` prints the full operator symbols when `debug=true` (_i.e._ `+$`, `+^`, etc.)
- and the "simplified" operators otherwise, which are a bit better for readability, but won't work for OCaml output.

but that doesn't match the name for the `debug` flag at all anymore ; I'll add a different function (even if it does the same under the hood)

tests/test_io/good/all_io.catala_en
AltGr: Actually that was the scope-level printer that was added recently that had some issue; it should now be improved as well.

Issue #451, opened 2023-04-13
Use [ocolor](https://opam.ocaml.org/packages/ocolor/) ?


AltGr:
For now, we are using `ANSITerminal` throughout for color printing to the console; and do almost all our printing using `Format` pretty-printers.

Those don't combine well:
- `ANSITerminal` only works on strings, so we cannot apply colors on things that are normally printed using formatters (unless we stringify them once, but that completely breaks the model of `Format`)
- `Format` assumes that string size in chars and number of screen columns are the same, and is completely confused by ANSI color codes that make strings longer (unless explicitely using `Format.pp_print_as` but that's not very convenient ; see for example [here](https://github.com/CatalaLang/catala/pull/450/files#diff-fc251333d2c1f3a7851642d74cace384bdfb7aaec7b7543b967d0b2608a60fdc) for a fix that was needed)

I haven't used it (I am familiar with their approach though), but `ocolor` uses `Format` support for semantic tags to integrate color into it in a much nicer and consistent way; formatting tags are written like `printf "@{<red>red@}"` and the above issues disappear.

There are additional bonuses, like the fact that nesting is correctly handled, hiding ANSI color low-level reset codes (e.g. `"@{<blue>@{<red>This is red@} this is blue@}"`)

PR #470, opened 2023-05-30
Improve compiler messages dev interface

denismerigoux:
In our quest to have beautiful error messages maybe powered in the future by https://gitlab.com/alistair.obrien/grace, we need a sensible API for emitting error messages in the compiler. This PR tries to clean some of the mess and create an error message abstraction. Also @AltGr will make the switch to `ocolor` (#451).


AltGr:
This is good to go, I am proceeding with the rebase

AltGr:
CI passed, but…
```
npm install

added 7 packages, and audited 8 packages in 21m
```
This really takes a lot of time (60% of the total run), we need to find a way to skip or cache this.

PR #472, opened 2023-06-03
Improve unused scope variable detection

denismerigoux:
The lint now relies on reachability analysis from the scope outputs in the dependency graph rather than a simple check whether the variable appears somewhere.


AltGr (approved):

AltGr:
Very nice!

Issue #473, opened 2023-06-07
Tutorial's Article 3 conflicts with Article 2


escott-mitre:
In the [tutorial](https://catala-lang.org/en/examples/tutorial), Article 3 is impelmented with

```catala
scope IncomeTaxComputation:
  definition fixed_percentage under condition
    individual.number_of_children >= 2
  consequence equals 15 %
  # Writing 15% is just an abbreviation for "0.15".
```

and the commentary says "When the Catala program will execute, the right definition will be dynamically chosen by looking at which condition is true."

But if I try adding a test like so:

```catala
declaration scope TestChildren:
  tax_computation scope IncomeTaxComputation
  output income_tax content money

scope TestChildren:
  definition tax_computation.individual equals Individual {
    -- income: $4,000
    -- number_of_children: 2
  }

  definition income_tax equals tax_computation.income_tax
  assertion income_tax = $600
```

There is a conflict:

<img width="893" alt="image" src="https://github.com/CatalaLang/catala/assets/90326618/2f6c35c6-2237-4d52-9e63-2d88550c2b0a">

 1. If I understand correctly, this is the expected behavior, because exceptions always must be explicitly specified in Catala if the conditions of two definitions are not mutually exclusive?
 2. If so, then the tutorial is confusing at this point: "the right definition will be dynamically chosen," but in this example this cannot be true, because the given definitions conflict.

denismerigoux:
Hi @escott-mitre, you correctly understood how things worked :) If you read the sentence of the tutorial that comes just after https://catala-lang.org/en/examples/tutorial#examples/tutorial_en/tutorial_en.catala_en-228, it says : 

> When the Catala program will execute, the right definition will be dynamically chosen by looking at which condition is true. A correctly drafted legislative source should always ensure that at most one condition is true at all times. However, if it is not the case, Catala will let you define a precedence on the conditions, which has to be justified by the law. But we will see how to do that later.

Just after the bit "the right definition will be dynamically chosen" that you quoted, you can see that the scenario that you're describing is alluded to by the tutorial. If you think that these next sentences are not clear enough, feel free to propose a PR to improve the wording. The way to define exceptions is covered later in the tutorial, here : https://catala-lang.org/en/examples/tutorial#The%20Catala%20language%20tutorial-Defining%20exceptions%20to%20rules.

PR #483, opened 2023-06-23
Tweak tutorial language in response to #473.

escott-mitre:
Change some wording so it's easier for users to understand that the conflicting conditional definitions in the tutorial will fail (see #473).

I tried to keep the French and English tutorials in sync with this PR, but _my French skills are poor, so please check the edits I made to the French tutorial for grammar!_


denismerigoux (approved):
Thanks very much!

PR #484, opened 2023-06-28
Refactor the main `Driver` module, and add a 'modules' plugin

AltGr:
- Use separate functions for successive passes in module `Driver.Passes`
- Use other functions for end results printing in module `Driver.Commands`

As a consequence, it is much more flexible to use by plugins or libs and we
no longer need the complex polymorphic variant parameter.

This patch leverages previous changes to use Cmdliner subcommands and
effectively specialises the flags of each Catala subcommand.

Other changes include:

- an attempt to normalise the generic options and reduce the number of global
references. Some are ok, like `debug` ; some would better be further cleaned up,
e.g. the ones used by Proof backend were moved to a `Proof.globals` module and
need discussion. The printer no longer relies on the global languages and prints
money amounts in an agnostic way.

- the plugin directory is automatically guessed and loaded even in dev setups.
Plugins are shown by the main `catala` command and listed in `catala --help`

- exception catching at the toplevel has been refactored a bit as well; return
codes are normalised to follow the manpage and avoid codes >= 128 that are
generally reserved for shells.


compiler/driver.ml
```diff
@@ -16,21 +16,206 @@
...
+      =
+    match Ident.Map.find_opt scope ctxt.typedefs with
+    | Some (Desugared.Name_resolution.TScope (uid, _)) -> uid
+    | _ ->
+      Message.raise_error
+        "There is no scope @{<yellow>\"%s\"@} inside the program." scope
+
+  (* TODO: this is very weird but I'm trying to maintain the current behaviour
+     for now *)
+  let get_random_scope_uid (ctxt : Desugared.Name_resolution.context) =
```
denismerigoux, line 218: Yes you're right about what this function does

compiler/driver.ml
```diff
@@ -43,47 +228,27 @@ let get_scope_uid
...
-          exec_opt
-            (compile
-            @@ whole_string
-            @@ seq [group (rep1 (compl [char '.'])); char '.'; group (rep1 any)]
-            )
-            name)
-      with
-      | None -> name, None
-      | Some groups -> Re.Group.get groups 1, Some (Re.Group.get groups 2)
+      match String.index_opt variable '.' with
```
denismerigoux, line 238: ha ha ha you're better at string manipulation than me :)

compiler/driver.ml
```diff
@@ -576,58 +849,60 @@ let main () =
...
-         ~extra:plugins)
-  in
-  exit return_code
+  let command = catala_t plugins in
+  let open Cmdliner in
+  match Cmd.eval_value ~catch:false ~argv command with
+  | Ok _ -> exit Cmd.Exit.ok
+  | Error _ -> exit Cmd.Exit.cli_error
+  | exception Cli.Exit_with n -> exit n
+  | exception Message.CompilerError content ->
```
denismerigoux, line 882: Ok the error handling is here now

compiler/driver.ml
denismerigoux, line 1: Overall this a good refactoring, I like how each pass is very cleanly individualized. Congrats!

compiler/plugins/modules.ml
```diff
@@ -0,0 +1,215 @@
...
+  let ml_file =
+    gen_ocaml options link_modules optimize check_invariants (Some modname) None
+  in
+  let flags = ["-I"; Lazy.force runtime_dir] in
+  let shared_out = basename ^ ".cmxs" in
+  Message.emit_debug "Compiling OCaml shared object file @{<bold>%s@}..."
+    shared_out;
+  run_process "ocamlopt" ("-shared" :: ml_file :: "-o" :: shared_out :: flags);
+  (* let byte_out = basename ^ ".cmo" in
+   * Message.emit_debug "Compiling OCaml byte-code object file @{<bold>%s@}..." byte_out;
```
denismerigoux, line 123: Dead code?

compiler/verification/globals.mli
```diff
@@ -0,0 +1,29 @@
...
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+(** This sub-lib relies on global refs in many places. This should be cleaned
+    up. *)
+
+val setup : optimize:bool -> disable_counterexamples:bool -> unit
+
+val optimize : unit -> bool
+(** Not sure what that means in this context: does it turn on optimisations in
```
denismerigoux, line 24: It turns on optimisations of VCs before sending them to Z3 or Mopsa. It does not trigger special handling that is dependent on the input code having been optimised in my understanding.

compiler/verification/globals.mli
```diff
@@ -0,0 +1,29 @@
...
+    the verif backend, or trigger special handling that is dependent on the
+    input code having been optimised ?? *)
+
+val disable_counterexamples : unit -> bool
+(** This should really be passed along through arguments *)
```
denismerigoux, line 29: I agree

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
```diff
@@ -75,24 +75,26 @@ champ d'application Exemple2:
 ```catala-test-inline
 $ catala Interpret -s Exemple1
-[RESULT] montant = 96.48 €
+[RESULT] montant = ¤96.48
```
denismerigoux, line 78: Not OK

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
```diff
@@ -75,24 +75,26 @@ champ d'application Exemple2:
...
 [RESULT] Computation successful! Results:
-[RESULT] montant = 85.00 €
+[RESULT] montant = ¤85.00
 ```
 
 ```catala-test-inline
 $ catala Interpret -s Exemple2 --avoid_exceptions
-[RESULT] Computation successful! Results:
-[RESULT] montant = 85.00 €
+catala: unknown option '--avoid_exceptions'.
```
denismerigoux, line 95: This one should not fail like this

denismerigoux (commented):
Thanks Louis ! I like the refactoring but I think this PR could be improved by finishing the refactoring about command-specific options and fixing the behaviour of the compiler on test (their output should not change apart from the error codes).

compiler/catala_utils/cli.mli
```diff
@@ -17,111 +17,98 @@
...
-  disable_counterexamples : bool;
-  check_invariants : bool;
-  optimize : bool;
-  ex_scope : string option;
-  ex_variable : string option;
-  output_file : string option;
-  closure_conversion : bool;
-  print_only_law : bool;
-  link_modules : string list;
+type options = private {
```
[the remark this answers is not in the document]
AltGr: In this context, it means you're not allowed to mutate the fields from outside the module.

compiler/verification/globals.mli
AltGr: ok so maybe it should be a separate flags ? Weird to enforce that this is turned on iff optimisations in dcalc/lcalc are

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
AltGr: Ah yes missed that one, the option must be removed (it makes no sense with `--interpret`)

compiler/verification/globals.mli
denismerigoux: @rmonat @R1kM what do you think?

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
AltGr: There was quite a bit of pondering for this change, so let me explain my reasoning before insisting that you disagree :D

- We have a printer for expressions in the different variants of the AST ; that printer does not correspond to the Catala syntax, and is not localised ; we use it mainly for debugging.
- There was a single spot where that printer actually relied on a global reference, affecting only the printing of money literals, to guess what unit the money should be printed as.

My point actually is — I've met this issue already with the experimental explanations plugin — that this printer isn't suitable for printing end-user results and that we should have a separate (probably fully localised) one ; that way we could have a consistent french syntax (`10,25 €` with a comma) or english syntax ($10.25 with a dot), among other things. (note that for results, this could be limited to literals hence fairly quick to write. for the explanation backend where we print formulas it'll be much more work, but I'm convainced that will be necessary at some point)

So for the sake of consistency I feel we need to use an undeterminate money symbol here. That's exactly what the unicode character `¤` is  :). It also means we get consistent output for a given program, not depending on configuration.

Of course, I fully understand that this symbol is little known and that users will be confused by this output ;) . The reason I am pushing this already is because the refactoring required that I got rid of the corresponding global reference. I can write a localised literal printer asap to recover better `[RESULT]` printing :)

AltGr:
> finishing the refactoring about command-specific options

Hm that should be about complete with this PR, what specifically do you have in mind ?

compiler/lcalc/compile_with_exceptions.ml
```diff
@@ -64,8 +64,10 @@ and translate_expr (ctx : 'm ctx) (e : 'm D.expr) : 'm A.expr boxed =
     Expr.ecatch (translate_expr ctx arg) EmptyError
       (Expr.eraise NoValueProvided m)
       m
-  | EDefault { excepts = [exn]; just; cons } when !Cli.optimize_flag ->
-    (* FIXME: bad place to rely on a global flag *)
+  | EDefault { excepts = [exn]; just; cons }
+  (* Specific optimisation for this case *) ->
+    (* FIXME: this case used to be disabled when optimisations are disabled, but
```
denismerigoux, line 69: Like this FIXE here

compiler/verification/globals.ml
```diff
@@ -0,0 +1,28 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+let optimize_ref = ref false
+let disable_counterexamples_ref = ref false
+
+(** This sub-lib relies on global refs in many places. This should be cleaned
```
denismerigoux, line 21: Or this hack here?

compiler/lcalc/compile_with_exceptions.ml
AltGr: Ah right here I am not sure what is best: I'd like to keep the case unconditionally (without checking for the optimisations flag) but it's also possible to pass the flag along if it makes sense.

Subsidiary question: if we enable optimisations in `lcalc`, doesn't this term get rewritten in the same way by the following optimisation pass ? In which case just removing this case would yield the same results as before (both with and without `-O`)

compiler/verification/globals.ml
AltGr: ah yes this needs to be cleaned up but I was a bit afraid to touch the verification code, which is why I left it like that but clearly delimited. As proposed it is on par with the previous behaviours.

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
denismerigoux: >  I can write a localised literal printer asap to recover better [RESULT] printing :)

OK for deleting the global reference but I would argue the localised literal printer should be in this PR (with a local optional parameter that defaults to English), I don't want to have a temporary state of Catala with `¤` everywhere :)

compiler/lcalc/compile_with_exceptions.ml
denismerigoux: > In which case just removing this case would yield the same results as before (both with and without -O)

Yeah you're right, let's just delete this code!

compiler/verification/globals.ml
denismerigoux: Ok then we can leave it as is for the moment :)

examples/aides_logement/tests/tests_calcul_al_accession_propriete.catala_fr
AltGr: Fair enough :wink:

compiler/driver.ml
```diff
@@ -16,21 +16,206 @@
...
+    Message.emit_debug "Disambiguating...";
+    let prg = Desugared.Disambiguate.program prg in
+    Message.emit_debug "Linting...";
+    Desugared.Linting.lint_program prg;
+    prg, ctx
+  (* Note: we forward the name resolution context throughout in order to locate
+     uids from strings. Maybe a reduced form should be included directly in
+     [prg] for that purpose *)
+
+  let scopelang options ~link_modules :
```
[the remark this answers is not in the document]
AltGr: This approach resulted in a much simpler interface (see the `Commands` module: one function call to get to the expected AST instead of repeating the same sequence and passing the proper options and results along) ; also function calls allow for a more flexible call graph and this feels easier to read overall ; the situation could be different if we had passes with similar types that could be optional or reordered, but such cases are limited in Catala.

AltGr:
Added a user-facing value printer. A few questions remain open:
- see my comment above about the tests that were purposefully returning functions (note that another possibility would be to extend the scope of the localised printer; that will need doing anyway for the explanation plugin)
- I don't use the output functions from the runtime, that are left unchanged but may become redundant
- I have been exposing the types in the runtime to be able to print them from the outside, but that can be discussed

About points 2 and 3, mid-term I'd like to remove the printers from the runtime (limiting it to what the catala generated code strictly requires) but provide them as a separate module that would be linked by the programs that call compiled Catala code. That would include the json conversion functions, etc.

tests/test_scope/good/scope_call4.catala_en
```diff
@@ -25,41 +25,13 @@ scope RentComputation:
...
-[RESULT]
-f1 = λ (x: integer) →
-  error_empty ⟨true ⊢ let x1 : integer = x + 1 in
-                      error_empty ⟨true ⊢ x1 + 1⟩⟩
-[RESULT]
-f2 = λ (x: integer) →
-  error_empty ⟨true ⊢ let x1 : integer = x + 1 in
-                      error_empty ⟨true ⊢ x1 + 1⟩⟩
+[RESULT] f1 = <function>
+[RESULT] f2 = <function>
```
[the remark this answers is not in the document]
AltGr: Last patch does just that

Issue #485, opened 2023-06-30
clerk should collect tests in included files


denismerigoux:
Consider the following situation : 

* a `code.catala_fr` file containing the code you want to test
* several `tests1.catala_fr`, `test2.catala_fr` files containing your test cases but that don't include `code.catala_fr`
* a `bundle.catala_fr` that includes `code.catala_fr` and then all the test files.

This situation is useful because calling `catala LateX bundle.catala_fr` creates a big PDF with all the code and test cases in it, something you can't do with the scheme where each of of the `test<n>.catala_fr` includes `code.catala_fr` of their own.

In that situation, how to have `clerk` execute all the tests in `test<n>.catala_fr` ? You have to use `clerk test bundle.catala_fr` because you can't call Catala independantly on `test<n>.catala_fr` which don't include `code.catala_fr`.

However `clerk test bundle.catala_fr` fails because `clerk` does not support collecting `catala-test-inline` declarations *across file inclusion*. This feature should be added to `clerk`.

AltGr:
#491 is a naive attempt to implement this that begs for further discussion :)

PR #491, opened 2023-07-18
Implement reversing the dependencies for tests

AltGr:
This works for the `--reset` option, but after trying it out I see several
problems with this approach: we should probably think a bit more about it.

1. Dependencies for tests are now defined from the outside. This means that the
   result of a single test may become inconsistent if included from different
   places.
2. This makes it pretty hard to run a single test (one idea is to allow mutual
   includes, and silently ignore loops; but that may cause a risk for
   inconsistenncies and would need support from Catala itself)
3. Another issue comes from the testing engine: `clerk runtest` has a two modes:
   - one where the output is printed back to stdout (to be diff'd with the
     original file with an external tool)
   - with `--reset`, the source file is updated with the new output of the
     inline tests
   
   Now the tests may stretch across different files. The `--reset` case is
   handled and will update all included files. This works but might be a problem
   because the targets of the build rule become hard to trace. For the first case,
   on the other hand, the interface no longer makes sense with multiple files --
   the only consistent option would be not to run included tests there.

With all this in mind, we should either rethink the testing scheme, or find a
different way to provide the end-user feature planned in #485.

Some ideas to work around this:

- provide a specific non-recursive or literate-only include ? This is simplest
  and would make `pdf` generation possible without affecting the current test run
  process.
- (more complex) allow to specify dependencies on the command-line for running tests:
  * includes are defined as in the example (bundle includes test files)
  * individual test runs need a `--include bundle.catala_fr` flag with special
    handling for avoiding re-including themselves
  * Clerk uses detection of tests through inclusions to generate testing rules
    for each test included in `bundle.catala_fr`. It might be possible to
    implement some safeguards against 1. above.

Not sure if the 2nd idea is worth the trouble, but it might actually be
interesting if combined with modules: the bundle compiled to a module would
naturally exclude any inline tests, and the tests would just need the flag to
link that module upon execution. If we go that route, this PR can probably be
discarded but the code for detecting inclusions (which is the main part) can be
kept (and moved to Clerk proper instead of the specific `clerk runtest` command)

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -171,7 +171,7 @@ let readdir_sort (dirname : string) : string array =
     let dirs = Sys.readdir dirname in
     Array.fast_sort String.compare dirs;
     dirs
-  with Sys_error _ -> Array.make 0 ""
+  with Sys_error _ -> [||]
 
 type test = {
   text_before : string;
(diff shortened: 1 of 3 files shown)


denismerigoux:
Thanks for thinking about this, your points are all valid and I hadn't thought it all out. Let's discuss this when you come back!

AltGr:
Superseded by #497

PR #495, opened 2023-08-04
Simplify visitors ppx annotations

AltGr:
the small downside is that we lose a little bit of granularity by making
defining all the types at once (with `type ... and`) ; but one well-placed
annotation seems to be enough.

also discarded the `iter` visitor that was unused.

### DIFF
--- a/compiler/surface/ast.ml
+++ b/compiler/surface/ast.ml
@@ -31,49 +31,17 @@ open Catala_utils
 (** {1 Type definitions} *)
 
 type uident = (string[@opaque])
-[@@deriving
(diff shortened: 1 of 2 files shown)


AltGr:
Not completely sure about this since I am not very experienced with `visitors` ; but the weight of annotations was a bit overwhelming and hindered readability of the surface ast module. @denismerigoux what do you think ?

denismerigoux:
Wouhaou that's clever. I'm okay with this!

denismerigoux (approved):

PR #496, opened 2023-08-05
R backend for Catala

denismerigoux:
This R backend is based on the Python backend. It has been debugged only on small examples for should do for a proof of concept. 

The R code extracted from Catala cannot operate directly on R arrays for maximum efficiency. You have to map the functions over each element of your input array. This is not optimal as the point of R is array computations. Hence, this R backend should be tweaked to offer the possibility to directly compute over arrays.

I didn't do it in this PR because I just learnt R yesterday and am not yet familiar enough with the language. What is proposed here is a basic thing that computes the right thing :)


compiler/scalc/to_r.ml
```diff
@@ -0,0 +1,551 @@
...
+       (fun fmt info ->
+         Format.fprintf fmt "\"%s\""
+           (Re.replace sanitize_quotes ~f:(fun _ -> "\\\"") info)))
+    uids
+
+let avoid_keywords (s : string) : string =
+  if
+    match s with
+    (* list taken from
+       https://www.programiz.com/python-programming/keyword-list *)
```
AltGr, line 114: You probably want to find a similar list of R keywords :)

AltGr (approved):
Nice!

Variable renaming could be simplified (normally a single preliminar call to `Expr.rename_vars` with the appropriate options should be enough to avoid the need for all uses of `string_counter_map` etc.) — but that's something that was already needed in the Python backend so it's fine for the moment.

PR #497, opened 2023-08-15
Module support: handle structs, enums and scope calls across modules

AltGr:
Current state:

- [X] Works: modules compiled with `catala module --build foo.catala_en` then used
  through `catala interpret --use foo.catala_en bar.catala_en`
  (this is the use-case for running the interpreter on tests that will actually
  do a native-code computation)

- [X] Usage from native code ~doesn't work yet (but it's a matter of bugfixing)~ works as well (use `catala module --link` for now)

- [ ] Desambiguation doesn't yet work cross-modules yet (but no technical
  difficulty, just needs handling ; can be left for a later PR)

- [x] Code needs cleanup (debugs remaining here and there...)

- [X] ~On second thought storing the paths as part of the `program_ctx` isn't the
  best approach: I'll revert that part and add the paths to an expanded version
  of `Uid.t`. The information will be closer to where it's needed, it will avoid
  limitation in printers, and this just needs the current path to be passed along
  during name resolution, when initialising the `(Q)Uids`.~
  Done now, code is much simpler but one needs to remember to lookup the correct module context when querying scope or def info.

- [x] ~`web_api` and `french_law` friends are broken and temporarily disabled~ Now fixed, at least all tests and benches pass now

- no support yet for other backends (but that will be left for later PRs)


AltGr:
(and obviously, the git history needs a rebase ;) )

compiler/plugins/json_schema.ml
```diff
@@ -76,13 +76,14 @@ module To_json = struct
       (ctx : decl_ctx)
       (fmt : Format.formatter)
       (sname : StructName.t) =
+    let path, fields = StructName.Map.find sname ctx.ctx_structs in
     Format.pp_print_list
       ~pp_sep:(fun fmt () -> Format.fprintf fmt ",@\n")
       (fun fmt (field_name, field_type) ->
-        Format.fprintf fmt "@[<hov 2>\"%a\": {@\n%a@]@\n}"
+        Format.fprintf fmt "@[<hov 2>\"%a%a\": {@\n%a@]@\n}" Print.path path
```
denismerigoux, line 83: @EmileRolley this will impact you in the future. Now structs can be imported from external modules. @AltGr what is `path` here from structs from their own module? Does `Print.path` return `""`, or `"bla."` (with the `.`)?

compiler/scopelang/ast.mli
```diff
@@ -34,19 +34,22 @@ val locations_used : 'm expr -> LocationSet.t
...
-  scope_mark : 'm mark;
   scope_options : Desugared.Ast.catala_option Mark.pos list;
 }
 
 type 'm program = {
-  program_scopes : 'm scope_decl ScopeName.Map.t;
+  program_scopes : 'm scope_decl Mark.pos ScopeName.Map.t;
   program_topdefs : ('m expr * typ) TopdefName.Map.t;
+  program_modules : nil program ModuleName.Map.t;
+  (* Using [nil] here ensure that program interfaces don't contain any
```
denismerigoux, line 50: That is very neat!

compiler/scopelang/dependency.ml
```diff
@@ -147,6 +149,8 @@ let build_program_dep_graph (prgm : 'm Ast.program) : SDependencies.t =
             used_defs g)
         g scope.Ast.scope_decl_rules)
     prgm.program_scopes g
+(* TODO FIXME: Add submodules here, they may still need dependency resolution
```
denismerigoux, line 152: So normally I would say we work under the assumption that dependent modules have already been typechecked and the typechecking is OK. So no need to redo the checks here.

compiler/shared_ast/definitions.ml
```diff
@@ -22,6 +22,11 @@
 
 open Catala_utils
 module Runtime = Runtime_ocaml.Runtime
+module ModuleName = String
+(* TODO: should probably be turned into an Uid once we implement module import
```
denismerigoux, line 26: Yes!

compiler/shared_ast/scope.mli
```diff
@@ -1,4 +1,4 @@
-(* This file is part of the Catala compiler, a specification language for tax
+(* This file is part of the Catala compiler, a specification language for tax <
```
denismerigoux, line 1: Typo

compiler/shared_ast/typing.ml
```diff
@@ -353,6 +368,10 @@ module Env = struct
   let add_toplevel_var v typ t =
     { t with toplevel_vars = A.TopdefName.Map.add v typ t.toplevel_vars }
 
+  let add_module modname ~module_env t =
+    Message.emit_debug "ADD MODULE %a" A.ModuleName.format modname;
```
denismerigoux, line 372: `emit_info`

examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
```diff
@@ -35,13 +35,13 @@ You could have written : "condition",
 or "content"
 
 Error token:
-┌─⯈ examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en:11.21-11.26:
+┌─⯈ test_nsw_social_housie.catala_en:11.21-11.26:
```
denismerigoux, line 38: this should not have changed

examples/aides_logement/dune
```diff
@@ -1,17 +1,17 @@
...
-   (run
-    catala
-    api_web
-    -t
-    --plugin-dir=compiler/plugins
-    %{dep:aides_logement.catala_fr}))))
+;; (rule
+;;  (targets aides_logement.ml aides_logement_api_web.ml)
+;;  (deps
+;;   (source_tree ..)
```
denismerigoux, line 4: dead code?

examples/dune
```diff
@@ -0,0 +1 @@
+(dirs :standard \ aides_logement)
```
denismerigoux, line 1: Why?

french_law/dune
```diff
@@ -0,0 +1,3 @@
+(dirs catala_legifrance)
+
+;; ocaml js python
```
denismerigoux, line 3: ?

tests/test_scope/good/scope_call4.catala_en
```diff
@@ -51,13 +67,29 @@ f2 = λ (x: integer) →
...
+[DEBUG] Parsing scope_call4.catala_en
 [DEBUG] Name resolution...
 [DEBUG] Desugaring...
+[DEBUG] DESUGARED → prog scopes: HousingComputation
+RentComputation
+modules:
+[DEBUG] PROCESS_DEF f1@!
+[DEBUG] PROCESS_DEF f2@!
+[DEBUG] PROCESS_DEF result@!
+[DEBUG] PROCESS_DEF result@!
```
denismerigoux, line 79: These should not appear in the end

denismerigoux (approved):
Thanks @AltGr for another massive PR! It still needs some work though, but that'll be after your holidays :)

examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
AltGr: Ah that was done on purpose in edd6dce0d8183212488f3a27d0558fabd598c5d2 ;
my motivation was that it's annoying for the test results to fail depending on where you were running the test from. (e.g. `clerk runtest examples/.../test.catala_en` would work but `cd examples/.../ && clerk runtest test.catala_en` would fail)

compiler/dcalc/from_scopelang.ml
```diff
@@ -1131,7 +1277,7 @@ let translate_program (prgm : 'm Scopelang.Ast.program) : 'm Ast.program =
         ctx )
   in
   let items, ctx = translate_defs top_ctx defs_ordering in
-  {
-    code_items = Bindlib.unbox items;
-    decl_ctx = { decl_ctx with ctx_structs = ctx.structs };
-  }
+  (* WIP TODO FIXME HERE: the scopes in submodules are not translated here it
```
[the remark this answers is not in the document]
AltGr: haha this is a leftover it has been fixed since :grimacing: I need to fix these debugs and wip comments ^^

compiler/desugared/name_resolution.mli
```diff
@@ -82,11 +78,13 @@ type context = {
           between different enums *)
   scopes : scope_context ScopeName.Map.t;  (** For each scope, its context *)
   topdefs : TopdefName.t Ident.Map.t;  (** Global definitions *)
+  topdef_types : typ TopdefName.Map.t;
```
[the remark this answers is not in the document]
AltGr: Ah, I see the name is unclear: it's related to the def right above: types of the values defined at toplevel. Will add a comment.

compiler/dcalc/from_scopelang.ml
```diff
@@ -1021,56 +1080,132 @@ let translate_program (prgm : 'm Scopelang.Ast.program) : 'm Ast.program =
     Scopelang.Dependency.get_defs_ordering defs_dependencies
   in
   let decl_ctx = prgm.program_ctx in
+  Message.emit_debug "prog scopes: %a@ modules: %a"
```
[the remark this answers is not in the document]
AltGr: Hum yes that was meant for temporary debugging and I intended to just scrap it :grin:

compiler/dcalc/from_scopelang.ml
```diff
@@ -926,12 +970,26 @@ let translate_scope_decl
...
+    Var.make (Mark.remove (ScopeName.get_info scope_name) ^ "_in")
+  in
   let scope_input_struct_name = scope_sig.scope_sig_input_struct in
   let scope_return_struct_name = scope_sig.scope_sig_output_struct in
   let pos_sigma = Mark.get sigma_info in
+  let scope_mark =
+    (* Find a witness of a mark in the definitions *)
+    match sigma.scope_decl_rules with
+    | [] ->
+      (* Todo: are we sure this can't happen in normal code ? E.g. is calling a
```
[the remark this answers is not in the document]
AltGr: It's a bit more involved because of the GADT for annotations, we would need to take a type witness as parameter of the function. I'll see what I can do ^^

compiler/desugared/name_resolution.ml
```diff
@@ -235,10 +233,20 @@ let get_scope ctxt id =
...
     Message.raise_spanned_error (Mark.get id) "No scope named %s found"
       (Mark.remove id)
 
+let rec module_ctx ctxt path =
+  match path with
+  | [] -> ctxt
+  | (modname, mpos) :: path -> (
+    match ModuleName.Map.find_opt modname ctxt.modules with
+    | None ->
+      Message.raise_spanned_error mpos "Module %a not found" ModuleName.format
```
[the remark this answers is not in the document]
AltGr: Good point. Incidentally, I was wondering if it wouldn't be appropriate to have the various instances of `Uid.format` already define a specific color. That should work since we now discriminate between color or raw printing at the latest moment, once we know if the output is a terminal or not.

compiler/plugins/json_schema.ml
AltGr: Yes the output here will be `ModuleName.var_name`

examples/dune
AltGr: Ah this is temporary and will be removed from the finalised PR, see  0a98f6331aaf283838f78fcc8536ae29f2ba5c55

french_law/dune
AltGr: (same as above, this disables most of `french_law`)

examples/NSW_community_gaming/tests/test_nsw_social_housie.catala_en
denismerigoux: ok but the code doing this in `clerk` should have a comment as to why we `cd` into the inline test directory.

compiler/dcalc/from_scopelang.ml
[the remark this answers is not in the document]
denismerigoux: https://erratique.ch/software/logs

compiler/plugins/json_schema.ml
EmileRolley: Noted!

AltGr:
Should now be regression-free and ready for final review;
I should still do some cleanup on the branch git history, but without impact on the result.

AltGr:
Done cleaning up the git history a bit

compiler/lcalc/to_ocaml.ml
```diff
@@ -529,7 +530,7 @@ let format_ctx
       format_to_module_name (`Ename enum_name)
       (Format.pp_print_list
          ~pp_sep:(fun fmt () -> Format.fprintf fmt "@\n")
-         (fun _fmt (enum_cons, enum_cons_type) ->
+         (fun fmt (enum_cons, enum_cons_type) ->
```
denismerigoux, line 533: Well caught!

compiler/shared_ast/typing.ml
```diff
@@ -361,6 +368,26 @@ module Env = struct
         (A.ScopeName.Map.find scope_name t.scopes)
     in
     { t with scope_vars }
+
+  let rec dump ppf env =
+    let pp_sep = Format.pp_print_space in
+    Format.pp_open_vbox ppf 0;
+    (* Format.fprintf ppf "structs: @[<hov>%a@]@,"
```
denismerigoux, line 375: Dead code?

denismerigoux (approved):
OK to merge! While I'm thinking about it : the invariants checking phase that @adelaett made is only triggered by the `Invariants` CLI target, but shouldn't they be checked more often and at least in the CI ?

compiler/lcalc/to_ocaml.ml
```diff
@@ -245,11 +223,15 @@ let format_var (fmt : Format.formatter) (v : 'm Var.t) : unit =
   let lowercase_name = String.to_ascii lowercase_name in
   if
     List.mem lowercase_name ["handle_default"; "handle_default_opt"]
-    || String.begins_with_uppercase (Bindlib.name_of v)
+    (* O_O *)
```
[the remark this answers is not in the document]
AltGr: Hm I don't exactly remember the mood I was in when I wrote this :rofl: 
But certainly there should be a better way of special-casing without matching the final ident strings.

compiler/shared_ast/typing.ml
AltGr: Well technically this function is dead code, it shouldn't be used outside of debugging sessions — but pretty convenient nevertheless and we shouldn't need to rewrite it every time. The same applies to the commented code: while you generally wouldn't want all the typedecls listed here, I thought that having this available to be uncommented here could be helpful in debugging sessions where you might actually have use for it.

But well, since it's normally unused we may indeed prefer to uncomment to make sure the extended function continues to typecheck...

AltGr:
> the invariants checking phase that @adelaett made is only triggered by the `Invariants` CLI target, but shouldn't they be checked more often and at least in the CI ?

It may even make sense to enable it in tests by default, when applicable :thinking:

compiler/driver.ml
```diff
@@ -42,30 +42,35 @@ let get_lang options file =
...
+      { prg with program_modules = load_module_interfaces options link_modules }
+    in
+    prg, language
 
   let desugared options ~link_modules :
       Desugared.Ast.program * Desugared.Name_resolution.context =
-    let prg, _ = surface options in
-    let prg = load_module_interfaces prg options link_modules in
+    let prg, _ = surface options ~link_modules in
+    Message.emit_debug "- DESUGARED -";
```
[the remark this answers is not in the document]
denismerigoux: Okay, looks good to me. Yep color would be perfect :)

compiler/desugared/linting.ml
```diff
@@ -136,7 +136,8 @@ let detect_unused_struct_fields (p : program) : unit =
   in
   StructName.Map.iter
     (fun s_name fields ->
-      if
+      if StructName.path s_name <> [] then ()
```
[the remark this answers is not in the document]
AltGr: In time we may also want to fine-tune the warnings when compiling modules: in some cases it makes sense that stuff is unused here.

PR #499, opened 2023-08-17
Improvements to the Polish translation

arcz:
Some improvements to the Polish translation:
- Added Polish diacritics to the rest of the keywords and tests
- Fixed some untranslated keywords such as `MS_DATA` and `MS_XOR`.

Things I'm not sure about:
- `MS_MONEY_OP_SUFFIX` what is this? It shouldn't be $ for sure but I'm not confident it should be PLN.
- Should there be a `MR_*` version for every keyword with diacritics?


compiler/surface/lexer_pl.cppo.ml
```diff
@@ -20,18 +20,18 @@
 
 #define MS_SCOPE "zakres"
 #define MS_CONSEQUENCE "konsekwencja"
-#define MS_DATA "data"
-#define MS_DEPENDS "zalezy od"
-#define MR_DEPENDS "zalezy", space_plus, "od"
+#define MS_DATA "dane"
+#define MS_DEPENDS "zależy od"
```
denismerigoux, line 24: So each time you have a `MS_XXXX` that contains an non-ASCII character you should define just below a `MR_XXXX` line with a format like `"pieni", 0x0105, "dze"` where you concat ASCII strings and UFT8 hex encoding of the non-ASCII characters. You can use sites like https://www.compart.com/en/unicode/U+0105 to figure out the encoding of the letter you're putting. This is annoying but required for the lexer, we tried to find a better way but this is the least annoying solution.

compiler/surface/lexer_pl.cppo.ml
```diff
@@ -103,7 +102,7 @@
 
 (* Specific delimiters *)
 
-#define MS_MONEY_OP_SUFFIX "$"
+#define MS_MONEY_OP_SUFFIX "PLN"
```
denismerigoux, line 105: Yep that looks fine!

denismerigoux (changes requested):
Thank you ! The Polish lexer and the examples were done by @wozniakpl but we didn't reach out to him every time we changed the language or added features so some drift has appeared. I hope this will make Catala more usable for Polish speakers.

denismerigoux (approved):

PR #505, opened 2023-09-09
add missing french_law/python/src/runtime.py to .gitignore

zapashcanon:
Hi,

I had this file appearing as a new one when doing `git status` after a `dune build @all`. I guess it's generated and was missing in the `.gitignore` ?


AltGr:
Ah indeed! It used to be versionned but we have been doing some cleanup in 71e3ca21c09c345e94b48873dc52543c368edf79 and apparently this was forgotten. Thanks !

PR #507, opened 2023-09-13
replace `let _` by `let ()` or add type annotation

zapashcanon:
Hi,

Some nitpicking (this is actually sometimes useful as it avoids getting hard-to-debug issues when changing the arity of a function that has one of its call site ignored).


AltGr (approved):
Thanks!

PR #512, opened 2023-09-15
Rehaul the Clerk build system

AltGr:
The architecture of Clerk has been extended and improved, and now includes
support for modules and dependencies (including building the modules to OCaml
objects as required for either interpretation or linking).

A few notes:

- all artifacts are now written to _build/

- the architecture should make it easy to add rules for python, etc.
  Each catala file has a bunch of rules declared, that basically call upon the
  different Catala backends. In addition:
  * if a module name is defined, the corresponding targets
  * phony rules called `name@path/to/file` for tests, etc.

- the relevant code from the `catala module` plugin has been integrated into
  Clerk so the plugin can probably be recycled

- [ ] output tests (not inline) are supposed to be supported but I haven't yet re-added
  one to verify that they actually work

- to keep things simple, at the moment, Clerk will scan all files (including
  contents) at every startup, generate a big Ninja file with all of their rules
  in it (3000 lines for the catala repo which has lots of tests).
  Then ninja is run on that file.
  Clearly this is sub-optimal ; but I couldn't yet notice any delay. This should be
  optimised at some point but that can be left to future PRs.
  The easiest optimisation is to re-run ninja directly without rebuilding the
  file, assuming that the dependency graph hasn't changed, which should cover
  most incremental builds.
  It would probably be wiser to wait to see where the bottlenecks are when they
  happen before trying to be too clever.

- the code uses OCaml `Seq.t` throughout rather than lists. The original idea
  was to be able to pipeline the generation of the Ninja file as the files were
  being scanned, unfortunately at the moment we are required to scan for all
  module definitions to begin with, so the purpose is kind of defeated.
  It may still be useful for future optimisations though, and there was little
  value to rewriting to use lists so I kept it this way.

- [ ] having modules depending on other modules is expected to work, but that
  needs more tests.

- error messages will also need to be improved over time. We might need
  redundant checks in Clerk to avoid less clear messages from Ninja :/


AltGr:
Also `clerk runtest` should be updated to use the new lexer instead of its old regexp-based parsing.

denismerigoux:
I'll try and find the time to review this :)

denismerigoux:
You're using `biniou` now ? The CI build fails because it's missing.

compiler/driver.ml
```diff
@@ -46,8 +46,10 @@ let load_module_interfaces options link_modules =
   List.map
     (fun f ->
       let lang = get_lang options (FileName f) in
-      let modname = modname_of_file f in
-      let intf = Surface.Parser_driver.load_interface (FileName f) lang in
+      let modname, intf =
+        Surface.Parser_driver.load_interface (FileName f) lang
+      in
+      (* maybe warn here if [modname_of_file f <> modname] ? *)
```
denismerigoux, line 52: Yes there should be a warning.

compiler/literate/html.ml
```diff
@@ -200,6 +200,7 @@ let rec law_structure_to_html
          (law_structure_to_html language print_only_law complete_headings))
       children fmt_details_close ()
   | A.LawInclude _ -> ()
+  | A.ModuleDef _ | A.ModuleUse _ -> () (* TODO: show somehow ?*)
```
denismerigoux, line 203: Yes we should show it. I would be in favor of collecting all module def and use and show a summary of them at the very beginning, after the list of files weaved for the current module.

compiler/surface/lexer.cppo.ml
```diff
@@ -349,6 +349,9 @@ let digit = [%sedlex.regexp? '0' .. '9']
 (** Regexp matching at least one space. *)
 let space_plus = [%sedlex.regexp? Plus white_space]
 
+(** characters that can be present in idents (excluding first char) *)
+let idchar = [%sedlex.regexp? uppercase | lowercase | digit | '_' | '\'']
```
denismerigoux, line 353: Nice

compiler/surface/lexer_common.mli
```diff
@@ -45,6 +45,16 @@ val calc_precedence : string -> int
 val get_law_heading : Sedlexing.lexbuf -> Tokens.token
 (** Gets the [LAW_HEADING] token from the current [lexbuf] *)
 
+(** Simplified tokens for dependency extraction *)
+type line_token =
```
denismerigoux, line 49: I would have called it `dependency_line_token`

compiler/surface/lexer_common.mli
```diff
@@ -62,4 +72,7 @@ module type LocalisedLexer = sig
   val lexer : Sedlexing.lexbuf -> Tokens.token
   (** Entry point of the lexer, distributes to {!val:lex_code} or {!val:lex_law}
       depending of the current {!val:Surface.Lexer_common.context}. *)
+
+  val lex_line : Sedlexing.lexbuf -> (string * line_token) option
```
denismerigoux, line 76: `lex_dependency_line`?

compiler/surface/lexer_en.cppo.ml
```diff
@@ -111,6 +111,9 @@
 (* Directives *)
 
 #define MR_LAW_INCLUDE "Include"
+#define MR_MODULE_DEF "Module"
```
denismerigoux, line 114: Could you add these to all the syntax highlighters ? I know it's a pain but it's useful...

compiler/surface/lexer_pl.cppo.ml
```diff
@@ -136,6 +136,9 @@
 (* Directives *)
 
 #define MR_LAW_INCLUDE "Include"
+#define MR_MODULE_DEF "Module"
```
denismerigoux, line 139: @arcz we're adding modules to Catala, do you want to chime in with correct Polish terms for these?

tests/test_modules/good/mod_use.catala_en
```diff
@@ -20,10 +22,8 @@ scope T2:
   assertion o4 = 5.0
 ```
 
-The following test is disabled at the moment because it requires prior compilation of the module. This will be fixed with the new Clerk version that handles dependencies transparently.
-
-```disabled catala-test-inline
-$ catala interpret -s T2 --disable_warnings --use mod_def.catala_en
+```catala-test-inline
+$ catala interpret -s T2
```
denismerigoux, line 26: Yeeaaah!

denismerigoux (approved):
Thanks @altgr, this is getting really good!

compiler/catala_utils/file.ml
```diff
@@ -118,3 +120,39 @@ let check_directory d =
 let ( / ) = Filename.concat
 let dirname = Filename.dirname
 let ( /../ ) a b = dirname a / b
+let ( -.- ) file ext = Filename.chop_extension file ^ "." ^ ext
```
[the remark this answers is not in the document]
AltGr: Normally I avoid relying on those too much, but here I think they really improve legibility ; and `/../` is pretty self-explanatory ;)

compiler/surface/lexer_en.cppo.ml
AltGr: It's in the tree-sitter syntax already :upside_down_face: 

Joke aside, I think we'll want to add a specific declaration for externally-implemented modules (maybe just `> Module Foo external` ?) so might be a good idea to settle on this and do all of them in one go

build_system/clerk_driver.ml
```diff
@@ -157,366 +176,507 @@ type expected_output_descr = {
...
-    ~(rule_name : string)
-    (catala_exe_opts : string)
-    (rules : Rule.t Nj.RuleMap.t) : Rule.t Nj.RuleMap.t =
-  let reset_common_cmd_exprs =
-    Nj.Expr.
+(** Some functions that poll the surrounding systems (think [./configure]) *)
+module Poll = struct
+
+  (** Scans for a parent directory being the root of the Catala source repo *)
+  let catala_project_root: File.t option Lazy.t =
```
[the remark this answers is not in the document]
AltGr: 1. the fine-tuning is disabled if not running from the catala tree ; the issue I was trying to solve is that, in normal operation, all libraries should be installed inside `<prefix>/lib/catala` ; but when running the catala tests themselves it would be pretty inconvenient to have to install the full project every time, and you would always have the risk of inconsistency between the installed artifacts and what you actually wanted to test.
  Another possible workaround, more error-prone but simpler, would be to simply enforce that through the environment when running the catala tests.
2. well we don't want all these to unconditionnally run at program initialisation (imagine they fail and it breaks even `--help` ? You wouldn't get proper handling of the debug flags, etc.) — and we also certainly don't want to run them multiple times if queried multiple times. So that's what `lazy` is for :D

build_system/clerk_driver.ml
```diff
@@ -157,366 +176,507 @@ type expected_output_descr = {
...
+      | None -> "catala" (* Dynamically resolved from PATH *)
+  )
+
+  let build_dir: File.t Lazy.t = lazy (
+    File.(Sys.getcwd () / "_build")
+  )
+  (* TODO: probably detect the project root and put this there instead *)
+
+  (** Locates the main [lib] directory containing the OCaml libs *)
+  let ocaml_libdir: File.t Lazy.t =
```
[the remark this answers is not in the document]
AltGr: No, it's for getting the OCaml linking flags when building catala programs from the `ml` outputs, we manually link to the Catala runtime and its dependencies.

build_system/clerk_driver.ml
```diff
@@ -157,366 +176,507 @@ type expected_output_descr = {
...
+  let ocaml_link_flags: string list Lazy.t = lazy (
+    let link_libs =
       [
-        Var.catala_cmd;
-        Var.tested_file;
-        Lit redirect;
-        Var.expected_output;
-        Lit "2>&1";
-        Lit "|| true";
+        "biniou";
```
AltGr, line 357: I have a fix pending, `yojson.1.7` depended on `biniou` but it's no longer the case with `2.0` so I am making the version constraint stricter and removing this dep.

build_system/ninja_utils.ml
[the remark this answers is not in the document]
AltGr: Yes, that's planned ^^

compiler/surface/lexer_common.mli
AltGr: I didn't change the name in the end because by itself it's not so strictly related with dependencies; it just recognises some kinds of lines like inclusion and module directives, but also the ones delimiting code blocks (it's the ones for test blocks that we are interested in for clerk)

AltGr:
Huge update (and rebase) of this PR, @denismerigoux you might want to have another look.

compiler/surface/lexer_en.cppo.ml
AltGr: Hm I had a look but I can't figure if there is any support for directives at the moment... I didn't find any reference to the existing `Include` for example.

denismerigoux: `> module Foo external` is a good idea!

compiler/shared_ast/operator.ml
```diff
@@ -75,7 +75,7 @@ let name : type a. a t -> string = function
   | Div_int_int -> "o_div_int_int"
   | Div_rat_rat -> "o_div_rat_rat"
   | Div_mon_mon -> "o_div_mon_mon"
-  | Div_mon_rat -> "o_div_mon_mon"
+  | Div_mon_rat -> "o_div_mon_rat"
```
denismerigoux, line 78: Well caught!!!

denismerigoux (approved):
It's beeter than the last time :) You can merge for me!

PR #520, opened 2023-09-21
Fix linebreaking in LaTeX Catala code

denismerigoux:
The problem is coming from the fact that the `breaklines` option of `fancyvrb` is not compatible with the `PY` macro output by pygmentize. 

See 7.3.2 of https://ctan.tetaneutral.net/macros/latex/contrib/fvextra/fvextra.pdf

### DIFF
--- a/compiler/literate/latex.ml
+++ b/compiler/literate/latex.ml
@@ -58,7 +58,7 @@ let wrap_latex
 \usepackage{fontspec}
 \usepackage[hidelinks]{hyperref}
 %s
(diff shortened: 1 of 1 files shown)


denismerigoux:
Capture d’écran du 2023-09-21 17-56-19

AltGr (approved):
I'll trust you on this dark magic ;)

AltGr:
Ah I see you require one more LaTeX package. We'll see how the CI copes...

PR #527, opened 2023-10-31
Move examples and french_law to their own repositories

AltGr:


AltGr:
This part should be OK, but the examples and french_law packages still need to be adjusted and tested ; they should also be made part of the CI

AltGr:
Update: this now works with the current versions of
- https://github.com/CatalaLang/catala-examples and
- https://github.com/CatalaLang/french-law

and should build everything that was built previously.
Some work remains though:

- the CI should be updated. In particular, we want both:
  * a CI on the two new repos that rely on an up-to-date catala installation
  * a CI on catala that fetches the examples and french-law to build them (maybe based on their `dev` branches)

  we will also want to get the generated artifacts published as they are now

- ensure we have a clean way to generate the website assets

- publish all the required runtimes where needed
  * [x] python runtime on [pypi](https://pypi.org/project/catala-runtime/0.9.0/)
  * [ ] rescript runtime on [npm](https://www.npmjs.com/package/@catala-lang/rescript-catala) (outdated)
  * [ ] R runtime on CRAN (???)
  * OCaml and jsoo runtimes are already installed with Catala on opam so that should be enough

AltGr:
This is still marked draft (due to the remaining work listed above) but actually works on its own, so we have two choices here:
- wait for everything to be ready (in particular, avoiding any regressions in CI)
- merge now, accept a period of mild instability regarding the CI and examples and fix as we go

The point in favor of the second option would be that any patches to the examples start being done on the `catala-examples` repo, avoiding painful porting. Also fixing CI that way might be easier. But the main concern here should be the imminent release !

@denismerigoux what do you think ?

denismerigoux:
I'm okay with the second option but on the condition that you block ~ a week of work on this exclusively once you merge to fix all the bugs in a short period of time :)

PR #528, opened 2023-11-07
Internally type default terms in the compiler

AltGr:
This reinforces the internal typing discipline by treating the output of default
terms (which could be ∅) differently from raw terms.

Most of the changes are to adjust the actual compiler passes to adhere to the
new type discipline. In particular, as of `scopelang` the input and output types
of a given scope variable may be different (a reentrant variable would require
`'a default` input, but be accessed as `'a`).

In addition to adding more guarantees to the correctness of the compiler
passes (we now have strong assurance that `error_empty` calls are in the right
places, since they do type translation), I expect this to be extremely helpful
in simplifying the with/without exceptions backends.

Note: the git history should be cleaned up before merging.

### DIFF
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -293,11 +293,13 @@ let rec translate_expr (ctx : 'm ctx) (e : 'm Scopelang.Ast.expr) :
               ]
               "Definition of input variable '%a' missing in this scope call"
               ScopeVar.format var_name
-          | None, Some _ ->
-            Message.raise_multispanned_error
+          | None, Some e ->
+            Message.raise_multispanned_error_full
+              ~suggestion:(List.map (fun v -> Mark.remove (ScopeVar.get_info v))
+                             (ScopeVar.Map.keys sc_sig.scope_sig_in_fields))
               [
-                None, pos;
-                ( Some "Declaration of scope '%a'",
+                None, Expr.pos e;
+                ( Some (fun ppf -> Format.fprintf ppf "Declaration of scope %a" ScopeName.format scope),
                   Mark.get (ScopeName.get_info scope) );
               ]
               "Unknown input variable '%a' in scope call of '%a'"
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -117,9 +117,12 @@ let merge_defaults
           (Expr.with_ty m_callee
              (Mark.add (Expr.mark_pos m_callee) (TLit TBool)))
       in
-      let d = Expr.edefault [caller] ltrue (Expr.rebox body) m_body in
+      let d =
+        Expr.edefault [caller] ltrue (Expr.rebox body)
+          (Expr.map_ty (fun (_, pos as t) -> TDefault t, pos) m_body)
+      in
       Expr.make_abs vars
-        (Expr.eerroronempty d m_body)
+        (Expr.make_erroronempty d)
         tys (Expr.mark_pos m_callee)
     | _ -> assert false
     (* should not happen because there should always be a lambda at the
@@ -138,7 +141,8 @@ let merge_defaults
         Expr.elit (LBool true)
           (Expr.with_ty m (Mark.add (Expr.mark_pos m) (TLit TBool)))
       in
-      Expr.eerroronempty (Expr.edefault [caller] ltrue callee m) m
+      Expr.make_erroronempty (Expr.edefault [caller] ltrue callee
+          (Expr.map_ty (fun (_, pos as t) -> TDefault t, pos) m))
     in
     body
 
@@ -216,7 +220,7 @@ let input_var_typ typ io_in =
 let thunk_scope_arg var_ctx e =
   match var_ctx.scope_input_io, var_ctx.scope_input_thunked with
   | (Runtime.NoInput, _), _ -> invalid_arg "thunk_scope_arg"
-  | (Runtime.OnlyInput, _), false -> Expr.eerroronempty e (Mark.get e)
+  | (Runtime.OnlyInput, _), false -> Expr.make_erroronempty e
   | (Runtime.Reentrant, _), false -> e
   | (Runtime.Reentrant, pos), true ->
     Expr.make_abs [| Var.make "_" |] e [TLit TUnit, pos] pos
@@ -626,8 +630,7 @@ let translate_rule
       | OnlyInput -> failwith "should not happen"
       (* scopelang should not contain any definitions of input only variables *)
       | Reentrant -> merge_defaults ~is_func a_expr new_e
-      | NoInput ->
-        if is_func then new_e else Expr.eerroronempty new_e (pos_mark_as a_name)
+      | NoInput -> new_e
     in
     let merged_expr =
       tag_with_log_entry merged_expr
@@ -682,11 +685,11 @@ let translate_rule
     let thunked_or_nonempty_new_e =
       match a_io.Desugared.Ast.io_input with
       | Runtime.NoInput, _ -> assert false
-      | Runtime.OnlyInput, _ -> Expr.eerroronempty new_e (Mark.get new_e)
+      | Runtime.OnlyInput, _ -> Expr.make_erroronempty (new_e)
       | Runtime.Reentrant, pos -> (
         match Mark.remove tau with
-        | TArrow _ -> new_e
-        | _ -> Expr.thunk_term new_e (Expr.with_pos pos (Mark.get new_e)))
+        | TArrow _ -> (new_e)
+        | _ -> Expr.thunk_term (new_e) (Expr.with_pos pos (Mark.get new_e)))
     in
     ( (fun next ->
         Bindlib.box_apply2
@@ -890,7 +893,7 @@ let translate_rule
                      defined, we add an check "ErrorOnEmpty" here. *)
                   Mark.add
                     (Expr.map_ty (fun _ -> scope_let_typ) (Mark.get e))
-                    (EAssert (Mark.copy e (EErrorOnEmpty new_e)));
+                    (EAssert (Expr.unbox (Expr.make_erroronempty (Expr.box new_e))));
                 scope_let_kind = Assertion;
               })
           (Bindlib.bind_var (Var.make "_") next)
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -739,10 +739,22 @@ let translate_rule
           | _ -> true)
         all_subscope_vars
     in
+    let called_scope_input_struct = subscope_sig.scope_sig_input_struct in
+    let called_scope_return_struct = subscope_sig.scope_sig_output_struct in
     let all_subscope_output_vars =
-      List.filter
+      List.filter_map
         (fun var_ctx ->
-          Mark.remove var_ctx.scope_var_io.Desugared.Ast.io_output)
+           if Mark.remove var_ctx.scope_var_io.Desugared.Ast.io_output then
+             (* Retrieve the actual expected output type through the scope output struct definition *)
+             let str =
+               StructName.Map.find called_scope_return_struct ctx.decl_ctx.ctx_structs
+             in
+             let fld =
+               ScopeVar.Map.find var_ctx.scope_var_name scope_sig_decl.out_struct_fields
+             in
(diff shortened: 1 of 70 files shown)


AltGr:
Now passes all tests _and_ examples

compiler/dcalc/from_scopelang.ml
```diff
@@ -216,7 +220,7 @@ let input_var_typ typ io_in =
 let thunk_scope_arg var_ctx e =
   match var_ctx.scope_input_io, var_ctx.scope_input_thunked with
   | (Runtime.NoInput, _), _ -> invalid_arg "thunk_scope_arg"
-  | (Runtime.OnlyInput, _), false -> Expr.eerroronempty e (Mark.get e)
+  | (Runtime.OnlyInput, _), false -> e
```
denismerigoux, line 223: OK well spotted, indeed for true inputs of scopes we don't need error on empty because the input has been defined earlier and there's already an error on empty there. Is that the right reason?

compiler/dcalc/from_scopelang.ml
```diff
@@ -626,8 +638,7 @@ let translate_rule
       | OnlyInput -> failwith "should not happen"
       (* scopelang should not contain any definitions of input only variables *)
       | Reentrant -> merge_defaults ~is_func a_expr new_e
-      | NoInput ->
-        if is_func then new_e else Expr.eerroronempty new_e (pos_mark_as a_name)
+      | NoInput -> new_e
```
denismerigoux, line 641: I guess you're also sure that the error on empty is not needed here.

compiler/dcalc/from_scopelang.ml
```diff
@@ -682,7 +693,7 @@ let translate_rule
     let thunked_or_nonempty_new_e =
       match a_io.Desugared.Ast.io_input with
       | Runtime.NoInput, _ -> assert false
-      | Runtime.OnlyInput, _ -> Expr.eerroronempty new_e (Mark.get new_e)
+      | Runtime.OnlyInput, _ -> new_e
```
denismerigoux, line 696: Wouahou I was putting a lot of errors on empty where it was not necessary...

compiler/dcalc/from_scopelang.ml
```diff
@@ -890,7 +914,7 @@ let translate_rule
                      defined, we add an check "ErrorOnEmpty" here. *)
```
denismerigoux, line 914: If you're sure the error on empty is not needed here then the comment above should be deleted as well.

compiler/desugared/ast.ml
```diff
@@ -174,7 +174,10 @@ let empty_rule
     (parameters : (Uid.MarkedString.info * typ) list Mark.pos option) : rule =
   {
     rule_just = Expr.box (ELit (LBool false), Untyped { pos });
-    rule_cons = Expr.box (EEmptyError, Untyped { pos });
+    rule_cons =
+      Expr.eerroronempty
```
denismerigoux, line 178: I thought you could avoid `ErrorOnEmpty EmptyError` no?

compiler/shared_ast/definitions.ml
```diff
@@ -549,6 +550,9 @@ and ('a, 'b, 'm) base_gexpr =
       cons : ('a, 'm) gexpr;
     }
       -> ('a, < defaultTerms : yes ; .. >, 'm) base_gexpr
+  | EPureDefault :
```
denismerigoux, line 553: A comment should explain what this is here.

compiler/shared_ast/expr.mli
```diff
@@ -337,20 +354,10 @@ val make_default :
   ('a, 'm) boxed_gexpr list ->
   ('a, 'm) boxed_gexpr ->
   ('a, 'm) boxed_gexpr ->
-  'm mark ->
+  (* 'm mark -> *)
```
denismerigoux, line 357: Comment?Dead code?

compiler/shared_ast/typing.ml
```diff
@@ -216,26 +224,44 @@ let handle_type_error ctx (A.AnyExpr e) t1 t2 =
...
+      [
+        ( (fun ppf ->
+            Format.fprintf ppf "@[<hv 2>@[<hov>%a@ %a@]:" Format.pp_print_text
+              "This expression has type" (format_typ ctx) t1;
+            if Cli.globals.debug then Format.fprintf ppf "@ %a@]" Expr.format e
+            else Format.pp_close_box ppf ()),
+          e_pos );
+        ( (fun ppf ->
+            Format.fprintf ppf
+              "@[<hov>Expected@ type@ %a@ coming@ from@ expression:@]"
```
denismerigoux, line 241: Thanks for the improvement!

compiler/shared_ast/typing.ml
```diff
@@ -267,7 +293,8 @@ let polymorphic_op_type (op : Operator.polymorphic A.operator Mark.pos) :
   let it = lazy (UnionFind.make (TLit TInt, pos)) in
   let cet = lazy (UnionFind.make (TClosureEnv, pos)) in
   let array a = lazy (UnionFind.make (TArray (Lazy.force a), pos)) in
-  let option a = lazy (UnionFind.make (TOption (Lazy.force a), pos)) in
+  (* let option a = lazy (UnionFind.make (TOption (Lazy.force a), pos)) in *)
```
denismerigoux, line 296: Dead code?

compiler/shared_ast/typing.ml
```diff
@@ -282,10 +309,18 @@ let polymorphic_op_type (op : Operator.polymorphic A.operator Mark.pos) :
...
     | HandleDefaultOpt ->
-      [array (option any); [ut] @-> option bt; [ut] @-> option any]
-      @-> option any
+      (* FIXME: when translating with avoid_exceptions, defaults morally become
+         options everywhere. However, this is not reflected in the environment
+         at the moment, so this operator may have to mix-and-match options and
+         default terms ; since the typing of default is expected to guide a
+         simplification of the avoid-exceptions backend anyway, this is left
+         untyped at the moment *)
+      any
```
denismerigoux, line 321: Wait if there's this mix and match at the moment, how are you able to type output OCaml programs compiled without exceptions?

compiler/verification/conditions.ml
```diff
@@ -142,12 +142,7 @@ let match_and_ignore_outer_reentrant_default (ctx : ctx) (e : typed expr) :
         (Print.expr ()) e)
   | EErrorOnEmpty d ->
     d (* input subscope variables and non-input scope variable *)
-  | _ ->
-    Message.raise_spanned_error (Expr.pos e)
-      "Internal error: this expression does not have the structure expected by \
-       the VC generator:\n\
-       %a"
-      (Print.expr ()) e
+  | _ -> e
```
denismerigoux, line 145: Not sure you want to make that completely silent, maybe a debug message only?

denismerigoux: This PR should not break the proof mode, this should be updated to follow the new calling conventions.

tests/test_default/bad/empty.catala_en
```diff
@@ -19,7 +19,8 @@ $ catala Interpret -s A
   │          ‾
   └─ Article
 [ERROR]
-This variable evaluated to an empty term (no rule that defined it applied in this situation)
+This variable evaluated to an empty term (no rule that defined it applied in this situation): 
+error_empty ⟨ ⟨false ⊢ ⟨error_empty ∅⟩⟩ | false ⊢ ∅ ⟩
```
denismerigoux, line 23: Not sure you want to print a term here without the `--debug` option.

tests/test_func/good/closure_conversion.catala_en
```diff
@@ -19,33 +19,41 @@ type S_in = { x_in: eoption bool; }
 
 type S = { z: eoption integer; }
  
-let topval closure_f : (closure_env, integer) → eoption integer =
+let topval closure_f : (closure_env, integer) → eoption any =
```
denismerigoux, line 22: The Lcalc without exceptions should have typechecking able to infer everything, this is a regression (because of the mix-and-match of `HandleDefaultOpt` I guess). We should investigate to solve this in this PR.

tests/test_name_resolution/good/toplevel_defs.catala_en
```diff
@@ -112,84 +112,129 @@ let glob2_10 = A {"y": glob1_2 >= 30., "z": 123. * 17.}
...
+    decl temp_a_20 : any;
+    let temp_a_20 (__21 : unit) =
+      return false;
     decl temp_a_14 : any;
     let temp_a_14 (__15 : unit) =
-      return true;
-    temp_a_13 = handle_default [] temp_a_14 temp_a_16
+      decl temp_a_18 : any;
+      let temp_a_18 (__19 : unit) =
+        return glob3_3 ¤44.00 + 100.;
```
denismerigoux, line 125: This should use the localized printer!

denismerigoux (commented):
Thanks @AltGr ! There's a couple of things that should be fixed before merging though I think.

compiler/dcalc/from_scopelang.ml
AltGr: I think that's right, yes :)

compiler/dcalc/from_scopelang.ml
AltGr: Yes, well it was very hard to ensure they were everything where needed without the extra typing :)

compiler/desugared/ast.ml
AltGr: in the normal case yes, because we expect a default term to be of type `< 'a >`; 
but when building the exception tree on the lhs of a default term, we may need rules that return a non-default term.

I'll definitely clean this up (removing these empty exceptions that are just leftovers from the construction of the tree seems like the best way) to avoid this `error_empty` appearing at a disturbing place.

compiler/shared_ast/typing.ml
AltGr: the mix-and-match only concerns the inputs of the operator that are either of type option, or default. But @adelaett will look into it, it should definitely not stay like that

tests/test_name_resolution/good/toplevel_defs.catala_en
AltGr: No, this is a debug intermediate representation output, the localised printer is intended for end-user directed values (like results of computations)

tests/test_default/bad/empty.catala_en
AltGr: Indeed, this looks like a debugging leftover

compiler/verification/conditions.ml
AltGr: I'll double-check, but I think this is the correct output of this function: if it is wrapped inside `ErrorOnEmpty`, logging calls, etc, we strip them but this way it also works with non-wrapped terms (which may happen with optimisations on, see https://github.com/CatalaLang/catala/pull/528/files/5f5788370f664da0df9a36ea417cd87fefc3641d#diff-1b5c91edc326615f7262dc492bee44c128b18d1f0ae943b3127079c22a98cae7R218-R222

Hm, on second thought it might be `Expr.skip_wrappers e` ?

adelaett:
Thanks to this update, I've implemented the without exceptions pass. Most tests pass, but a few ones fail. The missing ones are:
* the ones related to the closure conversion pass. I'm not sure how to fix it
* the ones related to modules. I'm not sure how to fix it.

I've not yet tested the ocaml/python extraction but only the interpreter. The generated code is really way easier to read.

denismerigoux:
@adelaett your nix fix didn't work on master : https://github.com/CatalaLang/catala/actions/runs/6893184952/job/18752160534

denismerigoux:
@AltGr can we merge this quickly so that we don't accumulate too much unmerged changes for your subsequent PRs?

tests/test_date/good/rounding_option_fr.catala_fr
```diff
@@ -27,3 +27,8 @@ $ catala Interpret -s Test
 [RESULT] Computation successful! Results:
 [RESULT] r = vrai
 ```
+```catala-test-inline
+$ catala Interpret_Lcalc -s Test --avoid_exceptions --optimize
+[RESULT] Computation successful! Results:
+[RESULT] r = vrai
+```
```
AltGr, line 34: Please ensure to have newlines at the end of files. I think I fixed it, but I remember in particular that ``` wouldn't correctly parse as an end of test if not followed by a newline.

PR #535, opened 2023-11-30
Clerk and testing improvements, rework resolution of module elements

AltGr:
Clerk improvements
------------------

- Add a `-I` option that allows defined modules to be available from other
 directories

- Add reporting of the number of successful / failed tests

- Locate the project root, and always run the commands from there

Module elements
---------------

This changes the `decl_ctx` to be toplevel only, with flattened references
to uids for most elements. The module hierarchy, which is still useful in a
few places, is kept separately.

Module names are also changed to UIDs early on, and support for module
aliases has been added (needs testing).

This resolves some issues with lookup, and should be much more robust, as
well as more convenient for most lookups.

The `decl_ctx` was also extended for string ident lookups, which avoids
having to keep the desugared resolution structure available throughout the
compilation chain.


compiler/driver.ml
```diff
@@ -29,66 +29,89 @@ let modname_of_file f =
...
         ms
   in
-  let load_file f =
-    let (mname, intf), using =
-      Surface.Parser_driver.load_interface (Cli.FileName f)
-    in
-    (ModuleName.of_string mname, intf), using
-  in
-  let rec aux req_chain acc modules =
+  (* modulename * program * (id -> modulename) *)
```
denismerigoux, line 65: you can turn that into a proper type annotation no?

compiler/driver.ml
```diff
@@ -274,43 +283,32 @@ module Passes = struct
...
   open Cmdliner
 
-  let get_scope_uid (ctxt : Desugared.Name_resolution.context) (scope : string)
-      =
-    match Ident.Map.find_opt scope ctxt.typedefs with
-    | Some (Desugared.Name_resolution.TScope (uid, _)) -> uid
-    | _ ->
+  let get_scope_uid (ctx : decl_ctx) (scope : string) : ScopeName.t =
+    if String.contains scope '.' then
+      Message.raise_error "Only references to the top-level module are allowed";
```
denismerigoux, line 300: "... are allowed in scope reference %a ..."

compiler/scopelang/from_desugared.ml
```diff
@@ -789,87 +760,74 @@ let translate_program
...
+                reentrant_vars =
+                  Option.fold reentrant
+                    ~some:(fun ty ->
+                      ScopeVar.Map.add scope_var ty ctx.reentrant_vars)
+                    ~none:ctx.reentrant_vars;
+              })
+            scdef.D.scope_vars ctx)
+        modul.D.module_scopes ctx
+    in
+    (* Todo: since we rename all scope vars at this point, it would be better to
```
denismerigoux, line 823: I agree.

compiler/shared_ast/definitions.ml
```diff
@@ -666,14 +673,20 @@ type scope_info = {
   out_struct_fields : StructField.t ScopeVar.Map.t;
 }
 
+(** In practice, this is a DAG: beware of repeated names *)
+type module_tree = M of module_tree ModuleName.Map.t [@@caml.unboxed]
```
denismerigoux, line 677: Why don't you use an `ocamlgraph` type to represent the DAG then?

compiler/surface/lexer.cppo.ml
```diff
@@ -819,6 +819,15 @@ let line_dir_arg_re =
       eol
     ])
 
+let line_dir_arg_upcase_re =
```
denismerigoux, line 822: What does this catch?

dune
```diff
@@ -10,7 +10,7 @@
  ; don't stop building because of warnings
  (dev
   (flags
-   (:standard -warn-error -a)))
+   (:standard -warn-error -a -w -67)))
```
denismerigoux, line 13: Why did you exclude `-67` warnings?

tests/test_modules/bad/mod_bad_include.catala_en
```diff
@@ -0,0 +1,21 @@
+> Include: mod_badname.catala_en
+
+
+```catala-test-inline
+$ catala typecheck
+[ERROR]
+A file that declares a module cannot be used through the raw '> Include' directive. You should use it as a module with '> Use This_is_not_the_file_name' instead.
```
denismerigoux, line 7: I love this error

tests/test_modules/bad/mod_badname.catala_en
```diff
@@ -0,0 +1,14 @@
+> Module This_is_not_the_file_name
+
+
+```catala-test-inline
+$ catala typecheck
+[ERROR]
+Module declared as This_is_not_the_file_name, which does not match the file name "tests/test_modules/bad/mod_badname.catala_en"
```
denismerigoux, line 7: "... To match the filename, the module should be named XXX"

tests/test_modules/good/mod_use2.catala_en
```diff
@@ -1,10 +1,10 @@
-> Using Mod_middle
+> Using Mod_middle as M
```
denismerigoux, line 1: Yes!

denismerigoux (approved):
Thank you @AltGr for this very clear and thorough improvement!

compiler/shared_ast/definitions.ml
AltGr: It wouldn't be convenient for lookup by key, which is the main use-case here: we only have to be careful about repetitions during the initial load

compiler/surface/lexer.cppo.ml
AltGr: Yeah here we are using a low-cost parser on the directive lines, but the trouble is that e.g. the french translation uses many words (`> Usage de MyModule en tant que M`) so recovering the actual parameters can't be done with simple word indexing (the previous version would only return the last `M` here). Rather than enriching the line-lexer for sub-tokens on these directives and then a more serious parser, the cheaper solution here was to just rely on capitalisation of the module names :sweat_smile:

dune
AltGr: it's about unused functor parameters, so unlikely to be problematic, and most annoyingly is triggered on `.mli` files, forcing you to write `module Foo (_: TARG): ...` instead of `module Foo (MeaningfulArgName: TARG): ...`. Which I find less helpful.

tests/test_modules/bad/mod_bad_include.catala_en
AltGr: It's an arbitrary limitation tho ;)

dune
denismerigoux: OK fine :)

compiler/surface/lexer.cppo.ml
denismerigoux: Woupsy that's kind of sketchy, this deserves at least a comment :)

compiler/driver.ml
AltGr: Takes 5x as much space but fine ;)

PR #538, opened 2023-12-01
Document and first test for externals

AltGr:
- [x] based on #535 and waiting for merge and rebase

- [ ] is `doc/devel` a correct place to document this kind of experimental
  features ?

- [ ] maybe add a warning or error on modules dubbed `external` if they contain
  code

Although clumsy this is actually fairly simple in practice...

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -578,12 +578,15 @@ let gen_build_statements
       ~outputs:[inc srcv]
   in
   let module_deps =
-    Option.map (fun m ->
+    Option.map
+      (fun m ->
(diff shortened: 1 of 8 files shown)


AltGr:
Quick link to the mentionned doc: https://github.com/CatalaLang/catala/blob/7034e229b11eef24de73f23de07eb9974196b3a7/doc/devel/externals.md

AltGr:
Ah that could be a good idea ; but isn't the `.mld` for developers on the Catala compiler itself (while this — *in theory* — could target users of the language only) ?

AltGr:
Thanks!

PR #541, opened 2023-12-04
Adelaett invariant typing default

adelaett:
Adding Typing Invariant for TDefault

Added a new type safety invariant to ensure that the type `TDefault` can only appear in certain positions,

* On the left-hand side of an arrow with arity 1, as the type of a scope (for scope calls).
* At the root of the type tree (outside a default).
* On the right-hand side of the arrow at the root of the type (occurs for rentrant variables).

This is crucial to maintain the safety of the type system, as demonstrated in the formal development.

The invariant was checked on all tests cases and on family and housing benefits.

Adjusted inversion invariant about app to handle external objects as well.

Depends on #535. Putting this PR as draft in the meantime.

### DIFF
--- a/compiler/dcalc/invariants.ml
+++ b/compiler/dcalc/invariants.ml
@@ -19,7 +19,7 @@ open Ast
 open Catala_utils
 
 type invariant_status = Fail | Pass | Ignore
-type invariant_expr = typed expr -> invariant_status
+type invariant_expr = decl_ctx -> typed expr -> invariant_status
 
 let check_invariant (inv : string * invariant_expr) (p : typed program) : bool =
   (* TODO: add a Program.fold_left_map_exprs to get rid of the mutable
@@ -33,7 +33,7 @@ let check_invariant (inv : string * invariant_expr) (p : typed program) : bool =
         (* let currente = e in *)
         let rec f e =
           let r =
-            match inv e with
+            match inv p.decl_ctx e with
             | Ignore -> true
             | Fail ->
               Message.raise_spanned_error (Expr.pos e) "%s failed\n\n%a" name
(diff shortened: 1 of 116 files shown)


adelaett:
things left to do : 

- [ ] add check invariant in the make test to ensure invariants are always true
- [ ] double check the invariant is enough to proof the progress.

compiler/dcalc/invariants.ml
```diff
@@ -121,6 +122,103 @@ let invariant_match_inversion () : string * invariant_expr =
...
+    let s = EnumName.Map.find n ctx.ctx_enums in
+    EnumConstructor.Map.for_all (fun _k ty -> check_typ_no_default ctx ty) s
+  | TOption ty -> check_typ_no_default ctx ty
+  | TArrow (args, res) ->
+    List.for_all (check_typ_no_default ctx) args && check_typ_no_default ctx res
+  | TArray ty -> check_typ_no_default ctx ty
+  | TDefault _t -> false
+  | TAny ->
+    (* found TAny *)
+    Message.emit_warning
```
AltGr, line 166: `emit_warning "Internal error"` seems contradictory.
I believe this could be just a debug message ? After all it might be legitimate to want to check what can be checked in specific passes...

compiler/driver.ml
```diff
@@ -541,14 +541,27 @@ module Commands = struct
     (* Strictly type-checking could stop here, but we also want this pass to
        check full name-resolution and cycle detection. These are checked during
        translation to dcalc so we run it here and drop the result. *)
-    let _prg = Dcalc.From_scopelang.translate_program prg in
+    let prg = Dcalc.From_scopelang.translate_program prg in
+    let prg =
```
AltGr, line 545: This re-typing should only be done if required by `check_invariants`: in normal use everything that needs to be checked at this point has already been done.

AltGr (approved):
A little nitpicking, but otherwise all good

compiler/driver.ml
adelaett: done!

PR #546, opened 2023-12-07
[breaking changes] update syntax and CLI

### LINKED ISSUE #534: Renaming "collection" as "list"
The type "collection" should be renamed to "list" ("liste" in French). We should also specify "list of integer" instead of "list integer" (if the parser allows). This renaming will avoid any confusion that the user could have about order-preservation properties of operations on collections.

AltGr:
- Unify all CLI args to use `-` rather than `_`
- Change syntax to use `list of` instead of `collection` (closes #534)


examples/droit_successions/tests/tests_droit_succession.catala_fr
```diff
@@ -69,20 +69,112 @@ champ d'application Test4:
 
 ```catala-test-inline
 $ catala Interpret -s Test1
-[RESULT] Computation successful!
+[ERROR]
```
denismerigoux, line 72: This test should be updated.

denismerigoux (approved):
Thanks @AltGr ! I'm okay with these breaking changes but we should : 
* publish a new version  `0.9.0` of the opam package with a changelog containing this 
* make sure `clerk` binary is included in the `catala` opam package (it's not the case at the moment)

denismerigoux:
Let's merge this quickly ?

AltGr:
As soon as the tests pass :)

PR #547, opened 2023-12-07
[DRAFT] use modules in the examples

AltGr:
Running the basic tests and computing Allocations or Aides au logement fully
works, *but* there are currently blocking limitations:

- the current scheme of having inline-tests with different options (*e.g.*
  `--avoid-exceptions) won't work with modules, for which these options need to be
  project-wide (they change the call convention, *ergo* the interface of modules)

- we also use them for generating Python, for which modules aren't implemented
  yet


AltGr:
Replaced by https://github.com/CatalaLang/catala-examples/pull/3

PR #553, opened 2023-12-19
Small optimisations

AltGr:
Three auxiliary changes that happened during a larger work but may warrant separate discussion:

- some more parens in the printer
- an optimisation in the interpreter that avoids an unnecessary full traversal, but is somewhat dirty (although accompanied with a compile-time "proof" that it is correct)
- an optimisation (in the sense of `catala -O`) that removes variable aliases. @denismerigoux I need your opinion on this one as it is pretty efficient, but might disrupt some of the invariants you expect after closure conversion ?

These are in 3 different patches that could be reviewed separately.

### DIFF
--- a/compiler/shared_ast/optimizations.ml
+++ b/compiler/shared_ast/optimizations.ml
@@ -215,12 +215,15 @@ let rec optimize_expr :
       when StructName.equal name name1 ->
       Mark.remove (StructField.Map.find field fields)
     | EErrorOnEmpty
-        (EDefault { excepts = []; just = ELit (LBool true), _; cons }, _)
-      when false
-           (* FIXME: this optimisation is correct and useful, but currently
-              breaks expectations of the without-exceptions backend *) ->
+        ( EDefault
+            {
+              excepts = [];
+              just = ELit (LBool true), _;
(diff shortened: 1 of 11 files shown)


compiler/shared_ast/optimizations.ml
```diff
@@ -207,8 +207,9 @@ let rec optimize_expr :
       in
       EMatch { e = arg; cases; name = n1 }
     | EApp { f = EAbs { binder; _ }, _; args }
-      when binder_vars_used_at_most_once binder ->
-      (* beta reduction when variables not used. *)
+      when binder_vars_used_at_most_once binder
```
denismerigoux, line 210: This supercedes the same thing I just did yesterday in the C backend :) https://github.com/CatalaLang/catala/pull/532/commits/91f845189907b921dbcb0a999008937b6aa19b1f

denismerigoux (approved):

denismerigoux:
Be careful, I also fixed this optim yesterday : [compiler/shared_ast/optimizations.ml](https://github.com/CatalaLang/catala/pull/532/commits/35f829cd656a017261a8596029e7895b07020ae5)

PR #554, opened 2023-12-19
Add tuples to the surface language, allow tuples for argument passing

AltGr:
Upon adding tuples to the surface language, we wanted to ensure that, at least
on surface, there was no difference between applying a function to a pair or to
two arguments.

Technically, defining a function of a pair is forbidden (and not syntactically
possible), and during conversion to `scopelang`, we process with de-tuplification of
function arguments when needed. This transformation is based on typing
information that is of the same kind as what is used for overload resolution.

Towards this purpose, explicit types have been added to application nodes in the
AST ; besides and for convenience, the application of operators is split into a
different `EAppOp` node, replacing the plain `EOp` node.


compiler/scopelang/from_desugared.ml
```diff
@@ -130,22 +131,50 @@ let rec translate_expr (ctx : ctx) (e : D.expr) : untyped Ast.expr boxed =
              ScopeVar.Map.add v' e' args')
            args ScopeVar.Map.empty)
       m
-  | EApp { f = EOp { op; tys }, m1; args } ->
+  | EApp { f; tys; args } -> (
+    (* Detuplification of function arguments *)
```
denismerigoux, line 135: You say in your PR description that de-tuplification happens during the translation to `lcalc`, but this is actually during the translation to `scopelang`.

compiler/shared_ast/optimizations.ml
```diff
@@ -100,61 +100,17 @@ let rec optimize_expr :
...
-            ( EOp { op = Not; _ }, _
-            | ( EApp
-                  {
-                    f = EOp { op = Log _; _ }, _;
-                    args = [(EOp { op = Not; _ }, _)];
-                  },
-                _ ) ) as op;
-          args = [e1];
-        } -> (
+    | EAppOp { op = Not; args = [(ELit (LBool b), _)]; _ } ->
```
denismerigoux, line 103: OK I see, because the logging call wraps the `EAppOp`, you no longer need to match on it. Clever!

compiler/shared_ast/print.mli
```diff
@@ -125,3 +125,8 @@ module UserFacing : sig
   (** This combines the user-facing value printer and the generic expression
       printer to handle all AST nodes *)
 end
+
+(**/*)
```
denismerigoux, line 129: Coding artefact to clean?

compiler/shared_ast/typing.mli
```diff
@@ -61,9 +61,12 @@ val expr :
...
     what you want.
 
-    Note that typing also transparently performs disambiguation of constructors:
-    [EDStructAccess] nodes are translated into [EStructAccess] with the suitable
-    structure and field idents (this only concerns [desugared] expressions). *)
+    Note that typing also transparently performs
+    - disambiguation of constructors: [EDStructAccess] nodes are translated into
+      [EStructAccess] with the suitable structure and field idents (this only
+      concerns [desugared] expressions).
+    - resolution of operator types, which are stored (monomorphised) back in the
```
denismerigoux, line 68: They are stored in `EAbs`, `EApp` and `EAppOp` then?

compiler/surface/ast.ml
```diff
@@ -180,13 +181,14 @@ and naked_expression =
   | FunCall of expression * expression list
   | ScopeCall of
       (path * uident Mark.pos) Mark.pos * (lident Mark.pos * expression) list
-  | LetIn of lident Mark.pos * expression * expression
+  | LetIn of lident Mark.pos list * expression * expression
```
denismerigoux, line 184: This is the first step towards patterns I guess?

tests/test_func/bad/bad_func.catala_en
```diff
@@ -33,17 +33,17 @@ $ catala Interpret -s S
 There is a conflict between multiple valid consequences for assigning the same variable.
 
 This consequence has a valid justification:
-┌─⯈ tests/test_func/bad/bad_func.catala_en:14.65-14.70:
+┌─⯈ tests/test_func/bad/bad_func.catala_en:14.67-14.68:
 └──┐
 14 │   definition f of x under condition (x >= x) consequence equals x + x
-   │                                                                 ‾‾‾‾‾
+   │                                                                   ‾
```
denismerigoux, line 39: So this is a little bit unfortunate. I have the impression that when you build the `EAppOp`, you take the mark of the operator rather than taking the mark of the entire expression. I would be in favor of reverting to the previous behavior.

tests/test_tuples/good/tuples.catala_en
```diff
@@ -0,0 +1,43 @@
...
+```
+
+```catala-test-inline
+$ catala typecheck
+[RESULT] Typechecking successful!
+```
+
+```catala-test-inline
+$ catala interpret -s Test
+[RESULT] Computation successful! Results:
```
denismerigoux, line 41: Could you add an `catala invariants` test, as well as a `catala interpret_lcalc` test ?

denismerigoux (approved):
Thank you very much @AltGr! As always, a few benign remarks before merging but this is excellent work.

compiler/desugared/from_surface.ml
```diff
@@ -1274,6 +1289,7 @@ let process_topdef
   let translate_typ t = Name_resolution.process_type ctxt t in
   let translate_tbase (tbase, m) = translate_typ (Base tbase, m) in
   let typ = translate_typ def.S.topdef_type in
+  Message.emit_debug "TTYP: %a@." Print.typ_debug typ;
```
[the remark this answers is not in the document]
AltGr: Oops indeed!

compiler/lcalc/compile_without_exceptions.ml
```diff
@@ -59,36 +59,55 @@ let rec translate_default
...
+      mark_default
   in
   exceptions
 
 and translate_expr (e : 'm D.expr) : 'm A.expr boxed =
   let mark = Mark.get e in
   match Mark.remove e with
-  | EEmptyError -> Ast.OptionMonad.empty ~mark
+  | EEmptyError ->
+    Expr.einj ~e:(Expr.elit LUnit mark) ~cons:Expr.none_constr
```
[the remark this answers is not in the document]
AltGr: The `bind` operation wasn't used anymore, so it made little sense to keep a complex module just for a single instance for each of `return` and `empty`, that are actually well expressed in this `translate` function. Granted, this is not really related to the rest of the PR (although IIRC it helped with some typing issues), but this was a pending cleanup done with @adelaett: the more expressive typing for defaults made this simplification possible.

compiler/scopelang/from_desugared.ml
AltGr: Ah indeed you're right, that changed since the initial prototype (fixed in the description)

compiler/shared_ast/print.mli
AltGr: Ah no this is in fact an `ocamldoc` directive that tells it to ignore everything that follows from the generated documentation. It allows to keep the proper doc for `skip_wrappers` in `print.mli`, while hiding it to avoid confusion since `Expr.skip_wrappers` should be used instead.

compiler/shared_ast/typing.mli
AltGr: Indeed `EApp` should be added to this list, I'll clarify this comment. `EAbs` nodes can be changed as well when their types wasn't fully specified (it may happen from generated code only), but that was already the case.

PR #557, opened 2024-01-10
Unify rounding operations in OCaml runtime

pierregoutagny:
4 operations in the OCaml runtime implemented different rounding schemes. This PR adds a unique `round` function to round rationals to bigints the way that it was originally intended. `round` rounds rationals to the nearest bigint. On ties, the "half away from zero" rule is used. For example, `0.5` is rounded to `1.0` and `-0.5` is rounded to `-1.0`.

This function is documented, the 4 operations that use rounding are updated to use it, and their documentations explicitly refer to it. Moreover, a new test `tests/test_arithmetic/good/rounding.catala_en` shows the intended behaviour on key examples.

This PR does not change unit tests. However, some assertions in examples (`examples/aides_logement`) break. I have not updated them because as I understand it they come from official guides, so I think they should be checked more carefully before fixing.

### DIFF
--- a/examples/aides_logement/tests/tests_calcul_apl_accession_propriete.catala_fr
+++ b/examples/aides_logement/tests/tests_calcul_apl_accession_propriete.catala_fr
@@ -33,7 +33,7 @@ champ d'application Exemple1:
     calcul.aide_finale_formule
   assertion montant = 181,91 €
   assertion calcul.mensualité_éligible = 533,91 €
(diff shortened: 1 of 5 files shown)


denismerigoux:
The computation of `mensualité_minimale` and `equivalence_loyer_minimale` has always been a thorny rounding-related topic in the APL computation where even the official guidelines are blurry. I'm okay with the fact that your rounding fix offsets by one cent some of these results.

PR #560, opened 2024-01-16
Fixes missing minus sign printing for small negative amounts of money

### LINKED ISSUE #555: UserFacing printer forgets minus sign for money in (-1, 0)
I think the exact same bug as in #551 also occurs for money, even after the #552 patch. Sorry for not catching both at the same time.

denismerigoux:
Fixes #555.


AltGr (approved):
I am shocked to find out that `-$12` is actually the correct convention :sweat_smile:

denismerigoux:
I don't know if it's the correct convention but it's the least ugly to me!

PR #562, opened 2024-01-17
[WIP] Rehaul CI to run examples and generate artifacts again

AltGr:
this requires to fetch catala-examples and french-law from their own repos


AltGr:
Green, finally!
The merge will allow us to see if the github-pages publication still works (but that part was mostly untouched... ?)

PR #564, opened 2024-01-19
Allow literal scope output structures

### LINKED ISSUE #563: Be able to use scope output struct as a real struct (build object)
Suppose the following scope : 

```catala 
declaration scope Foo:
  result a content boolean 
  result b content integer 
```

Now suppose you're using the output struct of `Foo` later in another calling scope. Right now you cannot build a new `Foo` object to, let's say, map the contents of the output of `Foo` into a new `Foo` object with tweaked values. This restriction should be lifted.

AltGr:
Closes #563


AltGr:
(cancelled the CI because I don't want to interfere with the ongoing rehaul, but this should be fine)

denismerigoux (approved):
Super cool, that was an easy fix, thanks!

PR #566, opened 2024-01-23
Improve missing field error message

denismerigoux:


AltGr (approved):

compiler/desugared/from_surface.ml
```diff
@@ -528,13 +528,22 @@ let rec translate_expr
...
-      (fun expected_f _ ->
-        if not (StructField.Map.mem expected_f s_fields) then
-          Message.raise_spanned_error pos
-            "Missing field for structure %a: \"%a\"" StructName.format s_uid
-            StructField.format expected_f)
-      expected_s_fields;
+    if
+      StructField.Map.exists
+        (fun expected_f _ -> not (StructField.Map.mem expected_f s_fields))
+        expected_s_fields
```
AltGr, line 534: I generally find it nicer to write `let missing = Map.diff expected fields in if not (Map.is_empty missing) then ...`, but it's generally less efficient and well `Map` doesn't have a proper `diff` function (although you can use `merge` for that :shrug:) so this is fine ;)

denismerigoux: I didn't know you coud use `Map.merge` to implement some sort of `Map.diff`. I'll remember it next time :)

PR #568, opened 2024-01-25
Support for list recombinations

AltGr:
The primary use-case for this was to be able to run computations on a list
of structures, then return an updated list with some fields in the
structures modified : that is what we need for distribution of tax amounts
among household members, for example.

This patch has a few components:

- Addition of a test as an example for tax distributions (and tests for various
  cases of the below desugaring)

- Internally added a `Map2` operator that is used as backend for the following

- Added a transformation, performed during desugaring, that -- where lists
  are syntactically expected, i.e. after the `among` keyword -- turns a
  (syntactic) tuple of lists into a list of tuples ("zipping" the lists)

- Arg-extremum transformation was also fixed to use an intermediate list
  instead of computing the predicate twice

- For convenience, allow to bind multiple variables in most* list
  operations (previously only `let in` and functions allowed it)

- Fixed the printer for tuples to differentiate them from lists

*Note: tuples are not yet allowed on the left-hand side of filters and
arg-extremums for annoying syntax conflict reasons.

### DIFF
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -226,6 +226,40 @@ let rec translate_expr
   let rec_helper ?(local_vars = local_vars) e =
     translate_expr scope inside_definition_of ctxt local_vars e
   in
+  let rec detuplify_list = function
+    (* Where a list is expected (e.g. after [among]), as syntactic sugar, if a
+       tuple is found instead we transpose it into a list of tuples *)
+    | S.Tuple ls, pos ->
+      let m = Untyped { pos } in
+      let ls = List.map detuplify_list ls in
+      let rec zip = function
+        | [] -> assert false
+        | [l] -> l
+        | l1 :: r ->
+          let rhs = zip r in
+          let rtys, explode =
+            match List.length r with
+            | 1 -> (TAny, pos), fun e -> [e]
+            | size ->
+              ( (TTuple (List.map (fun _ -> TAny, pos) r), pos),
+                fun e ->
+                  List.init size (fun index ->
+                      Expr.etupleaccess ~e ~size ~index m) )
+          in
+          let tys = [TAny, pos; rtys] in
+          let f_join =
+            let x1 = Var.make "x1" and x2 = Var.make "x2" in
+            Expr.make_abs [| x1; x2 |]
+              (Expr.make_tuple (Expr.evar x1 m :: explode (Expr.evar x2 m)) m)
+              tys pos
+          in
+          Expr.eappop ~op:Map2 ~args:[f_join; l1; rhs]
+            ~tys:((TAny, pos) :: List.map (fun ty -> TArray ty, pos) tys)
+            m
+      in
+      zip ls
+    | e -> rec_helper e
+  in
   let pos = Mark.get expr in
   let emark = Untyped { pos } in
   match Mark.remove expr with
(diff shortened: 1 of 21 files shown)


AltGr:
(note: CI shows red because of a quirk that made it attempt to publish its artifacts, but tests actually passed)

compiler/desugared/from_surface.ml
```diff
@@ -226,6 +226,40 @@ let rec translate_expr
...
+            Expr.make_abs [| x1; x2 |]
+              (Expr.make_tuple (Expr.evar x1 m :: explode (Expr.evar x2 m)) m)
+              tys pos
+          in
+          Expr.eappop ~op:Map2 ~args:[f_join; l1; rhs]
+            ~tys:((TAny, pos) :: List.map (fun ty -> TArray ty, pos) tys)
+            m
+      in
+      zip ls
+    | e -> rec_helper e
```
denismerigoux, line 261: No error message at this point ? Does this mean that the error will be raised later during typing?

compiler/desugared/from_surface.ml
```diff
@@ -629,16 +663,39 @@ let rec translate_expr
...
+      (* Detuplification (TODO: check if we couldn't fit this in the general
+         detuplification later) *)
+      match List.length param_names with
+      | 1 -> f_pred
+      | nb_args ->
+        let v =
+          Var.make (String.concat "_" (List.map Mark.remove param_names))
+        in
+        let x = Expr.evar v emark in
+        let tys = List.map (fun _ -> TAny, pos) param_names in
```
denismerigoux, line 690: Is there any way to be smarter than `TAny` here or in the other instances of changes in this file?

runtimes/ocaml/runtime.mli
```diff
@@ -353,6 +353,7 @@ module Oper : sig
   val o_xor : bool -> bool -> bool
   val o_eq : 'a -> 'a -> bool
   val o_map : ('a -> 'b) -> 'a array -> 'b array
+  val o_map2 : ('a -> 'b -> 'c) -> 'a array -> 'b array -> 'c array
```
denismerigoux, line 356: Missing comment : `raises NotSameLength`

tests/test_modules/good/prorata_syntax.catala_en
```diff
@@ -0,0 +1,65 @@
+> Using Prorata_external as Ext
```
denismerigoux, line 1: Where is `Prorata_external` defined ? Is this really how you're supposed to use this builtin? Will it be the same for other builtins?

denismerigoux (approved):
Thank you @AltGr !!

tests/test_modules/good/prorata_syntax.catala_en
AltGr: https://github.com/CatalaLang/catala/blob/master/tests/test_modules/good/prorata_external.ml

AltGr: https://github.com/CatalaLang/catala/blob/master/tests/test_modules/good/prorata_external.catala_en

AltGr: (far from perfect but there is an HOWTO about how to handle it in its current state: https://github.com/CatalaLang/catala/blob/master/doc/devel/externals.md)

PR #569, opened 2024-01-25
Reduce default verbosity of invariant checks

AltGr:
they break the tests too often for no good reason


denismerigoux (approved):
Good call!

PR #570, opened 2024-01-26
Add a few optimisations for combined list operations

AltGr:
useful with the previous pr about list recombinations ; also improve the
internal namings during that desugaring.


denismerigoux (approved):
Perfect, thank you @AltGr !

PR #571, opened 2024-01-30
Attempt to improve type annotations on the dcalc->lcalc translations

AltGr:
This passes all current tests and fixes the examples broken in #532 ; but might break some in-progress invariants of the c backend.

What I attempted to do is ensuring to add correct type annotations when transforming dcalc to lcalc (for both of the with/without exceptions versions) — when the translation is done on a typed AST of course (note that while I computed the correct types in the cases where it wasn't too complex, it's in general fine to use `TAny` there, what should be avoided is to copy annotations, keeping their types where they're not correct anyore)

- in these changes, I removed the need for `compile_without_exceptions` to run on a typed AST and to copy the type annotations into the `EAppOp` parameters. Since we don't need to untype the whole term anymore, I expect the remaining type annotations should make this unnecessary, but I'm not 100% sure it doesn't break anything @denismerigoux has been working on, so this part may need to be reverted.
- I didn't complete the same effort for the `closure_conversion` pass, but expressed the fact that it doesn't preserve types with an explicit `untype` at the end (that should at least be replaced by a removing of type annotations done during the main traversal)

### DIFF
--- a/compiler/shared_ast/expr.mli
+++ b/compiler/shared_ast/expr.mli
@@ -174,6 +174,8 @@ val ecustom :
   (< custom : Definitions.yes ; .. >, 'm) boxed_gexpr
 
 val fun_id : ?var_name:string -> 'm mark -> ('a any, 'm) boxed_gexpr
+(** The type of the mark, if typed, is assumed to correspond to the argument
+    type, not the function type *)
 
 (** {2 Manipulation of marks} *)
 
@@ -342,7 +344,7 @@ val make_erroronempty :
 val empty_thunked_term :
   'm mark -> (< defaultTerms : yes ; .. >, 'm) boxed_gexpr
 
-val thunk_term : ('a any, 'b) boxed_gexpr -> 'b mark -> ('a, 'b) boxed_gexpr
+val thunk_term : ('a any, 'b) boxed_gexpr -> ('a, 'b) boxed_gexpr
 val unthunk_term_nobox : ('a any, 'm) gexpr -> 'm mark -> ('a, 'm) gexpr
 
 val make_let_in :
--- a/compiler/shared_ast/expr.mli
+++ b/compiler/shared_ast/expr.mli
@@ -366,6 +366,9 @@ val make_tuple :
 (** Builds a tuple; the mark argument is only used as witness and for position
     when building 0-uples *)
 
+val make_tupleaccess :
+  ('a any, 'm) boxed_gexpr -> int -> int -> Pos.t -> ('a, 'm) boxed_gexpr
+
 (** {2 Transformations} *)
 
 val skip_wrappers : ('a, 'm) gexpr -> ('a, 'm) gexpr
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -689,7 +689,7 @@ let translate_rule
       | Runtime.Reentrant, pos -> (
         match Mark.remove tau with
         | TArrow _ -> new_e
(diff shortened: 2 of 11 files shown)


compiler/shared_ast/expr.mli
```diff
@@ -366,6 +368,9 @@ val make_tuple :
 (** Builds a tuple; the mark argument is only used as witness and for position
     when building 0-uples *)
 
+val make_tupleaccess :
```
denismerigoux, line 371: What's the difference with `etupleaccess` ?

compiler/shared_ast/typing.ml
```diff
@@ -312,13 +312,9 @@ let polymorphic_op_type (op : Operator.polymorphic A.operator Mark.pos) :
...
     | Log _ -> [any] @-> any
     | Length -> [array any] @-> it
-    (* The [HandleDefault] and [HandleDefaultOpt] need to be typed before and
-       after the Lcalc monomorphization which affects arrays and option types.
-       Because of that, we give the operators very lax typing rules with [any]
-       but it doesn't matter for unification because the concrete types on which
-       they will be instantiated are stored in the [EAppOp] node. *)
-    | HandleDefault -> [any2; [ut] @-> bt; [ut] @-> any] @-> any
-    | HandleDefaultOpt -> [any2; [ut] @-> bt; [ut] @-> any] @-> any
+    | HandleDefault ->
```
denismerigoux, line 315: This will break the C backend for sure, because the monomorphization pass changes the `Array` type and expression (to make monomorphic array). In this case, the type of `HandleDefaultOpt` is no longer `array any -> ...` but `(struct monomorphized_array) -> ...`. This is why I put `any2` in here.

denismerigoux (approved):
OK for this PR that seems to simplify things. But I'm testing first if it breaks things on my side for the C backend.

compiler/shared_ast/typing.ml
AltGr: Hm that's right... what I don't understand though is how the other operators working on arrays can be typed then ? Or have they disappeared at this point ?

Another interrogation is wether we really need to re-infer types after monomorphisation. If yes, it might be worth wondering if we could switch to a different typing for operators at that point.

An sneaky option could, maybe, to skip that function if the full annotations are already in place.

Of course, I am exploring things here but don't want to hinder progress on the C branch: feel free to ignore any part of this to go on, and this kind of improvements can be kept for later if needed at all.

denismerigoux:
No sorry Louis, this PR breaks the careful typing efforts I had done in the `c_backend` PR. Simply try `dune exec -- catala C tests/test_backends/simple.catala_en  -d` here and you'll see that the changes in this PR cause a lot of types required for the monomorphization pass to be absent. I've been debugging for the last hour but do not see and end to it, your PR makes me redo a lot of the work I've done on `c_backend` to have all the types correctly resolved for monomorphization.

compiler/shared_ast/typing.ml
denismerigoux: The other operators working on arrays will also have to have their typing tweaked here, it's just that I did not reach the point in the `c_backend` PR where I'm dealing with them.

denismerigoux: Yes we need to re-infer types after monomorphization. The reason for that is that it helps me debug my monomorphization pass. Furthermore, right now all my monomorphization tweaks are suhch that I can manually put the new type in the resulting terms, but that may not be true for future monomorphization tweaks.

AltGr:
Ok, sorry about that, then I guess we should keep it like this for now, and I'll later try to see if we can avoid it by not stripping away all type anotations at `closure_conversion` time

PR #573, opened 2024-02-06
Correctly pass types through monomorphisation

AltGr:
This includes a simplification of the monomorphisation pass that lets it
correctly propagate type annotations ; consequently, retyping afterwards is no
longer needed and we can stop being too lax on the typing of the HandleDefault
operators -- which didn't scale to code containing operations on lists anyway.

This also allows compilation to honor the `--no-typing` flag further, which can
help with debugging (it's easier to see the mistyped transformation to fix
them).

Of course, it may also be pretty useful to be able to call the typer even in the
later phases; but the correct way to do that will be to have the typer actually
handle the new operators. There could be a few ways:

- pass a translation table through the typing env (but that means it would need
  to always be stored besides the program once it's mono'd)

- add an option to the typer to "trust" the types of the arguments included in
  the `EAppOp` nodes so that it actually won't need to recheck that part (it'll
  still need specific code to compute the output type). This could be understood
  as a way of encoding that the op calls are now monomorphic ?

- since we monomorphise, shouldn't the polymorphic operators be translated
  anyway ? I guess it depends on the backend, but that would make sense ;
  In this case, the mono pass would actually replace these operators by specific
  backend function calls that are generated later on, and that will have the
  proper type.

### DIFF
--- a/compiler/shared_ast/print.ml
+++ b/compiler/shared_ast/print.ml
@@ -474,13 +474,20 @@ end
 
 module ExprGen (C : EXPR_PARAM) = struct
   let rec expr_aux :
-      type a.
+      type a t.
       Bindlib.ctxt ->
       Ocolor_types.color4 list ->
       Format.formatter ->
-      (a, 't) gexpr ->
+      (a, t) gexpr ->
       unit =
    fun bnd_ctx colors fmt e ->
+    (* (* Uncomment for type annotations everywhere *)
+     * (fun f ->
+     *    Format.fprintf fmt "@[<hv 1>(%a:@ %a)@]"
+     *      f e
+     *      typ_debug
+     *      (match Mark.get e with Typed {ty; _} -> ty | _ -> TAny,Pos.no_pos))
+     * @@ fun fmt e -> *)
     let exprb bnd_ctx colors e = expr_aux bnd_ctx colors e in
     let exprc colors e = exprb bnd_ctx colors e in
     let expr e = exprc colors e in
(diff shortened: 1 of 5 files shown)


compiler/shared_ast/print.ml
```diff
@@ -474,13 +474,20 @@ end
...
-      type a.
+      type a t.
       Bindlib.ctxt ->
       Ocolor_types.color4 list ->
       Format.formatter ->
-      (a, 't) gexpr ->
+      (a, t) gexpr ->
       unit =
    fun bnd_ctx colors fmt e ->
+    (* (* Uncomment for type annotations everywhere *)
```
denismerigoux, line 484: Aaaaargh this is so cool it should be activatable with a flag to `Expr.print` in my opinion.

denismerigoux (approved):
Ok you convinced me Louis, as always :) I'll manage to continue implement and debugging my C backend without re-typing after monomorphization I think.

AltGr:
In the meantime I guess option 2 above would be the easiest: it shouldn't be too difficult to add and would serve with debugging -- like what we had previously but scaling to collection operations.

denismerigoux:
> add an option to the typer to "trust" the types of the arguments included in
the EAppOp nodes so that it actually won't need to recheck that part (it'll
still need specific code to compute the output type). This could be understood
as a way of encoding that the op calls are now monomorphic ?

Yes this would be the most sensible option I guess

PR #575, opened 2024-02-07
Typing: add a "assume operator types" mode

AltGr:
This allows for retyping after monomorphisation: a new function just
extracts the return type of the operator, without checking the operand
types.

Also to avoid multiplying function arguments around the typer, the flags
have been gathered in a record that is included in the typing environment;
it's ok to give them default values as long as these are the strictest.

Also includes a small printer bonus.


AltGr:
(follow-up to #573)

denismerigoux (approved):
Thank you @AltGr !

PR #577, opened 2024-02-12
CI: fix the check for formatting and non properly promoted files

AltGr:


AltGr:
Ok this validates that the check now works: https://github.com/CatalaLang/catala/actions/runs/7874264275/job/21484502868#step:3:21

Issue #567, opened 2024-01-23
Being able to refer to one particular variable state outside of definition


denismerigoux:
## Edit: current progress
- [X] access an intermediate state of a variable within another definition in the same scope
- [ ] access a specific state of a variable within a definition of another state of the same variable, with the specific warning described below (the access is trivial, the warning complex to implement)

## The problem

Suppose we have the following variable declaration : 

```catala
internal foo 
  state one
  state two
```

Since #114, when referring to `foo` outside the definition of `foo`, you'll actually pick `foo.two` in the last state. But in real-world examples, you still need sometimes to use an earlier state of `foo` outside of the definition of `foo`. Right now, there is no way to do that with the surface syntax, even though under the hood `foo.one` and `foo.two` are two separate variables with separate dependency tracking, etc.

## Solution

A proposed solution is to add a surface explicit syntax to refer to a particular state of the variable outside of its definition. The syntax, at the level of expressions could be like that : 

```catala 
foo # refers implicitly to the last state as currently
(foo state one) # refers to the result of the definition of state one of foo
```

denismerigoux:
@AltGr what do you think ? Do we proceed with the syntax committee for this next week ?

AltGr:
Yes! Maybe `foo at state one` ?
Important note: this should be available only within the scope where the variable is defined.

AltGr:
Hm I have an additional interrogation here:
- Accessing `foo state one` from another definition within the same scope → ✅ 
- Accessing `foo state one` from another scope → 🟥 
- ... but what about accessing `foo state one` from *the definition of foo itself* (or more precisely, an earlier state thereof) ?
  - My idea at the moment is to allow it as long as it's a reference to a strictly earlier state
  - Another option is to forbid it entirely, it might bar some use-cases (?) but it keeps strong the current meaning that a state always refers (implicitely) to the state just before. Users without a clear knowledge of this might start referring to the previous stage explicitely, which would break the code if a new state is later inserted in-between.

denismerigoux:
> Hm I have an additional interrogation here:
> 
>     * Accessing `foo state one` from another definition within the same scope → ✅
yes !
>     * Accessing `foo state one` from another scope → 🟥
yes !
>     * ... but what about accessing `foo state one` from _the definition of foo itself_ (or more precisely, an earlier state thereof) ?
>       
>       * My idea at the moment is to allow it as long as it's a reference to a strictly earlier state
[...]

PR #578, opened 2024-02-12
Allow access to intermediate variable states

AltGr:
Through the syntax `var_name state state_name`.

Addresses #567, but
- requires documentation
- explicitely accessing states of the variable being defined is not allowed
  at the moment (warning proposed in #567 requires a whole-expression check,
  which is better done some place else entirely)


AltGr:
Ah diff seems huge because I included a renaming of the tests (`tests/test_%/` → `tests/%/`) ; I can split that in a separate PR if it makes reviewing tedious

PR #580, opened 2024-02-13
Syntax: add keywords in front of list filtering / arg-extremum

### LINKED ISSUE #574: Collection syntax conflicts
# The parsing issue
## Current state
Among our adhoc syntaxes for collection operations, two are a source of trouble for the implementation:

- `x among some_list such that x > 1` (the result of which is a list)
- `x among some_list such that x*x is maximum or if list empty then 0` (the result of which is a single element)

Without entering into details, this is due to the fact that they start with a variable name, so when you first see the `x` you don't know yet if it needs to be parsed as the *expression* `x` (e.g. in ` x + x`) or as an identifier for one of the forms above ; and having this ambiguity that can only be decided later is a problem for parsing.

## Why it's a problem
For now it works because it's just a single variable, but it breaks if we want to add anything more complex, like for example a pair `(x, y)`, which is now allowed for other list operations (forms like `(x * y) for (x, y) among (list1, list2)`).

In the future we intend to further extend this and allow more complex forms, like maybe nested structures (it could be the same as in `match` patterns); 

# Proposed solutions
## A workaround
For tuples, a trick to make it possible can be done (allow an expression instead of `x`, and check that it has the correct format later on). However, it won't scale to more complex patterns.

## Syntax update
(which if decided should rather be done earlier than later)

Adding a keyword in front of the `x` would be enough to avoid the ambiguity and solve the problem. But we'd rather avoid adding too many new keywords to the language.

### Proposition 1
Use `let`:
- `let x among some_list such that x > 1`
- `let x among some_list such that x * x is maximum or if list empty then 0`

Personally I find this a bit confusing with the `let x equals ... in ...` form, but I am not sure how well that reads ?

### Proposition 2
Reuse words from type definitions:
- `list of x among some_list such that x > 1`
- `content of x among some_list such that x * x is maximum or if list empty then 0`

What do you think ?

AltGr:
Closes #574


AltGr:
Note: examples will break, I haven't adjusted them yet 😅

AltGr:
Examples have been updated in the `listopsyn` branch (https://github.com/CatalaLang/catala-examples/pull/1), which CI will automatically pick up.
I also quickly checked the ir-catala code which doesn't seem to have any occurence of the two cases that were changed.

Makefile
```diff
@@ -238,27 +238,33 @@ all: \
...
-	    _french-law -b $$(git branch --show-current) || \
+	    french-law.tmp -b $(BRANCH) || \
 	  git clone https://github.com/CatalaLang/french-law \
 	    --depth 1 --reference-if-able ../french-law \
-	    _french-law; } && \
-	make -C _french-law all PY_VENV_DIR=$(CURDIR)/_python_venv
+	    french-law.tmp || \
+	  git clone -s ../french-law french-law.tmp $(BRANCH) || \
+	  git clone -s ../french-law french-law.tmp master; } && \
+	make -C french-law.tmp all PY_VENV_DIR=$(ROOT_DIR)/_python_venv
```
denismerigoux, line 267: Don't really know the meaning of these changes but I trust you on them.

compiler/surface/parser.messages
denismerigoux, line 1: This file would be worth re-generating completely and merged with the previous one to fill in the new messages for the current version of the syntax.

denismerigoux (approved):
Thank you very much @AltGr ! Are you going to make a symmetric PR on `catala-examples` and `ir-catala` ?

Makefile
AltGr: - The temp dirs starting with `_` were causing problems because clerk would skip them, so they're now `catala-examples.tmp`, etc.
- I make sure the Python venv is created, and the generated files are up-to-date to mimic the check done by the CI
- I added fallbacks to cloning `../catala-examples` etc. for the target not to fail in case network is not available

AltGr:
@denismerigoux yes there is already https://github.com/CatalaLang/catala-examples/pull/1 (on a similarly-named branch, or the CI here wouldn't pass) ; they will need to be merged simultaneously.
as for `ir-catala`, no changes were actually needed, it doesn't yet use the syntax we changed !

compiler/surface/parser.messages
AltGr: yes, ugh

PR #582, opened 2024-02-15
Small ocaml/js output rehaul

AltGr:
Print to json directly rather than depend on yojson and a ppx.

Note: this should be tested with the website in order to validate that the
Json output is 1-to-1.
(a second step could be to simplify this output, now that it's manual)


AltGr:
Note for the future: here we want a clearly defined datatype for interchange, and we even have a plugin for generating JSON schemas.

These are things that [ATD](https://github.com/ahrefs/atd) does pretty well so that may be an option to consider at some point. The definitiion format is very straightforward and readable, and from it we can generate json schemas, and (de)serialisation code in OCaml, Python, Typescript and many others, both in JSON and binary format.

PR #583, opened 2024-02-15
Allow generation of one latex file from multiple sources, and fixes

AltGr:


compiler/catala_utils/cli.ml
```diff
@@ -448,6 +448,12 @@ module Flags = struct
           "Disables the search for counterexamples. Useful when you want a \
            deterministic output from the Catala compiler, since provers can \
            have some randomness in them."
+
+  let extra_files =
+    value
+    & pos_right 0 file []
+    & Arg.info [] ~docv:"FILE" ~docs:Manpage.s_arguments
+        ~doc:"Additional input files."
```
denismerigoux, line 456: `"These additional files are only taken into account when using the compiler to weave a literate programming output"`

denismerigoux (approved):
Thanks @AltGr !

compiler/catala_utils/cli.ml
AltGr: This is not a "global" argument so it will only appear in the doc of the `latex` subcommand (and the 2nd patch makes sure that we now fail with extra arguments on the other subcommands)

Issue #442, opened 2023-04-04
js_of_ocaml switch 4.1->5.0 breaks our OCaml->JS compilation of Catala code


denismerigoux:
The compilation to JS of `french_law/ocaml/api_web.ml` works fine with `js_of_ocaml.4.1.0` and produces in debug mode a JS binary of around 6MB.

But if you switch to `js_of_ocaml.5.1.1` then the size of the JS binary in debug mode explodes to 13MB and the code complains with

```text 
TypeError: Cannot read properties of undefined (reading 'iAvaitEnfantAChargeAvant1erJanvier2012In')
    at interface_allocations_familial$2 (/home/demerigo/catala/french_law/js/french_law.js:57482:39)
    at interface_allocations_familial$3 (/home/demerigo/catala/french_law/js/french_law.js:57553:12)
    at caml_call1 (/home/demerigo/catala/french_law/js/french_law.js:117722:15)
    at _a_ (/home/demerigo/catala/french_law/js/french_law.js:117754:19)
    at caml_call1 (/home/demerigo/catala/french_law/js/french_law.js:45302:15)
    at execute_or_throw_error (/home/demerigo/catala/french_law/js/french_law.js:45515:19)
    at caml_call1 (/home/demerigo/catala/french_law/js/french_law.js:117722:15)
    at /home/demerigo/catala/french_law/js/french_law.js:117759:22
    at caml_call_gen (/home/demerigo/catala/french_law/js/french_law.js:2960:15)
    at /home/demerigo/catala/french_law/js/french_law.js:4935:16
```

upon being called in `french_law/js/examples.js`

We need to investigate more what's changed in the passage from 4.1 to 5.0 of js_of_ocaml and why it breaks our compilation of OCaml to JS for the Catala code.

PR #584, opened 2024-02-22
Support externals in scalc, correctly output them in Python

AltGr:
Other scalc-powered backends (C, R, etc.) will fail with `"TODO"`

(but before this patch... that would actually be a stack overflow in lcalc→scalc
due to catch-all patterns)


compiler/scalc/print.ml
```diff
@@ -226,15 +229,22 @@ let format_item decl_ctx ?debug ppf def =
   Format.pp_close_box ppf ();
   Format.pp_print_cut ppf ()
 
-let format_program decl_ctx ?debug ppf prg =
+let format_program ?debug ppf prg =
   let decl_ctx =
+    (* TODO: this is redundant with From_dcalc.add_option_type (which is already
```
denismerigoux, line 234: Well caught, I myself had some confusion with these.

compiler/scalc/to_python.ml
```diff
@@ -611,50 +618,58 @@ let format_ctx
...
-           (Format.pp_print_list
-              ~pp_sep:(fun fmt () -> Format.fprintf fmt ", ")
-              (fun fmt (var, typ) ->
-                Format.fprintf fmt "%a:%a" format_var (Mark.remove var)
-                  format_typ typ))
-           func_params (format_block p.decl_ctx) func_body))
-    p.code_items
+  Format.pp_open_vbox fmt 0;
+  let header =
+    [
```
denismerigoux, line 657: Smart and beautiful

denismerigoux (approved):
Thanks @AltGr, works for me! There's just one test failing because of an extra line at the moment.

AltGr:
@denismerigoux I added quite a few more changes in there, in particular related to the new `test-scope` catala pseudo-command used in tests (and related fixes in the compiler) ; in case you want to have another look

Makefile
```diff
@@ -230,7 +253,7 @@ WEBSITE_ASSETS_ALL = $(WEBSITE_ASSETS) $(addprefix catala-examples.tmp/,$(WEBSIT
 
 website-assets-base: build
 	$(call local_tmp_clone,catala-examples) && \
-	dune build $(addprefix _build/default/,$(WEBSITE_ASSETS_ALL))
+	dune build $(addprefix _build/default/,$(WEBSITE_ASSETS_ALL)) --profile=release
```
denismerigoux, line 256: Be careful, the absence of the release flag here may be intentional (I don't remember well what I did) due to https://github.com/CatalaLang/catala/issues/442.

compiler/shared_ast/expr.ml
```diff
@@ -253,6 +253,9 @@ let typed = Typed { pos = Pos.no_pos; ty = TLit TUnit, Pos.no_pos }
 (* - Predefined types (option) - *)
 
 let option_enum = EnumName.fresh [] ("Eoption", Pos.no_pos)
+
+(* Warning: order of these definitions is important, binary injection assumes
```
denismerigoux, line 257: What's binary injection? The way they're stored inside an `EnumConstructor.Map.t` ?

compiler/shared_ast/interpreter.ml
```diff
@@ -497,7 +499,17 @@ let rec runtime_to_val :
     in
     let e = runtime_to_val eval_expr ctx m ty (Obj.field o 0) in
     EInj { name; cons; e }, m
-  | TOption _ty -> assert false
+  | TOption ty -> (
+    match Obj.tag o - Obj.first_non_constant_constructor_tag with
+    | 0 ->
+      let e =
```
denismerigoux, line 505: What does this do??

tests/README.md
denismerigoux, line 1: Thank you for having update this `README.md` ! The `README.md` in `catala-examples/` should also be updated to reflect these changes.

denismerigoux (approved):
I approve the new changes!

Makefile
AltGr: Actually it *was* there before my rewrite, but passed through the `DUNE_PROFILE` variable in the `website-assets` target, so normally that should be restoring the previous behaviour...

compiler/shared_ast/expr.ml
AltGr: Ah I mean the way we make compiled modules and the interpreter interact by translating between runtime OCaml values and AST value nodes on the fly. I added code to translate options, but it requires this assumption.

Makefile
denismerigoux: If that's the case fine then :)

compiler/shared_ast/interpreter.ml
AltGr: OCaml variants are discriminated by tag ; since we only use non-constant constructors (they all have a single parameter, which may be unit), querying the OCaml tag gives us the index in the Catala constructor map (i.e. the order in which the constructors have been defined in the OCaml code).

Our constructor maps use comparison based on their UIDs, which are attributed in increasing order, at declaration time, so `ENone` will be 0 and `ESome` will be 1. (But yes, that's quite a few assumptions)

PR #586, opened 2024-02-27
Some tweaks helping with separate compilation of the examples

AltGr:


runtimes/ocaml/runtime.mli
```diff
@@ -29,7 +29,11 @@ type integer = Z.t
 type decimal = Q.t
 type date = Dates_calc.Dates.date
 type duration = Dates_calc.Dates.period
-type date_rounding = Dates_calc.Dates.date_rounding
+
+type date_rounding = Dates_calc.Dates.date_rounding =
+  | RoundUp
```
denismerigoux, line 34: Why do you need to do that? To reexport the constructor names?

denismerigoux (approved):
Thanks @AltGr!

runtimes/ocaml/runtime.mli
AltGr: Yes, the generated code uses `RoundUp` directly, and the constructor wasn't in scope ; the other fix would have been to replace that with `Dates_calc.Dates.RoundUp` but this seemed nicer.

denismerigoux: I wasn't even aware of the possibility of doing that in OCaml ! It's indeed nicer :)

AltGr:
I expect CI to pass, but it requires the PR to be merged to take into account the changes to the CI scripts

AltGr:
See sister PRs:
- https://github.com/CatalaLang/catala-examples/pull/5
- https://github.com/CatalaLang/french-law/pull/2

→ CI here automatically picks them up (based on "same branch name" policy)
→ Merging the 3 PRs shall be done in sync for correct CI

compiler/driver.ml
```diff
@@ -989,6 +1007,80 @@ module Commands = struct
         $ Cli.Flags.optimize
         $ Cli.Flags.check_invariants)
 
+  let depends options includes prefix extension extra_files =
```
denismerigoux, line 1010: Can you remove the `Makefile` command of the Catala compiler now that we have this?

tests/scope/good/191_fix_record_name_confusion.catala_en
```diff
@@ -25,6 +25,7 @@ $ catala Typecheck --check-invariants
 
 ```catala-test-inline
 $ catala OCaml -O
+Generating entry points for scopes: ScopeA ScopeB
```
denismerigoux, line 28: What is this line? A comment in the resulting `OCaml` file? A debug print by the compiler on `stdout`?

denismerigoux (approved):
Thanks @AltGr!

Makefile
```diff
@@ -187,8 +194,8 @@ syntax:
 # High-level test and benchmarks commands
 ##########################################
 
-CATALA_OPTS?=
-CLERK_OPTS?=--makeflags="$(MAKEFLAGS)"
+CATALA_OPTS ?=
```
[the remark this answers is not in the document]
AltGr: Shell scripting frowns upon these spaces, but they're OK in Makefiles (and better for readability; besides; the equal operator in makefiles is closer to Prolog than to the simple assignment of shellscripts :fear)

Makefile
```diff
@@ -244,19 +251,23 @@ WEBSITE_ASSETS_EXAMPLES = \
   tutorial_en/tutorial_en.html \
   tutoriel_fr/tutoriel_fr.html \
   us_tax_code/us_tax_code.html \
-  allocations_familiales/allocations_familiales.html \
-  allocations_familiales/allocations_familiales_schema.json \
-  aides_logement/aides_logement.html \
-  aides_logement/aides_logement_schema.json
+  allocations_familiales/Allocations_familiales.html \
```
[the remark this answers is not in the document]
AltGr: Catala module names need to be capitalised, just like OCaml ones ; we tolerate file names that match up to capitalisation, but the catala module artifacts (e.g. generated ml files) will match the module names and therefore be capitalised. To handle all files related to a single origin catala file, it was much easier to uniformise the capitalisation and have `Allocations_familiales.*` match both source and all targets.

Makefile
```diff
@@ -244,19 +251,23 @@ WEBSITE_ASSETS_EXAMPLES = \
...
 
 WEBSITE_ASSETS_ALL = $(WEBSITE_ASSETS) $(addprefix catala-examples.tmp/,$(WEBSITE_ASSETS_EXAMPLES))
 
 website-assets-base: build
 	$(call local_tmp_clone,catala-examples) && \
-	dune build $(addprefix _build/default/,$(WEBSITE_ASSETS_ALL)) --profile=release
-
-website-assets.tar:
-	# $(MAKE) DUNE_PROFILE=release website-assets-base
+	$(MAKE) -C catala-examples.tmp \
```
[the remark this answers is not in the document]
AltGr: Ah yes it could be added, but the rule will anyway remove it right after termination (because of https://github.com/CatalaLang/catala/pull/586/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R297), unless you kill make very violently.

build_system/clerk_driver.ml
```diff
@@ -506,44 +504,57 @@ let base_bindings catala_exe catala_flags build_dir include_dirs test_flags =
...
       ~command:["cp"; "-f"; !input; !output]
       ~description:["<copy>"; !input];
 
     Nj.rule "catala-ocaml"
-      ~command:[!catala_exe; "ocaml"; !catala_flags; !catala_flags_ocaml; !input; "-o"; !output]
+      ~command:[!catala_exe; "ocaml"; !catala_flags; !catala_flags_ocaml;
+                !input; "-o"; !output]
       ~description:["<catala>"; "ocaml"; "⇒"; !output];
 
+    Nj.rule "ocaml-object"
```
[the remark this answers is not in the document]
AltGr: This remains fairly constrained, at least because we have full control on the .ml sources. But you should see the Makefile rules in catala-examples :grimacing: : at that point we are indeed building full-fledged OCaml libraries that we install to the prefix... At least we can still use `ocamlfind`

The thing is, dune is really uncooperative with other build systems, so I am not sure if we have a better choice. We could consider creating a mock OCaml project in a sandbox, running dune on that and extracting the results, but I am not sure that would be simpler to handle and it would be infinitely more complex in the processing that is involved.

As long as we control the ml files, the inter-dependencies and the target libraries we want, while certainly not pleasant, working through the build through manual use of the OCaml toolchain remains doable.

compiler/driver.ml
AltGr: Yes; it's not completely equivalent yet (no Makefile-style output, but support for modules) but the two should certainly be merged.
Ninja actually supports dependency specs in the Makefile style, so we could leverage that at some point (at the moment, a given file is likely to get parsed many times just to extract its dependencies...)

tests/scope/good/191_fix_record_name_confusion.catala_en
AltGr: Yes it's just a message... it was already there but got moved around :shrug:

AltGr: It was a rogue `eprintf`, I'll turn that into a debug.

PR #602, opened 2024-04-10
Improve message formatting throughout

AltGr:
PR #601 changed the interface without impacting the output format. This builds
on top of it to improve message formatting, both in the module `Message`, and in
individual messages to allow word-wrapping where it was not possible.

### DIFF
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -174,8 +174,8 @@ let rec disambiguate_constructor
   | [] ->
     if EnumName.Map.cardinal possible_c_uids > 1 then
       Message.error ~pos:(Mark.get constructor)
-        "This constructor name is ambiguous, it can belong to %a. Disambiguate \
-         it by prefixing it with the enum name."
+        "This constructor name is ambiguous, it can belong to@ %a.@ \
+         Disambiguate it by prefixing it with the enum name."
         (EnumName.Map.format_keys ~pp_sep:(fun fmt () ->
              Format.pp_print_string fmt " or "))
         possible_c_uids;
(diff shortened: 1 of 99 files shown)


compiler/desugared/from_surface.ml
```diff
@@ -1395,6 +1398,7 @@ let check_unlabeled_exception
         Message.error ~pos:(Mark.get item)
           ~pos_msg:(fun ppf -> Format.pp_print_text ppf "Ambiguous exception")
           ~extra_pos:(List.map (fun p -> "Candidate definition", p) pos)
+          "%a" Format.pp_print_text
```
denismerigoux, line 1401: What's the use of this?

tests/variable_state/bad/no_cross_exceptions.catala_en
```diff
@@ -16,8 +16,7 @@ scope A:
 
 ```catala-test-inline
 $ catala Typecheck 
-[ERROR]
-Unknown label for the scope variable foo@baz: "thing"
+[ERROR] Unknown label for the scope variable foo@baz: "thing"
```
denismerigoux, line 19: Maybe replacing `foo@baz` by `foo state baz` (localized depending on the language) would be better.

denismerigoux (approved):
Wonderful, thanks @AltGr !

compiler/desugared/from_surface.ml
AltGr: `pp_print_text` enables reformatting, e.g. turning all spaces into `@ ` which allows word-wrapping. This is much nicer than putting `@ ` everywhere in the message by hand, but unfortunately only works as long as the message is a pure string without formatting instructions.

It's a little ugly though, it could be nicer to have a flag for this ; but we're already juggling with optional and extensible arguments so I'm not sure it can be done in a reasonable way (unless of course we have a separate function `error_str`...)

tests/variable_state/bad/no_cross_exceptions.catala_en
AltGr: It'd definitely be more readable, yes, I'll see what can be done.

PR #603, opened 2024-04-11
Support for structure updates

### LINKED ISSUE #592: Syntactic sugar for changing only part of a structure
Let's imagine the following code:

> For purposes of this article, the income should be assumed to be at most $1200

and that we have a structure `Person` containing lots of `data` fields besides `income`: `age`, `xx`, `yy`,...

We would have to define the specific structure for the person as concerned by this article (supposing `original_person` is an input of the scope):
```catala
definition person_for_this_article equals
  let income equals (if original_person.income > $1200 then $1200 else original_person.income) in
  Person {
    -- income: income
    -- age: original_person.age
    -- xx: original_person.xx
    -- yy: original_person.yy
    etc.
  }
```

This is tedious to write, verbose, and coud introduce hard-to-spot errors. We propose a shorter syntax that would be equivalent to the code above:
```catala
definition person_for_this_article equals
  let income equals (if original_person.income > $1200 then $1200 else original_person.income) in
  Person { original_person amended with
    -- income: income
  }
```

- It should be absolutely clear to the reader that `Person { original_person amended with ... }` is a new, distinct structure and that `original_person` is not altered in any way.
- is the `amended with` keyword ok ? I am also unsure about the french equivalent (does "amendé par" work ?). Using just `with` (without `amended`), if it's as clear, would avoid the addition of a new keyword.
- feel free to suggest other syntax options

AltGr:
Closes #592

A new node is added in `desugared`, and translated into an exploded
structure literal during translation to `scopelang`. The main reason to put
it there is that it needs to be after disambiguation, since that is used to
discover the type of the structure that is being updated.


denismerigoux (approved):
It's perfect!

PR #605, opened 2024-04-13
Improvements around external modules and error handling

AltGr:
This allows external modules to access all modules (see first patch), protects
the interpreter for better reporting of exceptions raised from modules (third
patch), and register exception printers in the OCaml runtime.

Quoting here the message from the second patch because it mentions some
important points, that may warrant discussion:

OCaml runtime: register fallback exception printers

This was a pending TODO: now the Catala program compiled into OCaml should
return better messages and a little more information about uncaught
exceptions.

Note that this also concerns, at the moment, compiled modules called from
the Catala interpreter: in this case, it's already better than nothing, but
what we need is proper interoperation between the runtime exceptions and
the interpreter handling (`EmptyError` should already be handled properly
since it is critical to the computation flow, but "error" exceptions are
left uncaught and will kill the interpreter).

This may be part of a bigger task on unifying the output of the runtime and
toplevel, which also concerns computation traces.

Note 2: All runtime exceptions don't have a position available, which is
quite unfortunate when your program hits an error. With `OCAMLRUNPARAM=b`
and if compiled with `-g` (which should normally be the case), you can get
an OCaml backtrace but that's not very friendly. Ideas for improvement:

- The runtime could force-enable backtrace recording (`Printexc.record_backtrace
  true`) to supersede the need for `OCAMLRUNPARAM`. We can also record our own
  handler to print the file position and/or backtrace in the way we see fit

- The printer of OCaml code in Catala could insert line directives so that the
  positions in the backtrace actually trace automatically back to the Catala
  code

- If we don't want to leverage any OCaml machinery in this way, the compiler
  should add position information to any operator that might fail (e.g.
  divisions, date comparisons, etc.).

Note that running in trace mode might already help pinpoint the location of the
error ?

### DIFF
--- a/compiler/shared_ast/interpreter.ml
+++ b/compiler/shared_ast/interpreter.ml
@@ -129,7 +129,7 @@ let rec evaluate_operator
       | _ -> assert false
     in
     try f x y with
-    | Division_by_zero ->
+    | Runtime.Division_by_zero ->
       Message.error
         ~extra_pos:
           [
--- a/compiler/shared_ast/interpreter.ml
+++ b/compiler/shared_ast/interpreter.ml
@@ -663,16 +663,21 @@ let rec evaluate_expr :
         Message.error ~pos "wrong function call, expected %d arguments, got %d"
           (Bindlib.mbinder_arity binder)
           (List.length args)
-    | ECustom { obj; targs; tret } ->
+    | ECustom { obj; targs; tret } -> (
       (* Applies the arguments one by one to the curried form *)
-      List.fold_left2
-        (fun fobj targ arg ->
-          (Obj.obj fobj : Obj.t -> Obj.t)
-            (val_to_runtime (fun ctx -> evaluate_expr ctx lang) ctx targ arg))
-        obj targs args
-      |> Obj.obj
-      |> fun o ->
-      runtime_to_val (fun ctx -> evaluate_expr ctx lang) ctx m tret o
+      match
+        List.fold_left2
+          (fun fobj targ arg ->
+            (Obj.obj fobj : Obj.t -> Obj.t)
+              (val_to_runtime (fun ctx -> evaluate_expr ctx lang) ctx targ arg))
+          obj targs args
+      with
+      | exception e ->
+        Format.ksprintf
+          (fun s -> raise (CatalaException (Crash s, pos)))
+          "@[<hv 2>This call to code from a module failed with:@ %s@]"
+          (Printexc.to_string e)
+      | o -> runtime_to_val (fun ctx -> evaluate_expr ctx lang) ctx m tret o)
     | _ ->
       Message.error ~pos "%a" Format.pp_print_text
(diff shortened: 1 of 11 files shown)


runtimes/ocaml/runtime.ml
```diff
@@ -53,9 +53,34 @@ exception IndivisibleDurations
 exception ImpossibleDate
 exception NoValueProvided of source_position
 exception NotSameLength
+exception Division_by_zero (* Shadows the stdlib definition *)
 
-(* TODO: register exception printers for the above
-   (Printexc.register_printer) *)
+(* Register exceptions printers *)
```
denismerigoux, line 58: OK so you're going through OCaml's exception printing mechanism, which is more general. Why go through that very general mechanism that forces you to print source code positions in a crude way when you could catch in `interpreter.ml` the exceptions of `Runtime_ocaml` properly and access to their positions in a structured way (and that you can print beautifully)?

runtimes/ocaml/runtime.mli
```diff
@@ -76,6 +76,7 @@ exception UncomparableDurations
 exception IndivisibleDurations
 exception ImpossibleDate
 exception NoValueProvided of source_position
+exception Division_by_zero (* Shadows the stdlib definition *)
```
denismerigoux, line 79: Yes.

denismerigoux (approved):
Thanks very much @AltGr! Nonetheless I still don't understand something, see my comments.

compiler/shared_ast/interpreter.ml
```diff
@@ -663,16 +663,21 @@ let rec evaluate_expr :
...
+          (fun fobj targ arg ->
+            (Obj.obj fobj : Obj.t -> Obj.t)
+              (val_to_runtime (fun ctx -> evaluate_expr ctx lang) ctx targ arg))
+          obj targs args
+      with
+      | exception e ->
+        Format.ksprintf
+          (fun s -> raise (CatalaException (Crash s, pos)))
+          "@[<hv 2>This call to code from a module failed with:@ %s@]"
+          (Printexc.to_string e)
```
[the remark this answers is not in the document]
AltGr: Well, we call `Printexc.to_string` which will resort to the `NoValueProvided` printer registered [here](https://github.com/CatalaLang/catala/pull/605/files#diff-8df1fdb118655f12227fe1088f206589bc4955c39a3e84d1bdf9e089d55a76f5R73-R74) , so the position will be available in the message and we don't gain much with special catching here.

runtimes/ocaml/runtime.ml
AltGr: At the moment, the position types in the runtime are different from the ones in the compiler, so we would need specific printing functions ; the interpreter doesn't seem like the right place to put those ?

Ideally, like for backtraces, we'd like unified printing for interpreter and runtime errors ; it would probably be best to start with unification of the underlying types. That's ongoing work but I wanted a quick fix: having a message with a usable position was the priority here.

PR #607, opened 2024-04-19
Add env variables to set flags `-O`, `--avoid-exceptions`, `--trace`

AltGr:
It makes it easier to batch-set them when testing a given repo with different
sets of options. Direct flags should still be preferred in general, of course.

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -208,7 +208,7 @@ module Cli = struct
     in
     let makeflags_to_ninja_flags (makeflags : string option) =
       match makeflags with
-      | None -> ""
+      | None -> ["-k0"]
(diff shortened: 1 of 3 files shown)


AltGr:
It may be wise for `clerk test` to do something about those, to avoid possible confusion:
- either clean them up entirely
- or fail if they are set
- or go on, but print a warning/hint

AltGr:
Update: cleaning up `CATALA_*` (except what's explicitely handled) with a warning. `clerk runtest` cleans them up silently (that last part is disputable as, normally, the command is run from `clerk test` and the cleanup should already be done)

denismerigoux (approved):
OK, thanks @AltGr !

PR #610, opened 2024-04-24
Concolic testing engine

rmonat:
This contains the work of @pierregoutagny done during his internship. I spent a few hours merging with master, I'm opening a PR to track upcoming breaking changes.


compiler/shared_ast/expr.ml
```diff
@@ -798,6 +804,7 @@ let rec compare : type a. (a, _) gexpr -> (a, _) gexpr -> int =
   | EPureDefault _, _ -> -1 | _, EPureDefault _ -> 1
   | EEmptyError , _ -> -1 | _, EEmptyError  -> 1
   | EErrorOnEmpty _, _ -> -1 | _, EErrorOnEmpty _ -> 1
+  | EGenericError , _ -> -1 | _, EGenericError  -> 1
```
AltGr, line 807: ⚠️ `EGenericError, EGenericError` is missing !!

compiler/concolic/interpreter.ml
```diff
@@ -0,0 +1,2776 @@
...
+      Expr.eappop ~tys ~args:(List.map f args) ~op:(Operator.translate op) m
+    | ( ( EAssert _ | ELit _ | EApp _ | EArray _ | EVar _ | EExternal _ | EAbs _
+        | EIfThenElse _ | ETuple _ | ETupleAccess _ | EInj _ | EStruct _
+        | EStructAccess _ | EMatch _ | EDefault _ | EPureDefault _ | EEmptyError
+        | EErrorOnEmpty _ | ECustom _ ),
+        _ ) as e ->
+      Expr.map ~f e
+    | _ -> .
+  in
+  (* /!\ don't be tempted to use the same trick here, the function does one
```
AltGr, line 206: What trick ? ;)

Maybe remove that comment

compiler/shared_ast/interpreter.mli
```diff
@@ -59,6 +61,12 @@ val interpret_program_lcalc :
     providing for each argument a thunked empty default. Returns a list of all
     the computed values for the scope variables of the executed scope. *)
 
+val addcustom :
+  (('a, 'b, 'c) interpr_kind, 't) gexpr ->
+  (('a, 'b, yes) interpr_kind, 't) gexpr
+(** Typing shenanigan to add custom terms to the AST type. This is an identity
```
AltGr, line 67: This comment is good for the ml file, but the mli doc should just state that it's a type conversion

rmonat:
Thank you for your comments @AltGr! I'll see what I can process and let Pierre handle the rest when he's back on the project.

AltGr:
Ah I see you've added `if Global.options.debug then Message.debug <xxx>;` everywhere... I am curious with what you observed about the costs: without those,
- when `%a` is used the cost of the corresponding string conversion won't be paid when debug is off
- you still always pay the costs for `%s`, obviously. I can see in particular a `Solver.string_of_model` which *looks* coslty
- I think you will also always pay the cost of parsing the format-string (for `%` chars), even with debug off, because that is needed to know the arity of the function to apply (even if they are going to be ignored by `ifprintf`)

Because of the last item indeed there should be a gain even if the debug is a pure string... but I wouldn't have expected that to be significant ?

>  ` (* TODO QU RAPHAEL: turns out del_genericerror is pretty costly (~2x), so I decided to remove it because I think it is a safety/debugging check now *)`

Yes, it will require a whole traversal and reconstruction of the term :/
Indeed, if you don't mind having the `GenericError` case still existing in your result type, you shouldn't bother. In the case of the "normal" interpreter it's not really a problem because the returned value is'nt expected to be so big (although at some point it got called on an intermediate closure and the cost was horrible).
Also be careful with timings and bindlib, anything done with `Bindlib.box_apply` and consorts is actually delayed until you unbind/unbox.

AltGr:
Ah my bad, after re-examining the code we *don't* actually use `ifprintf` but only a void `ppf` in debug mode, which means that indeed the full price of conversion is always paid. That will be easy to work around !

PR #611, opened 2024-04-26
Unify runtime error handling

AltGr:
- Clearly distinguish Exceptions from Errors. The only catchable exception
 available in our AST is `EmptyError`, so the corresponding nodes are made less
 generic, and a node `FatalError` is added

- Runtime errors are defined as a specific type in the OCaml runtime, with a
 carrier exception and printing functions. These are used throughout, and
 consistently by the interpreter. They always carry a position, that can be
 converted to be printed with the fancy compiler location printer, or in a
 simpler way from the backends.

- All operators that might be subject to an error take a position as argument,
 in order to print an informative message without relying on backtraces from
 the backend

BONUS: this PR also includes a tweak to have message adapt to the terminal
width, and fixes to the cheat-sheets.

### DIFF
--- a/compiler/catala_utils/file.ml
+++ b/compiler/catala_utils/file.ml
@@ -180,6 +180,42 @@ let process_out ?check_exit cmd args =
     assert false
   with End_of_file -> Buffer.contents buf
 
+(* SIDE EFFECT AT MODULE LOAD: sets up a signal handler on SIGWINCH (window
+   resize) *)
+let () =
+  let default = 80 in
+  let get_terminal_cols () =
+    let count =
+      try (* terminfo *)
+          process_out "tput" ["cols"] |> int_of_string
+      with Failure _ -> (
+        try
+          (* stty *)
+          process_out "stty" ["size"]
+          |> fun s ->
+          let i = String.rindex s ' ' + 1 in
+          String.sub s (i + 1) (String.length s - i) |> int_of_string
+        with Failure _ | Not_found | Invalid_argument _ -> (
+          try int_of_string (Sys.getenv "COLUMNS")
+          with Not_found | Failure _ -> 0))
+    in
+    if count > 0 then count else default
+  in
+  let width = ref None in
+  let () =
+    try
+      Sys.set_signal 28 (* SIGWINCH *)
+        (Sys.Signal_handle (fun _ -> width := None))
+    with Invalid_argument _ -> ()
+  in
+  Message.set_terminal_width_function (fun () ->
+      match !width with
+      | Some n -> n
+      | None ->
+        let r = get_terminal_cols () in
+        width := Some r;
+        r)
+
 let check_directory d =
   try
     let d = Unix.realpath d in
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -375,7 +375,7 @@ let rec translate_expr
           (try
              Runtime.date_of_numbers date.literal_date_year
                date.literal_date_month date.literal_date_day
-           with Dates_calc.Dates.InvalidDate ->
+           with Failure _ ->
              Message.error ~pos
                "There is an error in this date, it does not correspond to a \
                 correct calendar day")
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -42,7 +42,7 @@ let translate_binop :
     Ast.expr boxed =
  fun (op, op_pos) pos lhs rhs ->
   let op_expr op tys =
-    Expr.eappop ~op
+    Expr.eappop ~op:(op, op_pos)
       ~tys:(List.map (Mark.add op_pos) tys)
       ~args:[lhs; rhs]
       (Untyped { pos })
@@ -114,7 +114,10 @@ let translate_binop :
 
 let translate_unop ((op, op_pos) : S.unop Mark.pos) pos arg : Ast.expr boxed =
   let op_expr op ty =
-    Expr.eappop ~op ~tys:[Mark.add op_pos ty] ~args:[arg] (Untyped { pos })
+    Expr.eappop ~op:(op, op_pos)
+      ~tys:[Mark.add op_pos ty]
+      ~args:[arg]
+      (Untyped { pos })
   in
   match op with
   | S.Not -> op_expr Not (TLit TBool)
@@ -238,12 +241,12 @@ let rec translate_expr
   let rec_helper ?(local_vars = local_vars) e =
     translate_expr scope inside_definition_of ctxt local_vars e
   in
-  let rec detuplify_list names = function
+  let rec detuplify_list opos names = function
     (* Where a list is expected (e.g. after [among]), as syntactic sugar, if a
        tuple is found instead we transpose it into a list of tuples *)
     | S.Tuple ls, pos ->
       let m = Untyped { pos } in
-      let ls = List.map (detuplify_list []) ls in
+      let ls = List.map (detuplify_list opos []) ls in
       let rec zip names = function
         | [] -> assert false
         | [l] -> l
@@ -272,7 +275,7 @@ let rec translate_expr
               (Expr.make_tuple (Expr.evar x1 m :: explode (Expr.evar x2 m)) m)
               tys pos
           in
-          Expr.eappop ~op:Map2 ~args:[f_join; l1; rhs]
+          Expr.eappop ~op:(Map2, opos) ~args:[f_join; l1; rhs]
             ~tys:((TAny, pos) :: List.map (fun ty -> TArray ty, pos) tys)
             m
       in
@@ -286,7 +289,7 @@ let rec translate_expr
   match Mark.remove expr with
   | Paren e -> rec_helper e
   | Binop
-      ( (S.And, _pos_op),
+      ( (S.And, pos_op),
         ( TestMatchCase (e1_sub, ((constructors, Some binding), pos_pattern)),
           _pos_e1 ),
         e2 ) ->
@@ -302,14 +305,14 @@ let rec translate_expr
             let nop_var = Var.make "_" in
             Expr.make_abs [| nop_var |]
               (Expr.elit (LBool false) emark)
-              [tau] pos
+              [tau] pos_op
           else
             let binding_var = Var.make (Mark.remove binding) in
             let local_vars =
               Ident.Map.add (Mark.remove binding) binding_var local_vars
             in
             let e2 = rec_helper ~local_vars e2 in
-            Expr.make_abs [| binding_var |] e2 [tau] pos)
+            Expr.make_abs [| binding_var |] e2 [tau] pos_op)
         (EnumName.Map.find enum_uid ctxt.enums)
     in
     Expr.ematch ~e:(rec_helper e1_sub) ~name:enum_uid ~cases emark
@@ -493,7 +496,7 @@ let rec translate_expr
     in
     Expr.edstructaccess ~e ~field:(Mark.remove x) ~name_opt:(get_str ctxt path)
       emark
-  | FunCall ((Builtin b, _), [arg]) ->
+  | FunCall ((Builtin b, pos), [arg]) ->
     let op, ty =
       match b with
       | S.ToDecimal -> Op.ToRat, TAny
@@ -506,7 +509,7 @@ let rec translate_expr
       | S.FirstDayOfMonth -> Op.FirstDayOfMonth, TLit TDate
       | S.LastDayOfMonth -> Op.LastDayOfMonth, TLit TDate
     in
-    Expr.eappop ~op ~tys:[ty, pos] ~args:[rec_helper arg] emark
+    Expr.eappop ~op:(op, pos) ~tys:[ty, pos] ~args:[rec_helper arg] emark
   | S.Builtin _ ->
     Message.error ~pos "Invalid use of built-in: needs one operand"
   | FunCall (f, args) ->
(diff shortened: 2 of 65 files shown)


AltGr:
Added 4 patches dedicated to actually having better positions, before we need to report them
- the first may make the diff quite bigger, but it's just an addition of position marks to the operators in the `EAppOp` AST node
- some positions were otherwise lost in translation to lcalc, which took a bit to debug (I first suspected optimisations) ; all tests that used explicit `interpret` instead of `test-scope` because of this are now fixed and positions in error reporting are now consistent with the different backends (this is verified by `make testsuite`)

compiler/catala_utils/file.ml
```diff
@@ -180,6 +180,42 @@ let process_out ?check_exit cmd args =
     assert false
   with End_of_file -> Buffer.contents buf
 
+(* SIDE EFFECT AT MODULE LOAD: sets up a signal handler on SIGWINCH (window
```
denismerigoux, line 183: This is very fancy

compiler/desugared/from_surface.ml
```diff
@@ -366,7 +375,7 @@ let rec translate_expr
           (try
              Runtime.date_of_numbers date.literal_date_year
                date.literal_date_month date.literal_date_day
-           with Runtime.ImpossibleDate ->
+           with Dates_calc.Dates.InvalidDate ->
```
denismerigoux, line 378: In my mind the Catala compiler would never directly depend on `Dates_calc`, to ensure some sort of abstraction of how things like dates are actually implemented. Which is why the runtime re-exported this error and why we used it here. Do you have a good rationale for tweaking this?

compiler/scalc/ast.ml
```diff
@@ -61,16 +61,17 @@ and naked_expr =
...
   | EExternal of { modname : VarName.t Mark.pos; name : string Mark.pos }
 
 type stmt =
   | SInnerFuncDef of { name : VarName.t Mark.pos; func : func }
   | SLocalDecl of { name : VarName.t Mark.pos; typ : typ }
   | SLocalInit of { name : VarName.t Mark.pos; typ : typ; expr : expr }
   | SLocalDef of { name : VarName.t Mark.pos; expr : expr; typ : typ }
-  | STryExcept of { try_block : block; except : except; with_block : block }
-  | SRaise of except
+  | STryWEmpty of { try_block : block; with_block : block }
```
denismerigoux, line 72: Good

compiler/shared_ast/interpreter.ml
```diff
@@ -777,20 +750,21 @@ let rec evaluate_expr :
     match Mark.remove e with
     | ELit (LBool true) -> Mark.add m (ELit LUnit)
     | ELit (LBool false) ->
-      Message.error ~pos:(Expr.pos e') "Assertion failed:@\n%a"
+      Message.warning "Assertion failed:@ %a"
         (Print.UserFacing.expr lang)
         (partially_evaluate_expr_for_assertion_failure_message ctx lang
-           (Expr.skip_wrappers e'))
+           (Expr.skip_wrappers e'));
```
denismerigoux, line 756: You raise a warning and then you raise a runtime error ?

compiler/shared_ast/typing.ml
```diff
@@ -354,13 +354,11 @@ let polymorphic_op_return_type
 let resolve_overload_ret_type
     ~flags
     (ctx : A.decl_ctx)
-    e
-    (op : Operator.overloaded A.operator)
+    _e
```
denismerigoux, line 357: Unused argument that can be deleted?

runtimes/ocaml/runtime.ml
```diff
@@ -45,42 +45,60 @@ type source_position = {
...
+    Some (Printf.sprintf "At %a: %s" pposl pos (error_message err))
   | _ -> None
 
 let () =
   Printexc.set_uncaught_exception_handler
   @@ fun exc bt ->
   Printf.eprintf "[ERROR] %s\n%!" (Printexc.to_string exc);
   if Printexc.backtrace_status () then Printexc.print_raw_backtrace stderr bt
+(* TODO: the backtrace will point to the OCaml code; but we could make it point
+   to the Catala code if we add #line directives everywhere in the generated
```
denismerigoux, line 100: We could add those in the OCaml backend with the `-d` flag.

runtimes/ocaml/runtime.ml
```diff
@@ -703,61 +722,60 @@ end
...
-        | Some _, None -> acc
-        | Some _, Some _ -> raise (ConflictError pos))
-      None exceptions
+  let len = Array.length exceptions in
+  let rec filt_except i =
+    if i < len then
+      match exceptions.(i) () with
+      | new_val -> (new_val, i) :: filt_except (i + 1)
+      | exception Empty -> filt_except (i + 1)
+    else []
```
denismerigoux, line 737: Did you rewrite into a tail-call form to get more performance?

runtimes/ocaml/runtime.ml
```diff
@@ -703,61 +722,60 @@ end
...
 (* TODO: add a compare built-in to dates_calc. At the moment this fails on e.g.
    [3 months, 4 months] *)
-let compare_periods (p1 : duration) (p2 : duration) : int =
+let compare_periods pos (p1 : duration) (p2 : duration) : int =
   try
     let p1_days = Dates_calc.Dates.period_to_days p1 in
     let p2_days = Dates_calc.Dates.period_to_days p2 in
     compare p1_days p2_days
-  with Dates_calc.Dates.AmbiguousComputation -> raise UncomparableDurations
+  with Dates_calc.Dates.AmbiguousComputation ->
```
denismerigoux, line 769: Re the comment I made earlier : this raises a runtime exception, not a `Dates_calc` exception, so why catch a `Dates_calc` exception ?

tests/arithmetic/bad/division_by_zero.catala_en
```diff
@@ -32,90 +32,39 @@ scope Money:
 
 
 ```catala-test-inline
-$ catala Interpret -s Dec
-[ERROR] division by zero at runtime
+$ catala test-scope Dec
```
denismerigoux, line 35: Could you do a PR in https://gitlab.adullact.net/dgfip/ir-catala to make sure we use `test-scope` everywhere and no longer `Interpret -s` anymore? Can you remind me the difference between the two again?

tests/arithmetic/bad/division_by_zero.catala_en
denismerigoux, line 1: The error message is slightly less expressive now. Maybe `DivisionByZero` could carry two positions instead of one in the accompanying position list, the second pointing specifically to the null denominator?

tests/date/bad/uncomparable_duration.catala_en
```diff
@@ -42,83 +42,55 @@ scope Ge:
...
+[ERROR] Error during evaluation: comparing durations in different units (e.g.
+  months vs. days).
 
-┌─⯈ tests/date/bad/uncomparable_duration.catala_en:40.23-40.30:
+┌─⯈ tests/date/bad/uncomparable_duration.catala_en:40.31-40.33:
 └──┐
 40 │   definition d equals 1 month >= 2 day
-   │                       ‾‾‾‾‾‾‾
-   └┬ `UncomparableDurations` exception management
-    └─ `>=` operator
```
denismerigoux, line 53: So the runtime exceptions do not carry a law heading breadcrumb anymore. Maybe we should add law heading breadcrumbs to `Runtime.position`?

tests/default/bad/empty.catala_en
```diff
@@ -19,8 +19,7 @@ $ catala test-scope A
 6 │   output y content boolean
   │          ‾
   └─ Article
-[ERROR] This variable evaluated to an empty term (no rule that defined it
-  applied in this situation)
+[ERROR] Error during evaluation: no computation with valid conditions found.
```
denismerigoux, line 22: The new error message is more cryptic than the old one, plus we should really be explicit and learning-friendly with this precise error message as this is one that newcomers are going to encounter a lot and early.

tests/func/bad/bad_func.catala_en
```diff
@@ -27,21 +27,16 @@ $ catala test-scope R
...
 ```
 
-Note: ideally this could use test-scope but some positions are lost during translation to lcalc
-
 ```catala-test-inline
-$ catala interpret -s S
-[ERROR] There is a conflict between multiple valid consequences for assigning
-  the same variable.
+$ catala test-scope S
+[ERROR] Error during evaluation: two or more concurring valid computations.
```
denismerigoux, line 32: This error message should be more pedagogical ; why not leave the exact same text as the old message? It seems like it would work.

denismerigoux (approved):
Thanks you @AltGr, this is a great improvement ! Hopefully it paves the way to enabling tracing in OCaml-compiled modules so that we can have a full trace across compiled modules in the interpreter.

compiler/desugared/from_surface.ml
AltGr: Ok, no indeed you have a good point, I'll ensure it's properly shadowed

compiler/shared_ast/interpreter.ml
AltGr: Ah yes, this one is a bit tricky: here we indeed want to benefit from the greater powers we have in the interpreter to print a more detailed message, and arbitration was needed between that and the unification of error handling between runtime and interpreter.

The choice I made was to print the more detailed message as a side-effect, then raise the exception to be handled in the common path ; but indeed that will make the console show `[WARNING]` with the detail and then `[ERROR]` with the position and the message. The alternatives, either specific error handling of this exception at this point, or somehow embedding the added information into the runtime exception, didn't seem worth it.

Maybe this would be better if we add a non-raising variant for `Message.error` and use that here, though ?

runtimes/ocaml/runtime.ml
AltGr: Yes, I'll experiment with that a bit more: it's not perfect (granularity is the OCaml line of code; lots of backtrace items point to the runtime or libraries) and will make the OCaml code less readable but that can definitely be pretty useful for debugging.

compiler/catala_utils/file.ml
AltGr: Well I had the code lying around, so... 🤷🏿

runtimes/ocaml/runtime.ml
AltGr: Hm, the main reason for rewriting was to be able to report conflicts between more than 2 exceptions (here again, to match the reporting done in the interpreter).

The new version is actually not tail-recursive, while the original was (using `Array.fold`). Why ?
- There is no `Array.fold2` which would have been the simplest choice
- We have no recursion in Catala, so there is no risk that the number of exceptions gets anywhere near the stack limit
- The non-tail recursive version is more efficient

In addition, if we're still worried about stack size, since we already require OCaml 4.14, we could annotate with `[@ocaml.tail_mod_cons]` and get both bounded stack space and efficiency.

tests/arithmetic/bad/division_by_zero.catala_en
AltGr: Ah indeed — my first version only carried a single position for runtime error, but I had to extend it for acceptable `Conflict` exceptions; I didn't re-add the positions for operators though, at the moment it seemed superfluous.

Note, however, that we are now able to point precisely at the operator that failed, while before we had to point to the whole expression, which could for example include several divisions. But indeed, priority of operators could lead to things like `1 / 0 + 1` where underlining the `/` is not as explicit as underlining the `0`.

An issue is that runtime exceptions can have multiple positions, but no associated messages at the moment.

EDIT: after discussion, pointing *only* to the denominator seems to actualy be the best compromise.

tests/arithmetic/bad/division_by_zero.catala_en
AltGr: Yes!

Basically the purpose is to be able to run the test with various combination of interpreter options (dcalc/lcalc, exceptions, etc.), so the special syntax is purposefully less explicit than the real catala command `interpret`.

See the `make testuite` target; quoting `clerk test --help`:

        --test-flags=FLAGS [...]
           Flags to pass to the catala interpreter on catala test-scope tests.
           Comma-separated list. A subset may also be applied to the
           compilation of modules, as needed. [...] NOTE: if this is set, all inline
           tests that are not `catala test-scope` are skipped to avoid redundant
           testing.

tests/date/bad/uncomparable_duration.catala_en
AltGr: They do, they do, it's already there.

compiler/shared_ast/typing.ml
AltGr: Ah indeed, it was used for position but now the operator includes that

runtimes/ocaml/runtime.ml
AltGr: Exactly, here we are in the runtime so we catch the exception from `Dates_calc` to turn it into a proper `Runtime` exception ; the function in the earlier example actually had the comment:

```ocaml
(* This could fail, but is expected to only be called with known, already
   validated arguments by the generated code *)
```

since it would be called by the compiler, but is outside of what is expected to fail in the runtime (the compiled code *can* call it, but only on values already checked at compile-time)
I guess what led me to this was the intent to avoid catching `Error`s from the runtime.

AltGr: (in other words and to clarify my thoughts: `Runtime.date_of_numbers` failing _cannot_ be a runtime error, so it shouldn't belong to the type `Runtime.errors`)

AltGr:
All should be addressed with the 2 added patches

PR #615, opened 2024-05-03
Unify traces between interpreter and runtime

AltGr:
This is a first step into unifying trace handling. This patch only affects the interpreter, by delegating trace recording to the already existing runtime functions.

At end of interpretation, it recovers the registered trace from the runtime, and prints it.

NOTE: there are some limitations due to this approach, as runtime values going through this interface have to be converted to the "runtime embedded" type. In particular, functions can no longer be printed (which makes full sense if we want it to happen in the same way in compiled code) ; some information, like types, is lost, but it didn't appear to be used.

Also, a specific printer had to be added for runtime values (but it's very simple so that shouldn't be a problem).

@denismerigoux I'd like your input on how well this goes for your use-cases.

Further work should probably be cleanup and unification of the runtime logging interfaces ; there is already code for re-structuring the traces, printing to JSON, etc. which could be common to runtime and interpreter.

### DIFF
--- a/compiler/shared_ast/interpreter.ml
+++ b/compiler/shared_ast/interpreter.ml
@@ -60,7 +60,7 @@ let rec format_runtime_value lang ppf = function
          ~pp_sep:(fun ppf () -> Format.fprintf ppf ";@ ")
          (format_runtime_value lang))
       (Array.to_list elts)
-  | Runtime.Unembeddable -> Format.pp_print_string ppf "<?>"
+  | Runtime.Unembeddable -> Format.pp_print_string ppf "<object>"
 
 let print_log lang entry =
   let pp_infos =
--- a/compiler/shared_ast/interpreter.ml
+++ b/compiler/shared_ast/interpreter.ml
@@ -34,30 +34,88 @@ let is_empty_error : type a. (a, 'm) gexpr -> bool =
 let indent_str = ref ""
 
 (** {1 Evaluation} *)
-let print_log lang entry infos pos e =
-  if Global.options.trace then
-    match entry with
-    | VarDef _ ->
-      Message.log "%s%a %a: @{<green>%s@}" !indent_str Print.log_entry entry
-        Print.uid_list infos
-        (Message.unformat (fun ppf ->
-             (if Global.options.debug then Print.expr ~debug:true ()
-              else Print.UserFacing.expr lang)
-               ppf e))
-    | PosRecordIfTrueBool -> (
-      match pos <> Pos.no_pos, Mark.remove e with
-      | true, ELit (LBool true) ->
-        Message.log "%s@[<v>%a@{<green>Definition applied@}:@,%a@]" !indent_str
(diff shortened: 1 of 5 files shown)


AltGr:
NOTE: rebase before merge

compiler/shared_ast/interpreter.ml
```diff
@@ -34,30 +34,88 @@ let is_empty_error : type a. (a, 'm) gexpr -> bool =
...
+    Format.fprintf ppf "@[<hov 2>%s.%s@ (%a)@]" name constr
+      (format_runtime_value lang)
+      v
+  | Runtime.Struct (name, fields) ->
+    Format.fprintf ppf "@[<hv 2>%s {@ %a@;<1 -2>}@]" name
+      (Format.pp_print_list ~pp_sep:Format.pp_print_space (fun ppf (fld, v) ->
+           Format.fprintf ppf "@[<hov 2>-- %s:@ %a@]" fld
+             (format_runtime_value lang)
+             v))
+      fields
```
denismerigoux, line 56: Can't you reuse the code here ?

https://github.com/CatalaLang/catala/blob/59b0f21ff8ff27cd445ced0ab70923cb47b08165/compiler/shared_ast/print.ml#L111-L131

You could expose a `Print.UserFacing.struct` function in `print.mli`.

compiler/shared_ast/interpreter.ml
```diff
@@ -34,30 +34,88 @@ let is_empty_error : type a. (a, 'm) gexpr -> bool =
...
+  | Runtime.Array elts ->
+    Format.fprintf ppf "@[<hv 2>[@,@[<hov>%a@]@;<0 -2>]@]"
+      (Format.pp_print_list
+         ~pp_sep:(fun ppf () -> Format.fprintf ppf ";@ ")
+         (format_runtime_value lang))
+      (Array.to_list elts)
```
denismerigoux, line 62: Same comment as above.

compiler/shared_ast/interpreter.ml
```diff
@@ -34,30 +34,88 @@ let is_empty_error : type a. (a, 'm) gexpr -> bool =
...
+  | Runtime.Unembeddable -> Format.pp_print_string ppf "<?>"
```
denismerigoux, line 63: If a user sees `<?>` in their trace logs, they will be very confused. A better print would be `<object>`, mirroring what e.g. Python does.

denismerigoux (approved):
Thanks @AltGr ! I like where this is going. Of course we'll need to enrich the `embedded_value` type with tuples. If we want to add further information like types in traces later we'll enrich `embdedded_value` in the same way. 

For the use cases we have, this is ok !

compiler/shared_ast/interpreter.ml
AltGr: Hm, while I agree that the formatting should be similar, the runtime values use different types (strings and lists _vs_ uids and maps) so the conversion work would be as big as the printing function...

PR #618, opened 2024-05-10
Fixes in Python runtime

### LINKED ISSUE #612: Python rounding operators are incorrect
Following #557

The [rounding test](https://github.com/CatalaLang/catala/blob/master/tests/arithmetic/good/rounding.catala_en) fails on the Python backend for now.

rmonat:
Closes #612. 

The type annotations change are a personal preference, feel free to revert.

### DIFF
--- a/runtimes/python/src/catala/runtime.py
+++ b/runtimes/python/src/catala/runtime.py
@@ -6,6 +6,7 @@
 
 .. moduleauthor:: Denis Merigoux <denis.merigoux@inria.fr>
 """
+from __future__ import annotations # 'ClsType' ~> ClsType annotations
 
 # This file should be in sync with compiler/runtime.{ml, mli} !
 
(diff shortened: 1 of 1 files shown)


AltGr:
Looks fine to me, but @denismerigoux knows the Python backend in more detail than I do

runtimes/python/src/catala/runtime.py
```diff
@@ -78,34 +79,37 @@ def __repr__(self) -> str:
 
 
 class Decimal:
-    def __init__(self, value: Union[str, int, float]) -> None:
-        self.value = mpq(value)
+    def __init__(self, value: Union[str, int, float,Integer]) -> None:
+        if isinstance(value, Integer):
+            self.value = mpq(value.value)
```
denismerigoux, line 84: Good catch

runtimes/python/src/catala/runtime.py
```diff
@@ -416,6 +411,14 @@ def __init__(self, source_position: SourcePosition) -> None:
 # Constructors and conversions
 # ============================
 
+def round(q : Decimal) -> Integer:
```
denismerigoux, line 414: Maybe you could call this function `round_decimal_to_integer`, to not confuse it with `decimal_round`

rmonat: The name is the same in the ocaml runtime, what shall we do?

denismerigoux: Oh indeed, I let that slip in a previous PR :) You can leave it as `round` I guess.

Issue #623, opened 2024-05-21
Improve the output of clerk test


denismerigoux:
As of now, the typical output of `clerk test <folder<` is : 

```text
[27/27] <test> 'tests@test'

[PASS] tests:   5/5
```

Where `5` is the number of files containing one or more tests found inside <folder>. If there is a test failure, then what is shown is: 

```text
[21/27] <test> tests/benefices_non_commerciaux.catala_fr
--- reference
+++ current-output
@@ -159,7 +159,7 @@
 ┌─[RESULT]─
 │ sortie =
 │   Impot_revenu.BénéficesNonCommerciauxFoyerFiscal {
-│     -- résultats_liquidatin_bénéfices_non_commerciaux:
+│     -- résultats_liquidation_bénéfices_non_commerciaux:
 │       [
 │         Impot_revenu.BénéficesNonCommerciauxDéclarant {
 │           -- abattement_forfaitaire_micro_professionnel: 0,00 €
[27/27] <test> 'tests@test'
FAILED: tests@test 
out='tests@test' ; success=$( tr -cd 0 < '_build/tests@test' | wc -c ) ; total=$( wc -c < '_build/tests@test' ) ; pass=$( ) ; if test "$success" -eq "$total" ; then printf "\n[\033[32mPASS\033[m] \033[1m%s\033[m: \033[32m%3d\033[m/\033[32m%d\033[m\n" ${out%@test} $success $total ; else printf "\n[\033[31mFAIL\033[m] \033[1m%s\033[m: \033[31m%3d\033[m/\033[32m%d\033[m\n" ${out%@test} $success $total ; return 1 ; fi

[FAIL] tests:   4/5
````

However, because this command is the primary testing method we recommend for a typical Catala workflow, the output of the command should be improved to look better and provide more accurate information. Here is a list of improvements that could be made : 

* instead of displaying `5/5`, it should display `37/37 tests across 5 files`
* when there is a test failure, the output should be a clean listing of all tests that have failed, grouped by file, and not just the first test that failed.

I suspect the relevant code to tweak is here for these improvements : 

https://github.com/CatalaLang/catala/blob/e7853d69cf1f258142ef6d23a0bdd083d7e2d14e/build_system/clerk_driver.ml#L564-L566

https://github.com/CatalaLang/catala/blob/e7853d69cf1f258142ef6d23a0bdd083d7e2d14e/build_system/clerk_driver.ml#L580-L600

AltGr:
## Better test output

[...]

Issue #482, opened 2023-06-21
Module support (tracking issue)


AltGr:
### Synthetic overview of where module support is currently at

* Module interface: the `catala-metadata` sections of a normal Catala file

* Module name based upon filename at the moment
  - [x] `> Module` directives

* Compiling modules
  - [x] introduce a dynlink registration of compiled catala modules into the OCaml runtime for use by the interpreter
  - [x] add a module hash based on interface + Catala version
  - [x] facility to compile module to `cmxs` directly

* Using module interfaces (in the compiler)
  - [x] declare use using the CLI
  - [x] or a `Using` directive
  - [x] register toplevel values
  - [x] register toplevel scopes
  - [x] register type defs

* Using module implementations (in the interpreter)
  - [x] load the cmxs of the module through dynlink
  - [x] Validate the module hash to ensure consistency of the interface & backend
  - [x] Call the external functions with back-and-forth translation between exprs and runtime values
  - [x] (the above needs more tests. Come on, no segfault yet ??!)

* Using module implementations (OCaml backend)
  - [x] Manually compiled modules can be linked together using `ocamlopt`
  - [x] Some facility to handle the compile command ?

* Using module implementations (other backends)
  - [ ] For js: see how to port the OCaml linking command to jsoo
  - [ ] TODO (later)

* Native modules (aka externals)
  - [x] it's possible to manually provide the .ml file for a module
  - [ ] facilities to extract the expected mli interface from a catala file
  - [x] facilities to correctly compile the external modules

* Replacing the monomorphic operators
  - [ ] define an (auto opened) stdlib module
  - [ ] move the operators from the backend to there and adjust the compiler (TODO)

* Extend all of the above to the other runtimes, providing a suitable stdlib
  - [x] jsoo
  - [x] Python
  - [ ] Others ??

denismerigoux:
Where are you at here @AltGr ?

AltGr:
I updated the checklist :sweat_smile: 

Basically it's now usable (mainly thanks to the rehauled build system) although there still are some limitations:
- lack of documentation
- OCaml support only (I am working on Python atm)
- Clerk imposes that modules and their uses are in the same directory at the moment

Also the stdlib is still a todo, and convenience entry-points are lacking for external modules.

denismerigoux:
Closing this as mostly completed, other items will be subject to specific issues.

PR #625, opened 2024-05-27
Implement module hashes and checks

AltGr:
Related to #482

Hashes are made of 3 parts: the Catala version, the compilation flags (should be only those affecting the call convention), and the module interface

This includes a few changes to make it all work smoothly:
- Proper hash computation on interfaces (in particular, it was tricky to make them independant on the execution-specific uids, and on the surrounding module name, not mentioning map orderings etc.)
- Passing of more information along the compilation chain, like module hashes and wether external, and the visibility of individual declarations
- Adding some post-processing to `Clerk runtest` to erase unique info (module hash) from the output. Anyway this is work that will be required for #623 
- A bypass for external modules, because we don't want to update their hashes manually at every Catala update ; this is temporary and won't be needed once we have a better interface for supplying externals.

### DIFF
--- a/compiler/desugared/ast.ml
+++ b/compiler/desugared/ast.ml
@@ -71,13 +71,17 @@ module ScopeDef = struct
       ScopeVar.format ppf (Mark.remove v);
       format_kind ppf k
 
-    let hash_kind = function
-      | Var None -> Hashtbl.hash `VarNone
-      | Var (Some st) -> Hashtbl.hash (`VarSome (StateName.id st))
-      | SubScopeInput { var_within_origin_scope = v; _ } ->
-        Hashtbl.hash (`SubScopeInput (ScopeVar.id v))
+    open Hash.Op
 
-    let hash (v, k) = Hashtbl.hash (ScopeVar.id (Mark.remove v), hash_kind k)
+    let hash_kind ~strip = function
+      | Var v -> !`Var % Hash.option StateName.hash v
+      | SubScopeInput { name; var_within_origin_scope } ->
+        !`SubScopeInput
+        % ScopeName.hash ~strip name
+        % ScopeVar.hash var_within_origin_scope
+
+    let hash ~strip (v, k) =
+      Hash.Op.(ScopeVar.hash (Mark.remove v) % hash_kind ~strip k)
   end
 
   include Base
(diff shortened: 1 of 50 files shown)


compiler/desugared/ast.ml
```diff
@@ -239,21 +246,76 @@ type scope = {
...
 
 type modul = {
   module_scopes : scope ScopeName.Map.t;
-  module_topdefs : (expr option * typ) TopdefName.Map.t;
+  module_topdefs : topdef TopdefName.Map.t;
 }
 
 type program = {
-  program_module_name : Ident.t Mark.pos option;
+  program_module_name : (ModuleName.t * Hash.t * bool) option;
```
denismerigoux, line 264: What is the `bool` here? Shouldn't it be a record type instead of a tuple here?

compiler/lcalc/to_ocaml.ml
```diff
@@ -743,11 +757,17 @@ let format_module_registration
     (fun fmt (id, (var, _)) ->
       Format.fprintf fmt "@[<hov 2>%S,@ Obj.repr %a@]" id format_var var)
     fmt (String.Map.to_seq bnd);
+  (* TODO: pass the visibility info down from desugared, and filter what is
```
denismerigoux, line 760: Maybe create an issue for this new TODO :) ?

compiler/shared_ast/expr.ml
```diff
@@ -974,6 +974,63 @@ let rec size : type a. (a, 't) gexpr -> int =
...
+ *     let y,m,d = Dates_calc.Dates.date_to_ymd d in
+ *     y lxor m lxor d
+ *   | LDuration dt ->
+ *     let y,m, d = Dates_calc.Dates.period_to_ymds dt in
+ *     y lxor m lxor d
+ * 
+ * let rec full_hash : type a. (a, 't) gexpr -> int =
+ *  fun e ->
+ *   match Mark.remove e with
+ *   | EVar v  -> Bindlib.hash_var v
```
denismerigoux, line 993: Commented code?

compiler/surface/lexer.cppo.ml
```diff
@@ -778,9 +778,6 @@ let lex_raw (lexbuf : lexbuf) : token =
     | _ -> (
         (* Nested match for lower priority; `_` matches length 0 so we effectively retry the
            sub-match at the same point *)
-        let lexbuf = lexbuf in
```
denismerigoux, line 781: Great!

doc/devel/externals.md
```diff
@@ -31,10 +31,10 @@ catala implementation and compile to OCaml (removing the `external` directive):
     ```
 
 ```shell-session
-$ clerk build _build/.../Prorata_external.ml
+$ clerk build _build/.../prorata_external.ml
 ```
 
-(beware the `_build/`, and the capitalisation of the module name)
+(beware the `_build/`, it is required here)
```
denismerigoux, line 37: Excellent ! Can we revert to non-capitalized generated files names in `catala-examples` and `ir-catala` then?

denismerigoux (approved):
Thanks @AltGr !!

AltGr:
Unfortunately, CI is broken due to some Python nonsense (the latest ocamlpro/ocaml image is needed to get `dates_calc=0.0.6` ; but it has bumped Alpine (and musl) version and doesn't yet have the compiled `gmpy2` Python packages. The error message for that is `cannot find mpfr.h while building wheel` 😁 )

I can separate the dates_calc stuff into another PR if it doesn't get fixed by itself

compiler/surface/lexer.cppo.ml
AltGr: Moved to #626 because there is some silly CI breakage

PR #626, opened 2024-05-28
Update opam dependencies

AltGr:
- The newer sedlex version allows us to remove a weird workaround (https://github.com/ocaml-community/sedlex/issues/12)
- dates_calc 0.0.6 is required (no dependency on `Str`) so we can remove specific pinning from the CI

This is stalled at the moment though because of Alpine 3.20 and Python stuff

### DIFF
--- a/.github/workflows/harness.yml
+++ b/.github/workflows/harness.yml
@@ -112,7 +112,7 @@ jobs:
       - name: Install LaTeX deps
         # This is done late because caching would not benefit compared to
         # installation through apk (1,5G upload is slow)
-        run: sudo apk add texlive-xetex texmf-dist-latexextra texmf-dist-pictures font-dejavu groff
+        run: sudo apk add texlive-xetex texmf-dist-latexextra texmf-dist-binextra texmf-dist-pictures font-dejavu groff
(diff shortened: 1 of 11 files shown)


AltGr:
For CI to pass as is with Alpine 3.20 would require a musl 1.2 build to appear on this page: https://pypi.org/project/gmpy2/#files

rprimet:
It might not be related to the musl version change, but rather to python 3.12? See https://github.com/aleaxit/gmpy/issues/446

AltGr:
So, this *finally* all good, **but**
- newer alpine requires an additional texmf package for our docs builds, it's added here but not on the CI script on master (i.e. `Run integrated tests` should pass once merged)
- more annoying, the output of the test about backquotes seems to be dependent on the pandoc version we are using.
  image
  I see no good solution for now apart from disabling the test, @denismerigoux what do you think ?

AltGr:
For me this is ready for merge, all issues have been resolved and CI should pass once merged. @vincent-botbol can you have a look before I merge ?

vincent-botbol:
Looks good to me!

vincent-botbol (approved):

Issue #250, opened 2022-04-29
Printf multiple errors before exiting


denismerigoux:
Currently, the Catala compiler exits at the first error it encounters. It would be nice to only exit when the errors it has found forbid it to continue processing, so that it displays multiple related error messages in the same run.

For that, every AST should be augmented with an `Error` case produced when an error message is raised. The compiler could then continue its operations until it actually needs to do something with the location of the AST where `Error` is, in which case it would exit.

pierregoutagny:
I think it would be quite useful to me as well to have an `Error` case in the AST, because it would allow me to continue other branches of evaluation after an error is found.
I have been considering using OCaml's `Result` module to make the concolic evaluator return a `((dcalc,'m) gexpr, type_for_errors) result`, but an AST case would probably be better. How much effort do you think it would need to at least add the `Error` AST case (even if it's underused at first)?

denismerigoux:
Vincent has implemented this for parsing and typechecking errors only. This issue is closed in favor of a new one that should generalize this mechanism.

PR #630, opened 2024-06-05
Multiple parsing errors reporting

vincent-botbol:
This PR is a partial attempt at fixing #250. 

We first focus on the surface parsing in order to display possible multiple parsing errors.
In its current state, the (very) naive heuristic used is that when encountering a unexpected token, the parser registers the error, replaces the faulty token by guessing a "candidate-token" that would either be syntactically close to the original input or to pick a random valid token among the expected ones. 

This approach may yield confusing errors in particular when the selected candidates are contextually bad.

We will need to further discuss which strategies to implement as the state of the art on this topic is broad.

**EDIT**

I went for a heuristic that, when an unexpected token is encountered, first checks whether the parser could continue without this faulty token and also checks with possible candidates tokens. Then, it chooses the best strategy, i.e., the token which allowed the parser to progress the furthest, records the error, and resumes the parsing applying this strategy on the eventual conflicts.

This approach seems to yield reasonable results on the invalid set of programs I crafted. This may require a bit more testing and fine-tuning.

~~I'm still struggling to find a clean way to combine error messages but perhaps this can be done afterwards.~~
Thanks to @AltGr suggestions, the changes required to adapt to the error messaging are now reasonable.

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -327,6 +327,7 @@ open Content
 (** {1 Error exception} *)
 
 exception CompilerError of Content.t
+exception CompilerErrors of Content.t list
 
 (** {1 Error printing} *)
 
@@ -404,3 +405,31 @@ let result = make ~level:Result ~cont:emit
 let results r = emit (List.flatten (List.map of_result r)) Result
 let warning = make ~level:Warning ~cont:emit
 let error = make ~level:Error ~cont:(fun m _ -> raise (CompilerError m))
+
+(* Multiple errors handling *)
+let global_errors = ref None
+
+let delayed_error x =
+  make ~level:Error ~cont:(fun m _ ->
+      match !global_errors with
+      | None ->
+        failwith
+          "delayed error called outside scope: encapsulate using \
+           'with_delayed_errors' first"
+      | Some l ->
+        global_errors := Some (m :: l);
+        x)
+
+let with_delayed_errors (f : unit -> 'a) : 'a =
+  (match !global_errors with
+  | None -> global_errors := Some []
+  | Some _ -> failwith "delayed error scope already initialized");
+  let r = f () in
+  match !global_errors with
+  | None -> assert false
+  | Some [] ->
+    global_errors := None;
+    r
+  | Some errs ->
+    global_errors := None;
+    raise (CompilerErrors (List.rev errs))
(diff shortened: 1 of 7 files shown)


denismerigoux:
@maximedenes, let me introduce you to @vincent-botbol, the new research engineer on Catala ! @vincent-botbol, @maximedenes also works at Inria and does the liaison with Inria's DGDI for Catala. However, Maxime also knows a lot about parsing, error reporting and crafting language server protocol instances, so I'll add him to the discussion.

I don't know myself what are the common methods to resume a parsing interrupted by an error message. What you propose @vincent-botbol sounds reasonable and I guess we can start from here.

vincent-botbol:
@maximedenes Nice to meet you!

I'm quite rusty in parsing so if you have any input I'd be more than thankful.
I found some heuristics which might be a good fit for catala but there might be a simpler solution that I don't see.

vincent-botbol:
> I see it's no longer WIP :) Can we merge it ?

I think @AltGr wants to take a look

compiler/catala_utils/message.ml
```diff
@@ -404,3 +405,31 @@ let result = make ~level:Result ~cont:emit
...
+(* Multiple errors handling *)
+let global_errors = ref None
+
+let delayed_error x =
+  make ~level:Error ~cont:(fun m _ ->
+      match !global_errors with
+      | None ->
+        failwith
+          "delayed error called outside scope: encapsulate using \
+           'with_delayed_errors' first"
```
AltGr, line 418: Ideally this would be `error ~internal:true`: we'd like to avoid rogue `Failure`s

(also, including the original error might be helpful ? Not sure)

compiler/catala_utils/message.ml
```diff
@@ -404,3 +405,31 @@ let result = make ~level:Result ~cont:emit
...
+      | Some l ->
+        global_errors := Some (m :: l);
+        x)
+
+let with_delayed_errors (f : unit -> 'a) : 'a =
+  (match !global_errors with
+  | None -> global_errors := Some []
+  | Some _ -> failwith "delayed error scope already initialized");
```
AltGr, line 426: Same here

compiler/driver.ml
```diff
@@ -1204,6 +1204,11 @@ let main () =
     Message.Content.emit content Error;
     if Global.options.debug then Printexc.print_raw_backtrace stderr bt;
     exit Cmd.Exit.some_error
```
AltGr, line 1206: I might like a debug message telling me that there were pending delayed errors when a non-delayed fatal error cut them short :)

AltGr (approved):
Wow, pretty nice job on the parser checkpointing...

I have a couple small nitpicks, but feel free to merge and address them later if you prefer.

I guess we'll need to see how that fares in practice :)

PR #631, opened 2024-06-17
Multiple typing errors reporting

vincent-botbol:
Depends on  #630, do not merge before it lands.

This PR is a follow-up of #630 where we add multiple error reporting for typing errors.

While this PR is easier than the previous one, I did a small hack in order to fix the consistency of delayed error pretty-printing messages. Indeed, in order for the typing to progress and discover more errors, a faulty unification is performed which unifies incompatible types. The delayed look-up performed during the error pretty-printing would then display confusing error messages such as:
```
Error during typechecking, incompatible types:
─➤ integer
─➤ integer
```
which is not ideal. 
Thus, I decided transform union-find types in AST ones when registering the error to fix this behavior.

### DIFF
--- a/compiler/shared_ast/typing.ml
+++ b/compiler/shared_ast/typing.ml
@@ -175,30 +175,34 @@ let rec colors =
   let open Ocolor_types in
   blue :: cyan :: green :: yellow :: red :: magenta :: colors
 
+let dummy_flags = { fail_on_any = false; assume_op_types = false }
 let format_typ ctx fmt naked_typ = format_typ ctx ~colors fmt naked_typ
 
-let record_type_error ctx (A.AnyExpr e) t1 t2 =
-  (* TODO: if we get weird error messages, then it means that we should use the
-     persistent version of the union-find data structure. *)
-  let t1_repr = UnionFind.get (UnionFind.find t1) in
-  let t2_repr = UnionFind.get (UnionFind.find t2) in
+let record_type_error _ctx (A.AnyExpr e) t1 t2 =
+  (* We convert union-find types to ast ones otherwise error messages would be
+     hindered as union-find side-effects wrongly unify both types. The delayed
+     pretty-printing would yield messages such as: 'incompatible types (integer,
+     integer)' *)
+  let t1_repr = typ_to_ast ~flags:dummy_flags t1 in
+  let t2_repr = typ_to_ast ~flags:dummy_flags t2 in
   let e_pos = Expr.pos e in
   let t1_pos = Mark.get t1_repr in
   let t2_pos = Mark.get t2_repr in
+  let pp_typ = Print.typ_debug in
   let fmt_pos =
     if e_pos = t1_pos then
       [
         ( (fun ppf ->
             Format.fprintf ppf "@[<hv 2>@[<hov>%a@ %a@]:" Format.pp_print_text
-              "This expression has type" (format_typ ctx) t1;
+              "This expression has type" pp_typ t1_repr;
             if Global.options.debug then
               Format.fprintf ppf "@ %a@]" Expr.format e
             else Format.pp_close_box ppf ()),
           e_pos );
         ( (fun ppf ->
             Format.fprintf ppf
-              "@[<hov>Expected@ type@ %a@ coming@ from@ expression:@]"
-              (format_typ ctx) t2),
+              "@[<hov>Expected@ type@ %a@ coming@ from@ expression:@]" pp_typ
+              t2_repr),
           t2_pos );
       ]
     else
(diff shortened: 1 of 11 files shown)


compiler/shared_ast/typing.ml
```diff
@@ -159,73 +154,34 @@ let rec format_typ
...
-  (* TODO: if we get weird error messages, then it means that we should use the
-     persistent version of the union-find data structure. *)
-  let t1_repr = UnionFind.get (UnionFind.find t1) in
-  let t2_repr = UnionFind.get (UnionFind.find t2) in
+let record_type_error ctx (A.AnyExpr e) t1 t2 =
+  (* We convert union-find types to ast ones otherwise error messages would be
+     hindered as union-find side-effects wrongly unify both types. The delayed
+     pretty-printing would yield messages such as: 'incompatible types (integer,
+     integer)' *)
+  let t1_repr = typ_to_ast ~flags:dummy_flags t1 in
```
denismerigoux, line 183: Very good observation.

compiler/shared_ast/typing.ml
```diff
@@ -1107,3 +1098,7 @@ let program ?fail_on_any ?assume_op_types prg =
             prg.decl_ctx.ctx_enums;
       };
   }
+
+let program ?fail_on_any ?assume_op_types prg =
+  Message.with_delayed_errors (fun () ->
+      program ?fail_on_any ?assume_op_types prg)
```
denismerigoux, line 1104: Can there be an option to delay the errors or not if we want to?

denismerigoux (approved):
It looks good to me!

compiler/shared_ast/typing.ml
vincent-botbol: After discussions, I also added an optional argument that defaults to the CLI option.

compiler/shared_ast/typing.ml
```diff
@@ -959,17 +957,10 @@ and typecheck_expr_top_down :
...
-  try f e
-  with Type_error (e, ty1, ty2) -> (
-    let bt = Printexc.get_raw_backtrace () in
-    try handle_type_error ctx e ty1 ty2
-    with e -> Printexc.raise_with_backtrace e bt)
-
-let wrap_expr ctx f e =
+let wrap_expr f e =
   (* We need to unbox here, because the typing may otherwise be stored in
      Bindlib closures and not yet applied, and would escape the `try..with` *)
```
AltGr, line 962: Comment about `try..with` no longer appropriate ; probably the `unbox` can be inlined at the call sites.

tests/bool/bad/test_xor_with_int.catala_en
```diff
@@ -29,5 +29,24 @@ $ catala Typecheck
...
+├─ 'xor' should be a boolean operator
+│
+│ Expected type bool coming from expression:
+├─➤ tests/bool/bad/test_xor_with_int.catala_en:8.33-8.36:
+│   │
+│ 8 │   definition test_var equals 10 xor 20
+│   │                                 ‾‾‾
+└─ 'xor' should be a boolean operator
 #return code 123#
 ```
```
AltGr, line 52: Hm, this is a bit redundant... Maybe we could filter type errors that are "too close" to each other?
Or something much simpler could be to just give up typing of the current definition, and keep typing the rest of the program... since all types need to be declared, it wouldn't propagate errors.

Just thinking about the different options, it's good to experiment here.

tests/typing/bad/mult_errs1.catala_en
```diff
@@ -0,0 +1,52 @@
...
+│   │
+│ 7 │     Structure { -- i: 4.1 -- e: y };
+│   │                       ‾‾‾
+│
+│ Expected type integer coming from expression:
+├─➤ tests/typing/bad/common.catala_en:8.18-8.25:
+│   │
+│ 8 │   data i content integer
+│   │                  ‾‾‾‾‾‾‾
+└─
```
AltGr, line 50: In this case, it's pretty relevant though 🤔

compiler/shared_ast/typing.ml
```diff
@@ -159,73 +154,34 @@ let rec format_typ
...
+    | A.TAny when not Global.options.debug -> Format.pp_print_string fmt "list"
     | _ -> Format.fprintf fmt "@[list of@ %a@]" (format_typ ~colors) t1)
   | TDefault t1 ->
     Format.pp_print_as fmt 1 "⟨";
     format_typ ~colors fmt t1;
     Format.pp_print_as fmt 1 "⟩"
-  | TAny v ->
-    if Global.options.debug then Format.fprintf fmt "<a%d>" (Any.id v)
-    else Format.pp_print_string fmt "<any>"
+  | A.TAny -> Format.pp_print_string fmt "<any>"
```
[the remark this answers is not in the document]
vincent-botbol: I reverted this and I'm using `Print.typ_debug` instead

compiler/shared_ast/typing.ml
vincent-botbol: Fixed

tests/typing/bad/mult_errs1.catala_en
vincent-botbol: Same remark as above :)

compiler/shared_ast/typing.ml
```diff
@@ -111,12 +110,9 @@ let rec format_typ
     (ctx : A.decl_ctx)
     ~(colors : Ocolor_types.color4 list)
     (fmt : Format.formatter)
-    (naked_typ : unionfind_typ) : unit =
+    (typ : A.typ) : unit =
```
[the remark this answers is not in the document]
AltGr: This seems fine ; actually `typ_debug` here is a misnomer, and `typ` prints the details of types for historical reasons but we might want to change that globally.

AltGr (approved):

tests/bool/bad/test_xor_with_int.catala_en
AltGr: Note that in this case, this might be tweaked by changing the order in which typing is done (e.g. typing the arguments first, then the operator, rather than the opposite ; ofc it would be worse on `true xor 42`)

PR #632, opened 2024-06-18
Generate tests reports from 'clerk test'

AltGr:
This is a proper replacement for the previous shell-based placeholder hack.

Here is a summary:

- `clerk runtest` (normally run by ninja) is much extended:
  * besides generating the test@out file, it checks individual tests for success
    and can write a report file containing their status, and the positions for
    their (expected/current) outputs (this uses `Marshal`)
  * it now handles out-tests directly in addition to inline-tests, for which
    it generates the separate output file ; they are included in the report

- ninja is now tasked with building all the test reports (which shouldn't fail);
  for directories, individual reports are concatenated (as before).
  Removing intermediate report rules, and out-test rules means that the ninja
  file is much simplified.

- then, clerk takes back control, reads the final reports and formats them in a
  user-friendly way. Printing the reports may imply running `diff` internally.
  In particular, the commands to easily reproduce each test are provided.
  Resetting the test results if required is also done directly by clerk, at this
  stage.

A few switches are available to customise the output, but I am waiting for some
feedback before deciding what to make available from the CLI.

The `clerk report` command is available to manually explore test reports, but
normally the processing is done directly at the end of `clerk test` (i.e. ninja
will no longer call that command)

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -208,6 +208,34 @@ module Cli = struct
             "Flags or targets to forward to Ninja directly (use $(b,-- \
              ninja_flags) to separate Ninja flags from Clerk flags)")
 
+  let report_verbosity =
+    Arg.(
+      value
+      & vflag `Failures
+          [
+            ( `Summary,
+              info ["summary"] ~doc:"Only display a summary of the test results"
+            );
+            ( `Short,
+              info ["short"] ~doc:"Don't display detailed test failures diff" );
+            ( `Failures,
+              info ["failures"]
+                ~doc:"Show details of files with failed tests only" );
+            ( `Verbose,
+              info ["verbose"; "v"]
+                ~doc:"Display the full list of tests that have been run" );
+          ])
+
+  let use_patdiff =
+    Arg.(
+      value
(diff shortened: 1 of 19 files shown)


AltGr:
Sneak peek:
image

AltGr:
For the default diff output, I opted for a side-by-side (but reversed the colors so that reference is green and added current output is red) ; it seemed the easiest to grasp, although there is support for using `patdiff` instead if installed.
I am not very keen on recoding a custom diff algorithm, but it might otherwise be difficult to further customise that output and be sure we behave correctly in the absence of GNU diff.

AltGr:
Note that the "law context" (sections/subsections etc.) is not included in the reports ; maybe that will be useful to end-users at some point, but it would require an extension to our line-parser, and could significantly improve verbosity, so I am not sure if it is worth it at this point.
Before we go that way, we might want to include that information directly at parsing time rather than in a post-processing pass ?

vincent-botbol:
Made a suggestion to remove underlines to prevent weird coloring happening for users without a fancy terminal

vincent-botbol (approved):

PR #636, opened 2024-06-20
Refactor suggestions mechanism

vincent-botbol:
This PR (at least subjectively) improves the output of some suggestions in errors.

For instance, the previous output would suggest every possible struct field where now it only suggests few of them ordered by relevance:
example

It also removes weird `.. or ..` between each suggested item.

### DIFF
--- a/compiler/catala_utils/suggestions.ml
+++ b/compiler/catala_utils/suggestions.ml
@@ -48,51 +48,34 @@ let levenshtein_distance (s : string) (t : string) : int =
 
   d.(m).(n)
 
-(*We create a list composed by strings that satisfy the following rule : they
-  have the same levenshtein distance, which is the minimum distance between the
-  reference word "keyword" and all the strings in "candidates" (with the
-  condition that this minimum is equal to or less than one third of the length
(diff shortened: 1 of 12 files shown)


denismerigoux (approved):
Yes, thank you very much, this is very helpful!

compiler/catala_utils/suggestions.ml
```diff
@@ -48,59 +48,54 @@ let levenshtein_distance (s : string) (t : string) : int =
...
+  let candidates =
+    List.map
+      (fun (_, l) -> List.rev l)
+      (M.bindings (compute_candidates suggs given))
+  in
+  List.concat candidates |> sub [] max_elements
 
-let format (ppf : Format.formatter) (suggestions_list : string list) =
-  match suggestions_list with
+let pp ppf suggs =
```
AltGr, line 80: Other modules in Catala use `format` rather than `pp`.
No strong feeling about it though

AltGr (approved):

compiler/catala_utils/suggestions.ml
vincent-botbol: Oh, ok. Let's keep it consistent then.

PR #637, opened 2024-06-21
Testing and debugging closure conversion

AltGr:
This adds support for closure conversion in the OCaml backend (useful for testing, probably nothing else) ; brings support throughout the toolchain ; and fixes a number of small bugs, some of them nasty.

With this, `clerk test --test-flags=--lcalc,--closure-conversion` fully passes (it's not added to the default testsuite yet, though)

A few asides are included:
- fixed regression with printing internal typing errors (and now show full program in debug mode when they happen)
- some Clerk fine-tuning following the reporting feature


denismerigoux (approved):
Nice, and I like it with the extra debug function that will come very handy!

PR #638, opened 2024-06-21
Clerk reports: postprocess diff output

AltGr:
This relies less on specific color flags of GNU diff, and reformats and colorises the output.
(it may still depend on the specific layout of GNU diff with the `-y` flag though)

Since we already have `diffutils` advertised as dev-mode dependency this should be OK
(alpine / busybox basic diff doesn't even support `-y` anyway)

### DIFF
--- a/build_system/clerk_report.ml
+++ b/build_system/clerk_report.ml
@@ -69,31 +69,87 @@ let has_command cmd =
   let check_cmd = Printf.sprintf "type %s >/dev/null 2>&1" cmd in
   Sys.command check_cmd = 0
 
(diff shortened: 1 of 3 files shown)


denismerigoux (approved):
OK, so you've implemented the "parsing diff's output" strategy

AltGr:
Not really: I only use the already-formatted side-by-side formatting (without color) and add some borders and lines to it.

The other approach I had in mind would use a unified diff input and generate the whole layout. I guess we'll see how robust this proposition is.

PR #639, opened 2024-06-26
Support for basic `clerk.toml` configuration files

AltGr:
There aren't many options at the moment, but support is in place. This doesn't include any documentation yet.

`clerk.toml` files can already be used to:
- specify the root of the project to build (you could already do `cd examples && clerk test arithmetic/`, but it relied on ocaml / git project root markers only)
- specify the list of include directories. In particular, this makes it straightforward to just run `clerk` directly within `catala-examples` or `ir-catala`, while before you had to rely on `make` to properly specify the include flags.

The default config is currently as follows (it might be a good idea to offer a flag that generates it):
```toml
[project]
include_dirs = []

[build]
catala_opts = []
build_dir = "_build"
```

We rely on [`otoml`](https://opam.ocaml.org/packages/otoml/) for toml parsing, and I added a small layer of error reporting for unexpected keys or keys with wrong types.


AltGr:
Note: there is a sister branch on `catala-examples` (https://github.com/CatalaLang/catala-examples/pull/10) which adds the `clerk.toml` file and removes the global export of the `CATALA_INCLUDE` variable (although these includes are still needed for direct catala calls made from the Makefile, i.e. `catala depends`, `catala api_web`, etc. -- but these advanced targets are intended to be handled by Clerk at some point)

`catala-examples` would still build fine without any changes, but within the `make alltest` target from the catala project it may run into some trouble because detection of the project root now prioritises a `clerk.toml` file above e.g. a `.git` directory or file.

vincent-botbol (approved):
Looks good to me

PR #642, opened 2024-07-04
Add JUnit-compatible test output to `clerk report`

AltGr:
And use it for test reports in Github.


AltGr:
Hm, integrating test results was probably a good intent, but that was counting without the greatness of Github CI:
image

PR #643, opened 2024-07-08
Remove the R backend

AltGr:
There are no users at the moment, so we won't be actively maintaining it.

If and when the need arises again, we can revert this commit and resurrect it.


denismerigoux (approved):
Thanks @AltGr !

PR #659, opened 2024-07-17
Fix scopelang scope definition variable positions

vincent-botbol:
This PR adds the positions of scope variables (possibly multiple) definitions in scopelang's AST which was missing. This is necessary in order to implement the LSP jump-to-def feature. Previously, the only position hauled from desugared was the declaration's one.

### DIFF
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -589,20 +589,21 @@ let translate_rule
   match rule with
   | S.ScopeVarDefinition { var; typ; e; _ }
   | S.SubScopeVarDefinition { var; typ; e; _ } ->
+    let scope_var = Mark.remove var in
+    let decl_pos = Mark.get (ScopeVar.get_info scope_var) in
     let pos_mark, _ = pos_mark_mk e in
     let scope_let_kind, io =
       match rule with
       | S.ScopeVarDefinition { io; _ } -> ScopeVarDefinition, io
       | S.SubScopeVarDefinition _ ->
-        let pos = Mark.get var in
         ( SubScopeVarDefinition,
-          { io_input = NoInput, pos; io_output = false, pos } )
+          { io_input = NoInput, decl_pos; io_output = false, decl_pos } )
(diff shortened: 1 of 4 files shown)


compiler/dcalc/from_scopelang.ml
```diff
@@ -589,20 +589,21 @@ let translate_rule
   match rule with
   | S.ScopeVarDefinition { var; typ; e; _ }
   | S.SubScopeVarDefinition { var; typ; e; _ } ->
+    let scope_var, _ = var in
+    let decl_pos = snd (ScopeVar.get_info scope_var) in
```
AltGr, line 593: We prefer to use `Mark.get` / `Mark.remove` for the pairs defined as `Mark.ed` ; although we do match them as pairs very often (doing differently would be extremely verbose given how pervasive this is), this is still more descriptive than the built-ins `fst`/`snd`.

compiler/scopelang/ast.ml
```diff
@@ -40,13 +40,13 @@ let rec locations_used (e : 'm expr) : LocationSet.t =
 
 type 'm rule =
   | ScopeVarDefinition of {
-      var : ScopeVar.t Mark.pos;
+      var : ScopeVar.t * Pos.t list;
```
AltGr, line 43: For consistency we _might_ want to call that `(ScopeVar.t, Pos.t list) Mark.ed` ; but to be honest I am not sure how much it improves anything.

AltGr (approved):
Just a small remark on consistency w.r.t. usage of the `Mark` module, but I'm not even sure.
Thanks!

denismerigoux (approved):
Looks very good! Being able to quickly browse between all the definitions of a same variable will be a blessing...

compiler/dcalc/from_scopelang.ml
vincent-botbol: Agreed, fixed.

compiler/scopelang/ast.ml
vincent-botbol: It looks nicer this way anyhow :)

PR #660, opened 2024-07-30
Fix missing keywords in emacs mode

vincent-botbol:
This small PR adds some of the recently introduced keywords.


AltGr:
Note that if you have emacs with support for tree-sitter (i.e. 29+) there is a much better mode (including automatic indentation) at https://github.com/CatalaLang/tree-sitter-catala/

PR #661, opened 2024-07-30
Add LSP required features

vincent-botbol:
This PR adds/exposes features required to implement the LSP server:

1. Refactoring of the error messages so that the LSP server is able to be notified on each encountered error;
2. Making `clerk_config` a sub-package in order to be reachable by the LSP server;
3. Fixes some small bugs: e.g., error reporting not considering the correct input source, leaking lexing errors, etc.

With this PR, the current state of LSP server will have all what it needs upstreamed (and thus, won't require to pin on a branch).

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -275,10 +275,10 @@ module Content = struct
     restore_ppf ();
     Format.pp_print_newline ppf ()
 
-  let emit ?(pp_marker = pp_marker) (content : t) (target : level) : unit =
+  let emit ?ppf ?(pp_marker = pp_marker) (content : t) (target : level) : unit =
+    let ppf = Option.value ~default:(get_ppf target) ppf in
     match Global.options.message_format with
     | Global.Human -> (
-      let ppf = get_ppf target in
       match target with
       | Debug | Log -> basic_msg ~pp_marker ppf target content
       | Result | Warning | Error -> fancy_msg ~pp_marker ppf target content)
@@ -288,7 +288,6 @@ module Content = struct
          message everywhere there is not a more precise message. If we can't
          find a position without a more precise message, we just take the first
          position in the list to pair with the message. *)
-      let ppf = get_ppf target in
       Format.pp_print_list ~pp_sep:Format.pp_print_newline
         (fun ppf elt ->
           let pos, message =
@@ -327,10 +326,10 @@ module Content = struct
         ppf content;
       Format.pp_print_newline ppf ()
 
-  let emit_n (target : level) = function
+  let emit_n ?ppf (target : level) = function
     | [content] -> emit content target
     | contents ->
-      let ppf = get_ppf target in
+      let ppf = Option.value ~default:(get_ppf target) ppf in
       let len = List.length contents in
       List.iteri
         (fun i c ->
@@ -340,7 +339,7 @@ module Content = struct
           emit ~pp_marker c target)
         contents
 
-  let emit (content : t) (target : level) = emit content target
+  let emit ?ppf (content : t) (target : level) = emit ?ppf content target
 end
 
 open Content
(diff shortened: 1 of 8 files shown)


compiler/catala_utils/message.ml
```diff
@@ -350,6 +349,18 @@ open Content
...
 exception CompilerErrors of Content.t list
 
+type error_kind = Lexing | Parsing | Typing | Generic
+
+type error = {
+  kind : error_kind;
+  msg : string;
+  pos : Pos.t option;
+  suggestion : string list option;
+}
```
denismerigoux, line 359: What's the use for this? Doesn't it duplicate `Content.t` ?

compiler/catala_utils/message.mli
```diff
@@ -66,6 +66,17 @@ end
...
+type error_kind = Lexing | Parsing | Typing | Generic
+
+type error = {
+  kind : error_kind;
+  msg : string;
+  pos : Pos.t option;
+  suggestion : string list option;
+}
+
+val register_error_notifier : (error -> unit) -> unit
```
denismerigoux, line 78: So if I understand correctly this `error` type is the one expected by the LSP for an error to be displayed in an IDE ? If so, it might be renamed into `lsp_error` and the function to `register_lsp_error_notifier`.

denismerigoux (approved):
Thanks @vincent-botbol ! Good for me expect to renaming suggestions.

compiler/catala_utils/message.ml
vincent-botbol: It mostly does, I was wary of exposing `Content.t` but if you think that's fine, we can do it. Perhaps as a private type.

compiler/catala_utils/message.mli
vincent-botbol: It is not particularly meant to be specific to LSP. Although, as of right now, I don't see any other immediate use-cases. I'm fine renaming it if you'd like.

vincent-botbol: Changed `error` into `lsp_error`

AltGr (commented):
This is fine apart for the error message printing that is regressed back from formatters to raw strings.

compiler/catala_utils/message.ml
```diff
@@ -425,7 +436,20 @@ let log = make ~level:Log ~cont:emit
...
+
+let error ?(kind = Generic) : ('a, 'exn) emitter =
+ fun ?header ?internal ?pos ?pos_msg ?extra_pos ?fmt_pos ?outcome ?suggestion
+     fmt ->
+  make ?header ?internal ?pos ?pos_msg ?extra_pos ?fmt_pos ?outcome ?suggestion
+    fmt ~level:Error ~cont:(fun m _ ->
+      Option.iter
+        (fun f ->
+          let buf = Buffer.create 256 in
+          let ppf = Format.formatter_of_buffer buf in
```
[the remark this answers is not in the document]
vincent-botbol: There was a misconception about what was done. In any case, I switched the `string` part to a printing function instead.

compiler/catala_utils/message.mli
```diff
@@ -66,6 +66,17 @@ end
 exception CompilerError of Content.t
 exception CompilerErrors of Content.t list
 
+type error_kind = Lexing | Parsing | Typing | Generic
+
+type error = {
+  kind : error_kind;
+  msg : string;
```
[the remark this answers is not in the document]
vincent-botbol: I switched the string part to a printing function instead.

compiler/surface/parser_driver.ml
```diff
@@ -374,36 +374,46 @@ and expand_includes (source_file : string) (commands : Ast.law_structure list) :
...
-                acc.Ast.program_source_files;
-            Ast.program_items =
-              List.rev_append includ_program.program_items acc.Ast.program_items;
-            Ast.program_used_modules =
-              List.rev_append includ_program.program_used_modules
-                acc.Ast.program_used_modules;
-            Ast.program_lang = language;
-          }
+          let pos = Mark.get inc_file in
+          if not (Sys.file_exists sub_source) then
```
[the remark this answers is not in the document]
vincent-botbol: Done, thanks.

AltGr:
Thanks!

PR #662, opened 2024-08-01
Fix typing error missing position

vincent-botbol:


AltGr (approved):
`Message.join_pos` might deserve a small comment, esp. as it's pretty different from `Pos.join`

vincent-botbol:
> `Message.join_pos` might deserve a small comment, esp. as it's pretty different from `Pos.join`

True, I added a comment. Thanks.

PR #663, opened 2024-08-01
Introduce modules in toml config

vincent-botbol:
This PR follows the recent discussions introducing module definitions in toml config files.

The suggested design in this PR looks as follow in a config file:

```toml
[[module]]
name = "My_module_name"
module_uses = [
  "Foo",
  [ "Bar", "B" ], # translate to '> Using Bar as B'
]
includes = [
  "path/to/a.catala_en"
]

[[module]]
name = "Allocations_familiales"
module_uses = [
  [ "France" ],
  "Bmaf",
  "Smic",
  "Prestations_familiales",
]
includes = [
  "allocations_familiales/prologue.catala_fr",
  "allocations_familiales/autres_codes.catala_fr",
  "allocations_familiales/securite_sociale_L.catala_fr",
  "allocations_familiales/securite_sociale_R.catala_fr",
  "allocations_familiales/securite_sociale_D.catala_fr",
  "allocations_familiales/decrets_divers.catala_fr",
  "allocations_familiales/epilogue.catala_fr",
]
```

The expected workflow would be to define multiple ("meta-")modules per project's config file.

Edit: removed "language"

### DIFF
--- a/build_system/clerk_config.ml
+++ b/build_system/clerk_config.ml
@@ -18,7 +18,6 @@ open Catala_utils
 open Otoml
 
 type modul = {
-  language : Global.backend_lang;
   module_uses : (string * string option) list;
   includes : string list;
 }
@@ -38,9 +37,9 @@ let default =
     modules = String.Map.empty;
   }
 
-let parse_module_uses modul =
+let parse_module_uses ~fname ~name modul =
   match find_opt modul (get_array get_value) ["module_uses"] with
-  | None -> []
+  | None | (exception _) -> [] (* [opt] is a lie. *)
   | Some module_uses ->
     List.map
       (function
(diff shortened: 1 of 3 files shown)


AltGr (approved):
Two small remarks:
- I am not sure the `language` field is relevant, can we do without ?
- From reading the code, it seems that unrecognised fields in a module should correctly trigger errors, but I am not 100% sure

vincent-botbol:
> Two small remarks:
> 
>     * I am not sure the `language` field is relevant, can we do without ?
> 
>     * From reading the code, it seems that unrecognized fields in a module should correctly trigger errors, but I am not 100% sure

I removed the language field, it isn't particularly necessary at the moment. We can always determine it from the files.

PR #664, opened 2024-08-05
Implement safe renaming of idents for backend printing

AltGr:
Previously we had some heuristics in the backends trying to achieve this with a lot of holes ; this should be much more solid, relying on `Bindlib` to do the correct renamings.

**Note1**: it's not plugged into the backends other than OCaml at the moment.

**Note2**: the related, obsolete heuristics haven't been cleaned out yet

**Note3**: ~~we conservatively suppose a single namespace at the moment. This is required for e.g. Python, but it forces vars named like struct fields to be renamed, which is more verbose in e.g. OCaml. The renaming engine could be improved to support different namespaces, with a way to select how to route the different kinds of identifiers into them.~~ [EDIT: I haven't implemented real separate namespaces at the moment, but added an option to allow overlap of field names and constructors, so that there is no regression for the OCaml backend]

Similarly, customisation for what needs to be uppercase or lowercase is not available yet.

**Note4**: besides excluding keywords, we should also be careful to exclude (or namespace):
- the idents used in the runtime (e.g. `o_add_int_int`)
- the dynamically generated idents (e.g. `embed_*`)

**Note5**: module names themselves aren't handled yet. The reason is that they must be discoverable by the user, and even need to match the filenames, etc. In other words, imagine that `Mod` is a keyword in the target language. You can't rename a module called `Mod` to `Mod1` without knowing the whole module context, because that would destroy the mapping for a module already called `Mod1`.

A reliable solution would be to translate all module names to e.g. `CatalaModule_*`, which we can assume will never conflict with any built-in, and forbid idents starting with that prefix. We may also want to restrict their names to ASCII ? Currently we use a projection, but what if I have two modules called `Là` and `La` ?


vincent-botbol (approved):
Looks good overall. However, I'm discovering this part of the code base so my review is not in-depth but given the test output it seems to do what it says.

PR #666, opened 2024-08-08
Renaming idents, second stage: use proper renaming in scalc and Python

AltGr:
This completes (and builds on top of) #664 :
renaming is now in its own module, customisable enough, and called with the proper options for each backend. In particular, this simplifies the generated Python code, while being at the same time much more reliable.

- the runtime idents used are still not registered as reserved names
- no specific support for module names yet


AltGr:
CI fails because of a bug in the python printer, independent from this (an ident became longer, which triggered a line cut at a point that python doesn't like)

compiler/scalc/print.ml
```diff
@@ -21,10 +21,12 @@ open Ast
 let needs_parens (_e : expr) : bool = false
 
 let format_var_name (fmt : Format.formatter) (v : VarName.t) : unit =
-  Format.fprintf fmt "%a_%d" VarName.format v (VarName.id v)
+  VarName.format fmt v
+(* Format.fprintf fmt "%a_%d" VarName.format v (VarName.id v) *)
```
vincent-botbol, line 25:
```suggestion
```

compiler/scalc/print.ml
```diff
@@ -21,10 +21,12 @@ open Ast
...
 
 let format_func_name (fmt : Format.formatter) (v : FuncName.t) : unit =
-  Format.fprintf fmt "@{<green>%a_%d@}" FuncName.format v (FuncName.id v)
+  FuncName.format fmt v
+(* Format.fprintf fmt "@{<green>%a_%d@}" FuncName.format v (FuncName.id v) *)
```
vincent-botbol, line 29:
```suggestion
```

vincent-botbol (approved):
The outputs look much better overall. The surrounding parenthesis in the generated python code is unfortunate but, at least, it works.

Same remark as for #664: I'm not sufficiently knowledgeable for super relevant remarks but based on the output, it looks good to me.

PR #669, opened 2024-08-27
Cleanup nested matches

AltGr:
This PR has two related patches, but with different scopes:

- When translating default terms from dcalc to matches with the
  `HandleExceptions` operator in dcalc, be a little more clever when there is 0
  or 1 exception. This makes un-optimised code much more readable, which is very
  helpful for debugging

- During the optimisation pass, handling of nested matches is generalised from
  the few specific cases that were handled. When a matched term has only literal
  constructors in terminal position, the surrounding match cases are inlined as
  long as this doesn't lead to code duplication.

  This should be robust and preserve traces, and is particularly helpful with
  nested option matches that really obfuscate the resulting code. For example,
  we can get the following code for a top-level definition in lcalc:

```
let set requirements_ownership_met : bool =
  (aggregate_periods_from_last_five_years.0
     aggregate_periods_from_last_five_years.1
     property_ownage)
  >= [0 years, 0 months, 730 days]
```

while we previously had 4 levels of nesting (and of course this
code still has gone through all the expected exceptions-related
validations)

### DIFF
--- a/compiler/lcalc/from_dcalc.ml
+++ b/compiler/lcalc/from_dcalc.ml
@@ -59,7 +59,6 @@ let rec translate_default
     (mark_default : 'm mark) : 'm A.expr boxed =
   (* Since the program is well typed, all exceptions have as type [option 't] *)
   let pos = Expr.mark_pos mark_default in
-  let exceptions = List.map translate_expr exceptions in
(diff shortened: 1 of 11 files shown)


vincent-botbol (approved):
Looks good!

PR #689, opened 2024-09-16
Fix OCaml backend bug regarding exception source positions

AltGr:
Without the fix, the code would contain an array of length 1 for the positions, and printing the conflict message would result in an `Array_out_of_bounds` exception.


vincent-botbol (approved):
Ran the tests with & without the patch: it indeed fixes things!

PR #694, opened 2024-09-24
Fix inconsistent marks

vincent-botbol:
During the development of the Catala's LSP, I encountered weird cases where locations of structures and enumerations references and fields were inconsistent. These were overwrote by the position of their declaration.

This PR fixes this issue and retain the original mark for these constructs. While trying to do so, I needed a way to edit positions. I first tried to amend structures & enums ast nodes but the diff quickly bloated up. 
I settled for a compromise where I added a way to patch an existing `Uid`'s info which slightly breaks the abstraction but makes the changes quite trivial.

### DIFF
--- a/compiler/catala_utils/uid.mli
+++ b/compiler/catala_utils/uid.mli
@@ -47,6 +47,7 @@ module type Id = sig
 
   val fresh : info -> t
   val get_info : t -> info
+  val map_info : (info -> info) -> t -> t
   val compare : t -> t -> int
   val equal : t -> t -> bool
   val format : Format.formatter -> t -> unit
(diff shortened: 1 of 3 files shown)


compiler/catala_utils/uid.mli
```diff
@@ -47,6 +47,7 @@ module type Id = sig
 
   val fresh : info -> t
   val get_info : t -> info
+  val map_info : (info -> info) -> t -> t
```
denismerigoux, line 50: Ok, this is useful.

denismerigoux (approved):
Thanks @vincent-botbol !

PR #703, opened 2024-09-25
Fix handling of the lexer context

AltGr:
The lexing context uses a global reference. This patch ensures it is handled carefully for the lifetime of lexing a given file.

Before that, very weird bugs could happen, for example when a file ended with an unclosed verbatim block, the parsing of module interfaces done later on would start in verbatim context and fail to parse e.g. module use definitions.

The patch additionally adds checks and warns about unclosed blocks at EOF, which before that would be silently accepted (with the above pitfall).

### DIFF
--- a/build_system/clerk_runtest.ml
+++ b/build_system/clerk_runtest.ml
@@ -328,8 +328,8 @@ let run_tests ~catala_exe ~catala_opts ~test_flags ~report ~out filename =
       push_line str;
       run_inline_test lines
     | Some ((str, L.LINE_TEST id, _), lines) ->
-      push_line str;
-      run_output_test id lines
+      run_output_test id lines;
+      push_line str
     | Some ((str, _, _), lines) ->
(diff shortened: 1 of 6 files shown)


denismerigoux (approved):
Wouahou very good catch, thanks for the fix!

denismerigoux:
It fails on the syntax file :) https://github.com/CatalaLang/catala/actions/runs/11038376367/job/30664852918?pr=703#step:5:53

The error message could be a little explicit (what is a terminator ?)

AltGr:
Hm, maybe "Unclosed block or missing newline ?"

AltGr:
Ah apparently what I pushed wasn't the last (working) version 😬 ; fixed that and updated the message to:
```ocaml
        Message.warning "Unclosed block or missing newline at the end of file %a.@ \
                         Did you forget a @{<yellow>```@} ?"
```

Issue #679, opened 2024-09-05
Cross-backend testing


denismerigoux:
`clerk` should have a command like `clerk test --cross-backends` that compiles a module full of test scopes into all available backends, then launch the tests and checks whether they pass in the backends as well as the interpreter. For this to work, the tests should use assertions to check that computed outputs equal expected outputs.

denismerigoux:
Decision : @AltGr added a pass in the compilation chain at the Dcalc level. If enabled, this pass adds for every test scope (a scope with no input) a set of asserts on its result variables that check whether the output is equal to the output as computed by the interpreter. Then, these asserts can be carried to every compilation backend.

denismerigoux:
Missing clerk integration as well as test framework integration.

PR #706, opened 2024-10-01
Add `--autotest` option

AltGr:
This provides a solution to #679.

Also included in the PR are a few fixes to issues that were found on the way.

### DIFF
--- a/compiler/catala_utils/cli.ml
+++ b/compiler/catala_utils/cli.ml
@@ -286,6 +286,20 @@ module Flags = struct
     & flag
     & info ["check-invariants"] ~doc:"Check structural invariants on the AST."
 
+  let autotest =
+    value
+    & flag
+    & info ["autotest"]
+        ~env:(Cmd.Env.info "CATALA_AUTOTEST")
+        ~doc:
+          "Insert automatic test assertions in the compiled program. This \
+           detects all scopes that have no input or context variables, runs \
+           the interpreter to pre-compute their values, then adds runtime \
+           assertions to the program that ensure that the actual output of the \
+           scopes match their pre-computed values. If used on a testing \
+           program with a given backend, this guarantees consistency between \
+           the backend and the interpreter."
+
   let no_typing =
     value
     & flag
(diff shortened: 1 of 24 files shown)


compiler/catala_utils/cli.ml
```diff
@@ -286,6 +286,20 @@ module Flags = struct
...
     & info ["check-invariants"] ~doc:"Check structural invariants on the AST."
 
+  let autotest =
+    value
+    & flag
+    & info ["autotest"]
+        ~env:(Cmd.Env.info "CATALA_AUTOTEST")
+        ~doc:
+          "Insert automatic test assertions in the compiled program. This \
+           detects all scopes that have no input or context variables, runs \
```
denismerigoux, line 296: A test can have some context variables (I don't see any reason to forbid it)...

compiler/dcalc/autotest.mli
```diff
@@ -0,0 +1,38 @@
...
+  Global.backend_lang ->
+  ((dcalc, 'm) gexpr boxed -> (dcalc, 'm) gexpr boxed) ->
+  ScopeName.t ->
+  (dcalc, 'm) gexpr scope_body ->
+  (dcalc, 'm) gexpr scope_body Bindlib.box
+(** If the given scope has any inputs, does nothing but reboxing. Otherwise,
+    this function runs the interpreter on the scope to determine its outputs,
+    then inserts assertions within the scope.
+
+    NOTE: scopes with context variables are *not* treated at the moment ;
```
denismerigoux, line 35: It would be good to clarify if we expect tests never to have context variables. For me that would be useful when you want to define a test as "like this other test but tweak this value"

denismerigoux (commented):
Thanks @AltGr ! OK but reminder : we should decide if tests accept or not context variables. Also for me #679 is not yet solved because there is not a `clerk test --cross-backend` command in this PR :)

AltGr:
Yup! Indeed I'm still working on the Clerk aspect, that's why I didn't write the automatic "closes #XXX" ;)
About scopes with context variables, that wasn't a design decision, just a technical difficulty that I left for later : I agree that they could be nice to have :)

AltGr:
(on scopes with context variables, there is one more thing to do: actually ensure that the `main` function we generate on the backends and that just run the test scopes — currently in OCaml and C, I don't think we generate these at all in Python yet — do run them as well, I don't think it's the case at the moment.

denismerigoux:
Related to #679

PR #708, opened 2024-10-02
Add a catch to error raised by get_command

denismerigoux:
I tried `dune exec clerk -- run --exe="dune exec catala"` and got an uncaught exception.


AltGr (approved):
LGTM!

PR #710, opened 2024-10-02
Dcalc: don't thunk context inputs anymore

AltGr:
With the exceptions backend gone, there is no reason to turn context variables into functions of `unit` anymore: they can be handled directly as (`EPureDefault` | `EEmpty`) in dcalc, and options in lcalc.

This provides a big simplification on some programs, in particular when closure conversion is required.

Note that all complexity is not gone: it's now focused on the interesting case of context _functions_, which are not of type `default` but instead functions that _return_ a default type, allowing the possible exceptions to be delayed until the arguments of the functions are known. The behaviour here is unchanged.

I had to tweak the invariant check on default types (@adelaett could you have a look ?) :
- it now allows raw `default` types as fields of input structs

In the proof mode, unthunked `EPureDefault` had to be explicitely marked as safe with being empty (this check could probably be simplified now that we already get a lot of info by default typing; but I am not yet familiar with the code. @rmonat would you like to have a look ?)

### DIFF
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -28,10 +28,6 @@ type scope_input_var_ctx = {
   scope_input_name : StructField.t;
   scope_input_io : Runtime.io_input Mark.pos;
   scope_input_typ : naked_typ;
-  scope_input_thunked : bool;
-      (* For reentrant variables: if true, the type t of the field has been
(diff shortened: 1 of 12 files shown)


AltGr:
Tests seem to be broken because of https://github.com/CatalaLang/catala-examples/pull/15 , file `impot_revenu/oracles.ml` is missing ?

denismerigoux (approved):
Very cool !

PR #711, opened 2024-10-03
Small collection of fixes and cleanup for the C backend

AltGr:


build_system/clerk_driver.ml
```diff
@@ -485,21 +489,29 @@ let base_bindings catala_exe catala_flags build_dir include_dirs test_flags =
...
   let catala_flags_ocaml =
-    List.filter
-      (function
-        | "-O" | "--optimize" | "--closure-conversion" -> true | _ -> false)
-      test_flags
+    if test_flags = [] && catala_flags0 = [] then ["-O"]
+    else
+      List.filter
+        (function
+          | "-O" | "--optimize" | "--closure-conversion" -> true | _ -> false)
```
denismerigoux, line 498: So you're removing here all extra flags passed by the user that are not in this whitelist? Sounds good for nudging correct use but can lead to very confusing situations if the user does not see their manually specified flag in clerk get passed down to catala.

denismerigoux (approved):
Thanks!

build_system/clerk_driver.ml
AltGr: No, not really: this is limited to `--test-flags` (specific instances of running tests with specific flags), and more importantly to the native backends: e.g. when testing with `--closure-conversion` it must affect the generated OCaml modules ; but other interpreter flags don't make sense for the native backend.

Issue #692, opened 2024-09-19
RFC - Clerk TOML configuration & Expected usage


vincent-botbol:
# RFC - Clerk TOML configuration & Expected usage

Following discussions with Denis regarding a proposal for a project's
TOML configuration file, I drafted this proposal (based on the [catala
examples repo](https://github.com/CatalaLang/catala-examples)).

## Informal Spec.

A Catala project should have a single `Clerk.toml` file at the root of
the directory.

The configuration file may contain the project's global options, e.g., :

```toml
[project]
build_dir = "_build"
include_dirs = [
  "smic",
  "prestations_familiales",
  "allocations_familiales"
]
```

A project is then divided into several parts which may describe two
kind of build targets: modules and targets.

### Modules

A configuration file might contain several Catala's modules which are
meant to be used in other modules and to be tested using the existing
clerk's command: `clerk test <module_name>`.

TOML modules could be also used as entrypoints for generating scope
tests: every immediate scope exposed by this module could be generated
as a test in the vscode extension.

```toml
[[module]]
name = "Allocations_familiales"
module_uses = [
  "France",
  "Bmaf",
  "Smic",
  "Prestations_familiales"
]
includes = [
  "allocations_familiales/prologue.catala_fr",
  "allocations_familiales/autres_codes.catala_fr",
  "allocations_familiales/securite_sociale_L.catala_fr",
  "allocations_familiales/securite_sociale_R.catala_fr",
  "allocations_familiales/securite_sociale_D.catala_fr",
  "allocations_familiales/decrets_divers.catala_fr",
  "allocations_familiales/epilogue.catala_fr",
]

[[module]]
name = "Prestations_familiales"
module_uses = [
  "France",
  "Smic",
]
includes = [
  "prestations_familiales/prologue.catala_fr",
  "prestations_familiales/sécurité_sociale_L.catala_fr",
  "prestations_familiales/securite_sociale_R.catala_fr",
]
```

Question: modules are already declared in catala files using the `>
Module <name>` directive. Should we force every module to have its own
entry in the TOML or should we leave these implicit declarations and
let clerk work its magic?

### Targets

Targets are a project's building endgoal. These are used to combine
modules, create an abstraction barrier and provide Catala ready-to-use
components.

For example, one might want to create a specific component to
distribute it to a mainframe that computes the "allocation
familiales". Hence, one would declare:

```toml
[[target]]
name = "mainframe"
entrypoints = [ "Allocations_familiales" ]
[...]

PR #712, opened 2024-10-03
Add TOML encoding

vincent-botbol:
This PR cleans the clerk's TOML config ad hoc reader/writer by adding a generic type-safe encoder/decoder.
This allows to properly describe and update config files without much boilerplate. C.f. the second commit that replaces the existing parsing with this encoder.

The PR adds new fields to the config file (related to #692) that are currently unused but are retro-compatible with the existing TOML configuration files (in particular, the one present in https://github.com/CatalaLang/catala-examples).

It also displays proper error messages, for example:
image
or
image

### DIFF
--- a/build_system/clerk_toml_encoding.ml
+++ b/build_system/clerk_toml_encoding.ml
@@ -1,0 +1,536 @@
+(* This file is part of the Catala build system, a specification language for
+   tax and social benefits computation rules. Copyright (C) 2024 Inria,
+   contributors: Louis Gesbert <louis.gesbert@inria.fr>
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you may not
+   use this file except in compliance with the License. You may obtain a copy of
+   the License at
+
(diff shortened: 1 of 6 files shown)


denismerigoux (approved):
This is an excellent PR! I <3 the error messages. The code in `clerk_config` is maybe too type-safe with heavy use of combinators that might be hard to read for OCaml newbies, but let's assume `clerk` will be maintained by seasoned OCaml devs :)

build_system/clerk_toml_encoding.ml
```diff
@@ -0,0 +1,536 @@
+(* This file is part of the Catala build system, a specification language for
+   tax and social benefits computation rules. Copyright (C) 2024 Inria,
+   contributors: Louis Gesbert <louis.gesbert@inria.fr>
```
AltGr, line 3: Wrong ;)

AltGr (approved):
This very nice!

Issue #647, opened 2024-07-09
Add a list folding surface syntax 


denismerigoux:
Right now, many of our syntactic list manipulation operators are implemented underneath using `fold`. However, the user cannot directly write a `fold` in the Catala surface syntax. The CNAF expressed the need for such an operator to be exposed. As it is polymorphic, it will require special syntax.

The signature for this operator would be `fold : 'a list -> ('b -> 'a -> 'b) -> 'b -> 'b`.

AltGr:
One possible syntax:
Since we already have `<foo(v)> for <v> among <list>` for map
one possibility would be to extend that with details: `<foo(v,acc)> for <v> among <list> accumulating into <acc> starting from <init>`
(the `starting` could be removed for conciseness... maybe the `into` (or even the `accumulating` ?) as well ?)

E.g. Compute the sum of a list of money with:
```
x + y for x among revenues accumulating into y starting from $0
```

denismerigoux:
Decision of the syntax committee : 


```catala
combine <accumulator> initially <initial_value> with <expression depending on accumulator and element> for <element> among <list>

combinaison de <accumulateur> initialement <valeur initiale> avec <expression dépendant de valeur initiale et élément> pour <élément> parmi <liste>
```

denismerigoux:
@vincent-botbol from the above you can: 

* modify the parser and lexer in https://github.com/CatalaLang/catala/tree/master/compiler/surface accordingly
* modify your topyari formatter and LSP extension to take it into account
* update all syntax highlighting plugins.

Thanks!

PR #715, opened 2024-10-09
Add list folds to the surface syntax

AltGr:
Temporary syntax:
```
for a initially 0 and then (a + x) for x among lst
```

and in French:
```
pour a initialement 0 puis (a + x) pour x parmi lst
```

to be elaborated with the Syntax Committee of course. Two new keywords: `initially` and `and then` (it could have been the existing `then` keyword for english, relying on the ambiguity of the word, but that's not portable across languages).
Then added to the cheat-sheet and docs.

Note that detuplification takes place, making it possible both to fold multiple lists in parallel, and to use multiple accumulators. See the tests for examples.


AltGr:
I wonder if non-programmers will be able to make sense of this. Maybe the following conveys the meaning better ?
```
for a initially 0 and then equals (a + x) for x among lst
```

AltGr:
Closes #647

tests/array/good/fold.catala_en
```diff
@@ -1,35 +1,82 @@
...
+declaration insert content list of money
+  depends on element content money,
+             lst content list of money
+  equals
+  let (lower, higher) equals
+    for (lower, higher) initially ([], []) and then
+      (if x <= element then (lower ++ [x], higher) else (lower, higher ++ [x]))
+      for x among lst
+  in
+  lower ++ [element] ++ higher
```
denismerigoux, line 33: HAHAHA that's it we can program insertion sort in Catala, we're doomed !

denismerigoux (approved):
OK for merging this now, but syntax of fold should be on the agenda of the Catala meeting of Oct 22nd!

tests/array/good/fold.catala_en
AltGr: My thoughts exactly when I was writing it ;)

denismerigoux:
@AltGr did the syntax committee agree with this syntax? Should it be changed?

PR #718, opened 2024-10-10
C backend: workaround bulk freeing of constants

AltGr:
Constants may need initialisation and allocations, which we do lazily, but that
doesn't combine well with the way our allocator works, freeing everything in
bulk.

This patch adds a switch to the allocator that make it use raw `malloc`, which
is turned on within the initialisation functions of constants.


AltGr:
(note: this is broken as it is, the catala allocator restoration command is inserted after `return`)

AltGr:
Once we have functions to promote values allocated within the catala heap to persistent memory, these should be leveraged instead.

PR #719, opened 2024-10-10
Make the explain plugin compatible with vscode test-case extension

vincent-botbol:
This PR adds an `--inline-mod-uses` option to the `explain` plugin so that it's able to handle tests files generated by the test-case plugin.

This option detects module usage in the given file and tries to inline them with the source file (that is currently supposed to be in the same directory in vscode). It also removes the `> Module` declaration which would prevent them from being inlined. This is a temporary workaround until the clerk build system gets mature enough to handle these kind of situations.

I also added a `--line-format` option to be able to output vscode-like links. E.g.,
```bash
$ catala explain ~/catala-language-server/test-case-parser/examples/test_case_apl.catala_en -s Exemple1_test\
                --html --url-base vscode://file --line-format ':NN:0' --inline-mod-uses > test.html
```
will output vscode links like : `vscode://file/path/to/file.catala_en:123:0` which will give the ability to jump to the position in vscode.

### DIFF
--- a/compiler/plugins/explain.ml
+++ b/compiler/plugins/explain.ml
@@ -25,6 +25,7 @@ type flags = {
   show : string option;
   output : Global.raw_file option;
   base_src_url : string;
+  line_format : string;
 }
 
 (* -- Definition of the lazy interpreter -- *)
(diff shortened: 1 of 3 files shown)


denismerigoux (approved):
That's amazing if it works with @rprimet's UI!!!

rprimet:
Maybe we could keep this PR open but already merge its contents in the branch cnaf-demo?

vincent-botbol:
The last commit could be removed, it's probably a bad idea in the general case.

PR #726, opened 2024-10-16
Sub_date_dur should also generate a source position in OCaml!

denismerigoux:


AltGr:
Ha well spotted, indeed I added the location to this operator recently and must have overlooked the OCaml backend.

Let's have the complete testsuite applied to all backends ASAP! ;)

AltGr (approved):

PR #733, opened 2024-10-28
Mostly graphical improvements to the explain plugin

AltGr:


compiler/plugins/explain.ml
```diff
@@ -1361,11 +1433,20 @@ let to_dot lang ppf ctx env base_vars g ~base_src_url ~line_format =
...
+      expr_to_dot_label theme env ctx lang ppf e
     (* ; * Format.pp_print_flush ppf (); * Format.pp_set_formatter_out_functions
        ppf out_funs *)
 
-    let graph_attributes _ = [ (* `Rankdir `LeftToRight *) ]
+    let graph_attributes _ =
+      [
+        `BgcolorWithTransparency (Int32.of_int 0x00);
+        (* `Ratio (`Float 0.8); *)
+        (* `Concentrate true; *)
```
denismerigoux, line 1444: Should these be toggable with an option?

compiler/scopelang/from_desugared.ml
```diff
@@ -863,7 +863,7 @@ let translate_program
                   match states with
                   | D.WholeVar -> WholeVar (ScopeVar.fresh (var_name, var_pos))
                   | States states ->
-                    let var_prefix = var_name ^ "_" in
+                    let var_prefix = var_name ^ "#" in
```
denismerigoux, line 866: Are we sure this gets correctly changed back to `_` in the C or Python backends?

denismerigoux (approved):
Amazing!

compiler/scopelang/from_desugared.ml
AltGr: Yes, that would otherwise be a bug!

compiler/plugins/explain.ml
AltGr: Hm I was playing around trying to improve the layout, but from my tests they don't really provide meaningful improvements ; I certainly should cleanup all these comments though.

compiler/scopelang/from_desugared.ml
AltGr: Re: indeed, there was a bug where this would appear in the backend language (but it was a symptom of a wider bug where we would cleanup unicode characters but not non-ident characters that were part of ASCII like `é`).
The latest patch fixes that for all backends (just fixing the normalisation functions used by the `Renaming` module).

PR #734, opened 2024-10-28
Carry over binders positions in EAbs

vincent-botbol:
This PR augments the EAbs's node with the position of binded variables. This allows the LSP server to handle such variables and references them for the user which was not previously possible.


denismerigoux (approved):
Yeah, this looks very nice! I'm curious to discuss tomorrow how you found a solution that only adds positions to the binders and not the bound variables lower in the AST.

AltGr (approved):
Perfect, thanks!

PR #735, opened 2024-10-29
Fix parser conflict on `combine x initially output of y with`

### LINKED ISSUE #731: Fix shift/reduce introduced by the new fold syntax
The PR #730 introduces a shift/reduce conflict in the parser. It needs to be fixed later on (by someone smarter, looking at you @AltGr :))

Don't forget that the last commit of #730 removes conflicting error messages which have to also be reintroduced.

AltGr:
Closes #731


vincent-botbol (commented):
Nice, well done

PR #743, opened 2024-11-04
Add warning notification and add utility function to passes

vincent-botbol:
This PR plugs warnings notification mechanism that's currently used for errors messages.

~It also refactors the driver's passes so that it offers more control which is needed in LSP.~

~We also add a global option to handle include resolving introduced by #688~
Edit: I don't need this rework as of right now and I'm still not sure that it's sufficient for what I want to achieve so I removed it from the PR.

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -362,7 +362,7 @@ open Content
 exception CompilerError of Content.t
 exception CompilerErrors of Content.t list
 
-type lsp_error_kind = Lexing | Parsing | Typing | Generic
+type lsp_error_kind = Lexing | Parsing | Typing | Generic | Warning
 
 type lsp_error = {
   kind : lsp_error_kind;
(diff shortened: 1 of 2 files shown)


compiler/catala_utils/message.ml
```diff
@@ -459,6 +458,26 @@ let join_pos ~pos ~fmt_pos ~extra_pos =
     Some pos
   | _ -> None
 
+let warning
```
denismerigoux, line 461: Here you add a warning function but you do not use it in the rest of the compiler codebase in this PR? There are already some warnings thrown by the compiler (for instance in the linting module), I would have expected this PR to reactor them so that they're taken into account correctly by the LSP.

denismerigoux (approved):
Thanks! I let @AltGr give his opinion about the refactor of driver.ml

compiler/catala_utils/message.ml
vincent-botbol: We already have an existing generic processing for messages to convert them into a reasonable format for LSPs. We can tweak it if needed but from what I saw it seemed to output a decent message:
image

vincent-botbol: To make it clearer, every existing warning in the code base (e.g. the linter) will go through this function which will notify the message handler (for which we defined a hook in the LSP) that returns a printable content that was preprocessed to display a "vscode-friendly" message (because we also set the formatting mode to LSP).

PR #751, opened 2024-11-26
Tweaks and fixes to the C backend

AltGr:
This improves renaming of idents in the C backend, and fixes a bug in the passing of the visibility information from the surface AST


vincent-botbol (approved):
Looks good to me

PR #752, opened 2024-12-10
Compatibility with LSP server build on different platforms

denismerigoux:
Windows doesn't like rogue shell invocations. This change fixes this.


AltGr (approved):
This looks fine!

PR #753, opened 2024-12-12
Add error handler hook in C runtime

vincent-botbol:
This PR adds a naive mechanism to the C backend to allow the user to handle runtime errors  without crashing the program.


denismerigoux (approved):
Very good!

PR #754, opened 2024-12-16
Format trace to JSON format

denismerigoux:
Based on @rprimet's, excellent suggestion, we want to print the trace by the interpreter in a JSON format that could be exploitable by debug tools. However, the parsing of raw events to events is broken now and needs to be fixed. In the meantime, this PR prints the list of raw events on the standard output in JSON format.

### DIFF
--- a/runtimes/ocaml/runtime.ml
+++ b/runtimes/ocaml/runtime.ml
@@ -380,6 +380,17 @@ module BufferedJson = struct
     Printf.bprintf buf {|,"fun_inputs":[%a]|} (list var_def) fc.fun_inputs;
     Printf.bprintf buf {|,"body":[%a]|} (list event) fc.body;
     Printf.bprintf buf {|,"output":%a}|} var_def fc.output
+
+  and raw_event buf = function
+   | BeginCall name -> Printf.bprintf buf {|{"event": "BeginCall", "name": "%s"}|} (String.concat "." name)
+   | EndCall name -> Printf.bprintf buf {|{"event": "EndCall", "name": "%s"}|} (String.concat "." name)
+   | VariableDefinition (name, _io, _value) ->
+       Printf.bprintf buf {|{
+         "event": "VariableDefinition",
+         "name": "%s",
+         "value": "TODO"
+         }|} (String.concat "." name) 
+   | DecisionTaken _dectaken -> Printf.bprintf buf {|DecisionTaken|}
 end
 
 module Json = struct
--- a/runtimes/ocaml/runtime.ml
+++ b/runtimes/ocaml/runtime.ml
@@ -338,7 +338,7 @@ module BufferedJson = struct
   (* Note: the output format is made for transition with what Yojson gave us,
      but we could change it to something nicer (e.g. objects for structures) *)
   let rec runtime_value buf = function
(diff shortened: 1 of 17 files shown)


rprimet:
Note : it is possible that we are not outputting valid json yet (for instance, numbers are a footgun in json, and it seems that numbers with an unset fractional part are not accepted : e.g. `1.0` is valid but `1.` isn't -- that should probably be an easy fix)

runtimes/ocaml/runtime.ml
```diff
@@ -317,26 +317,51 @@ module BufferedJson = struct
...
+    let rec last_non0 n =
+      if n <= 1 || dec_str.[n - 1] <> '0' then n else last_non0 (n - 1)
+    in
+    Buffer.add_substring buf dec_str 0 (last_non0 max_decimals)
+
   (* Note: the output format is made for transition with what Yojson gave us,
      but we could change it to something nicer (e.g. objects for structures) *)
   let rec runtime_value buf = function
-    | Unit -> Buffer.add_string buf {|"Unit"|}
+    | Unit -> Buffer.add_string buf {|{}|}
```
AltGr, line 341:
```suggestion
    | Unit -> Buffer.add_string buf "{}"
```

AltGr (approved):
LGTM! An area that could be improved is the UI: we have 3 different flags for traces, and specifying an output file for example doesn't imply that the trace should be on.
Ideas:
- make the trace output file be an optional argument of `--trace`, so that no extra flag is needed (note though: it could break existing scripts that do e.g. `catala interpret -t foo.catala_en` because `foo.catala_en` would be taken as the argument of `-t` ; we don't have that many scripts around and I still think it's ok, though)
- add environment variables for the flags (in particular, that could be useful for the trace format)
- infer the trace format from the output file name if `.json`

This may require some habit of Cmdliner so I can help ^^

rprimet:
Adding the output file as an option of the `--trace` argument appears to have broken some scripts ( https://github.com/CatalaLang/catala/actions/runs/12827450876/job/35769769947?pr=754 ) ; not sure what the best course of action is, reinstate a `--trace-output` flag or fix the invocation in the scripts using `--` ?

AltGr:
yes that's the caveat I pointed to before: we should fix the scripts, either by expliciting `-t -` or by just moving the flag so that it doesn't appear right before the catala file name.
When we automatically handle the `.json` file extension, we should also add a warning pointing the correct course of action if we detect that the supposed trace file ends with `.catala_??` (or even an error, to be totally certain that a catala file won't get overwritten!)

rprimet:
Now that the catala-example CI has a candidate fix ( https://github.com/CatalaLang/catala-examples/pull/19 ), the main issue remaining in the CI is that some messages have a different format, due to a few issues of `Message.log` being replaced by straight `Format.fprintf` in the code of `print_log` (see for instance `git diff HEAD..master compiler/shared_ast/interpreter.ml`) ; as a result, there are a few line break inconsistencies and missing [LOG] prefixes. @AltGr any ideas on what may be the best fix, parametrize Message so that the ppf instance can be user-supplied or something else?

AltGr:
I wrote a workaround so that the format, when printing to the console, is the same as before (when in a file the `[LOG] ` prefixes are removed, as they should be)

Issue #755, opened 2024-12-19
Bad position in exception conflict


AltGr:
I got a very confusing error on two conflicting exceptions, where the positions reported didn't match the actual issue. I failed to reproduce for now on a simplified example, but it can be seen on https://github.com/CatalaLang/poc-cnaf/tree/debugConflictPos with `clerk test tests/fonctionnels/cas_1.catala_fr`

In more detail, on the follwing piece of code:
```catala
champ d'application Barèmes:
  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Isolé et
                   rôle sous forme PersonneÀCharge de n et n = 1
    conséquence égal à 50%

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Isolé et
                   rôle sous forme PersonneÀCharge de n et n = 2
    conséquence égal à 30%  # <=== position 1

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Isolé et
                   rôle sous forme PersonneÀCharge de n et n > 2
    conséquence égal à 40% # <=== position 2

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Couple et
                   rôle sous forme Conjoint
    conséquence égal à 50%

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Couple et
                   rôle sous forme PersonneÀCharge de n et n = 1
    conséquence égal à 30%

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Couple et
                   rôle sous forme PersonneÀCharge de n et n = 2
    conséquence égal à 30%

  étiquette base
  définition taux_majoration_montant_forfaitaire
      de isolement, rôle
    sous condition isolement sous forme Couple et
                   rôle sous forme PersonneÀCharge de n et n > 1
    conséquence égal à 40%
```

we get:
```
┌─[ERROR]─
│
│  During evaluation: conflict between multiple valid consequences for assigning the same variable.
│
├─➤ src/rsa.catala_fr:348.24-348.27:
│     │
│ 348 │     conséquence égal à 30%
│     │                        ‾‾‾
[...]

PR #756, opened 2024-12-20
Test and fix bad positions in conflict message

AltGr:
Related to #755, which this fixes *for the lcalc interpreter* ;
however, the real issue is the OCaml (and other) backends, where it's much harder.

Indeed, for those, the positions are passed as a flat list to the top-level default term, which has no information anymore about nested exceptions. The only reliable way I see to fix that is to switch evaluated default terms from `'a option` to `('a * pos) option`, and remove the the flat list of positions. It requires a change in the type of the `handle_exceptions` primitive, but it shouldn't be too difficult now that positions can be reified.

### DIFF
--- a/compiler/lcalc/to_ocaml.ml
+++ b/compiler/lcalc/to_ocaml.ml
@@ -408,6 +408,7 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
   | EFatalError er ->
     Format.fprintf fmt "raise@ (Runtime_ocaml.Runtime.Error (%a, [%a]))"
       Print.runtime_error er format_pos (Expr.pos e)
+  | EPos p -> format_pos fmt p
   | _ -> .
 
 (* TODO: move [embed_foo] to [Foo.embed] to protect from name clashes *)
--- a/compiler/lcalc/to_ocaml.ml
+++ b/compiler/lcalc/to_ocaml.ml
@@ -396,19 +396,8 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
         | Div_int_int | Div_rat_rat | Div_mon_mon | Div_mon_int | Div_mon_rat
         | Div_dur_dur ->
           Format.fprintf ppf "%a@ " format_pos (Expr.pos (List.nth args 1))
-        | HandleExceptions ->
-          let excs =
-            match args with [(EArray ex, _)] -> ex | _ -> assert false
-          in
-          Format.fprintf ppf "[|@[<hov>%a@]|]@ "
-            (Format.pp_print_list
-               ~pp_sep:(fun ppf () -> Format.fprintf ppf ";@ ")
-               format_pos)
-            (List.map Expr.pos excs)
         | _ -> ())
-      (Format.pp_print_list
-         ~pp_sep:(fun fmt () -> Format.fprintf fmt "@ ")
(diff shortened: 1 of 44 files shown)


AltGr:
(note that, although the test triggers it with the merged same-output exceptions, the issue would exist with any nested exceptions)

denismerigoux (approved):
I agree with this fix as well as the proposed generalized solution to change the accumulator type from `'a option` to `('a, pos) option` in the OCaml backend (and `runtime.ml` subsequently).

denismerigoux:
Can I merge this or do you want to put the refactoring of the runtime positions in default in this PR?

AltGr:
Hm, since ① this doesn't solve the issue where it is problematic (the backends) but only in the `lcalc` interpreter which the users shouldn't care about and ② the "real" fix should supersede this anyway, I don't think it's worth merging (although I'd like to keep it open for the testcase that is added)

AltGr:
Update: completely rewritten to reify positions and forward them up the stack of exceptions, ensuring proper conflict positions in all backends.

This got more involved than expected ;)
The web_api plugin is broken though, it generates bad code due to the change in the return type of default terms ; the error message is horrendous and I am not familiar with that code 😬

AltGr:
(still, marking as ready for review since the rest is in reviewable state)

compiler/lcalc/to_ocaml.ml
```diff
@@ -396,18 +396,21 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
...
-            (List.map Expr.pos excs)
+        (* | HandleExceptions ->
+         *   let excs =
+         *     match args with [(EArray ex, _)] -> ex | _ -> assert false
+         *   in
+         *   Format.fprintf ppf "[|@[<hov>%a@]|]@ "
+         *     (Format.pp_print_list
+         *        ~pp_sep:(fun ppf () -> Format.fprintf ppf ";@ ")
+         *        format_pos)
+         *     (List.map Expr.pos excs) *)
```
denismerigoux, line 407: Why is this commented? And below too?

tests/backends/output/simple.c
denismerigoux, line 1: This doesn't help with code readability but otherwise it's correct, so OK.

denismerigoux (approved):
Thanks @AltGr ! About the breaking of the module generating the `js_of_ocaml` API : OK for this PR to be merged, but can you file an issue about us repairing the whole toolchain to generate JS executables later?

compiler/lcalc/to_ocaml.ml
AltGr: Ah, well spotted, this is a leftover and should be removed: the HandleExceptions operator used to need a specially crafted list of exceptions, but now that is all handled at runtime so no special treatment is required anymore.

AltGr:
> Thanks @AltGr ! About the breaking of the module generating the `js_of_ocaml` API : OK for this PR to be merged, but can you file an issue about us repairing the whole toolchain to generate JS executables later?

I'll investigate some more, there might be a simple fix ; if no indeed I may leave it for later.

AltGr:
This should now be fixed: there was an unrelated bug in web_api related to tuples, that only got triggered now that tuples commonly appear in scope interfaces.

This will require the merge of the sister PR https://github.com/CatalaLang/catala-examples/pull/22 that adjusts `oracles.ml` for the new `handle_exceptions` interface

Issue #699, opened 2024-09-25
List available targets for clerk


denismerigoux:
`clerk --help` and `clerk --list-targets` should display a nice man page with the available build, run and test targets contextualized with the current project.

denismerigoux:
No need to list all subommands aof `clerk build`, we just want the basic stuff customized for the project.

PR #766, opened 2025-01-08
Clerk: list targets on `clerk build` without target

AltGr:
This is a partial quickfix for #699; at least, this would be usable for shell completion after `clerk build`


AltGr:
Ah, seems the patch was already merged outside of the PR; this has cosmetic iimprovements over the original version, so should still be merged once CI passes

PR #768, opened 2025-01-08
Fix confusing disambiguation errors appearing instead of type errors

### LINKED ISSUE #749: Failures to typecheck during disambiguation / any types
While working on a complex bit of code, it happened to me that typing failed during the disambiguation phase; the code indeed didn't type, but that resulted in a very unhelpful message.

I didn't have time to debug then so just resorted to defaulting `fail_on_any` to `false` in typing.ml:386, which pushed the failure to the typing phase and got me a better message ; but this requires further investigation.

* we likely don't want `fail_on_any` during disambiguation ; but for that we must consider what to do if an `any` actually hinders the processing of the disambiguation.
* while this flag has proved extremely useful for debugging, it makes the typing more strict than needed: `any` types may be valid if embedded or unused. For example, it fails the following (legitimate) code: `let x equals [] in []`. It should be sound to disable it altogether — but we should find other means to check that there are no gaping holes in the internal (re-)typings.

AltGr:
Closes #749 ;
the PR also has a patch that greatly improves real disambiguation errors when they happen.

The issue was actually with the way we attempt to gather multiple typing errors: this can get us in an inconsistent state after the first error, and any non-delayed exception raised at that point (including `internal inconsistency` compiler errors) would break out... and hide all delayed exceptions.

This is visible in test  `array/bad/type_error_in_filter.catala_en` which would raise a `Cannot resolve typing` error because assumptions of the typer are no longer guaranteed.

The behaviour implemented here is to print the errors that have been delayed, and actually hide the non-delayed error instead (it seems safer to assume, in general, that it is a consequence of an inconsistent state, rather than the opposite. We could put it back when in debug mode if we ever find cases where it would have been useful).

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -531,19 +531,22 @@ let with_delayed_errors
       "delayed error called outside scope: encapsulate using \
        'with_delayed_errors' first");
   global_errors.stop_on_error <- stop_on_error;
-  try
-    let r = f () in
-    match global_errors.errors with
-    | None -> error ~internal:true "intertwined delayed error scope"
-    | Some [] ->
-      global_errors.errors <- None;
-      r
-    | Some [err] ->
-      global_errors.errors <- None;
-      raise (CompilerError err)
-    | Some errs ->
-      global_errors.errors <- None;
-      raise (CompilerErrors (List.rev errs))
-  with e ->
+  let result =
+    match f () with
+    | r -> fun () -> r
+    | exception (CompilerError _ as e) ->
+      let bt = Printexc.get_raw_backtrace () in
+      fun () -> Printexc.raise_with_backtrace e bt
+    | exception e -> raise e
+  in
+  match global_errors.errors with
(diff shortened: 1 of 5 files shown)


compiler/catala_utils/message.ml
```diff
@@ -531,19 +531,22 @@ let with_delayed_errors
...
+  | Some [] ->
+    global_errors.errors <- None;
+    result ()
+  | Some [err] ->
+    global_errors.errors <- None;
+    raise (CompilerError err)
+  | Some errs ->
     global_errors.errors <- None;
-    raise e
+    raise (CompilerErrors (List.rev errs))
```
denismerigoux, line 552: This could use a little comment to understand what's going on :)

denismerigoux (approved):

vincent-botbol (approved):
Indeed, sorry about that..

AltGr:
@vincent-botbol haha don't worry you certainly have as much on my account ;)

PR #770, opened 2025-01-10
Handle carriage returns in the lexer

vincent-botbol:
This PR allows the lexer to handle catala files that contains carriage-returns alongside line feeds. Currently, on windows, such files fails to parse entirely. 
Following @AltGr suggestions, patching the lexer seems to be an easy fix. However, I'm not fully confident in this patch as there might be side-effects that I'm not aware of (e.g., lexing performances).

### DIFF
--- a/compiler/surface/lexer.cppo.ml
+++ b/compiler/surface/lexer.cppo.ml
@@ -358,6 +358,12 @@ let hspace = [%sedlex.regexp? Sub (white_space, Chars "\n\r")]
 (** Operator explicit typing suffix chars *)
 let op_kind_re = [%sedlex.regexp? "" | MR_MONEY_OP_SUFFIX | Chars "!.@^"]
 
+(** Regexp matching every character except newlines *)
+let any_but_eol = [%sedlex.regexp? (Sub (any, Chars "\n\r"))]
+
+(** Regexp matching newlines *)
+let eol = [%sedlex.regexp? Opt '\r', '\n' ]
+
 let op_kind = function
   | "" -> Ast.KPoly
   | "!" -> Ast.KInt
@@ -376,7 +382,7 @@ let rec lex_code (lexbuf : lexbuf) : token =
       (* Whitespaces *)
       L.update_acc lexbuf;
       lex_code lexbuf
-  | '#', Star (Compl '\n'), '\n' ->
+  | '#', Star any_but_eol, eol ->
       (* Comments *)
       L.update_acc lexbuf;
       lex_code lexbuf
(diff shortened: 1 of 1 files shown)


compiler/surface/lexer.cppo.ml
```diff
@@ -772,11 +773,13 @@ let lex_raw (lexbuf : lexbuf) : token =
         (* Nested match for lower priority; `_` matches length 0 so we effectively retry the
            sub-match at the same point *)
         match%sedlex lexbuf with
+        | Star (Compl '\r'), ("\r\n" | eof)
         | Star (Compl '\n'), ('\n' | eof) -> LAW_TEXT (Utf8.lexeme lexbuf)
```
AltGr, line 777: Maybe strip the `\r` from the lexbuf in the first case, so that we consistently get strings without it for law text ?

AltGr: Hm I'm afraid if you have a stray `\r` further down in your file this could cause everything until then to be gobbled by the first regexp...
I guess we need to expand the regexp complement here ? `Star (Plus (Compl "\r\n"), Opt ('\r', Compl '\n')), Opt '\r', '\n'` or something like that ?

AltGr (changes requested):

compiler/surface/lexer.cppo.ml
vincent-botbol: Hmm `Compl` only accepts single characters though. I'll figure something out.

vincent-botbol: Wouldn't this be sufficient ?
```
Star (Compl '\r' | Compl '\n'), Opt ('\r'), ('\n' | eof)
```

AltGr: You want `Compl "\r\n"` (or however this can be written in sedlex syntax), `(Compl '\r' | Compl '\n')` would include anything.
Otherwise, the difference is that this would forbid `\r` that appear outside of a newline (or just before eof)... but I guess that's actually fine with us, indeed.

vincent-botbol: Indeed.. I must have been tired on friday. I pushed a new version. I encoded `Compl "\r\n"` as `any_but_eol = Sub (any, "\n\r")`  which should be semantically equivalent. Otherwise, I think that's reasonable to forbid `\r` chars outside of newlines but we could also allow them. I have no strong opinions. I forbid them in this new version.

AltGr:
@vincent-botbol is this ready for merge ?

PR #772, opened 2025-01-10
Stop using String token in pygments lexers

### LINKED ISSUE #589: Pygments lexer breaks LaTeX escapeinside
I'm trying to use the Pygments lexer in `syntax_highlighting/en/pygments`.
With Pygments version 2.16.1 (I haven't tried other versions, but this is close to being the latest), running `pygmentize -l 'catala_en' -f latex -P escapeinside='!!' example.catala_en` on the following file:
````
```catala
declaration scope A:
  input x content integer!\label{line:x}!
```
````
renders the following:
````tex
\begin{Verbatim}[commandchars=\\\{\},codes={\catcode`\$=3\catcode`\^=7\catcode`\_=8\relax}]
\PY{l+s}{```catala}
\PY{k+kr}{declaration}\PY{l+s}{ }\PY{k+kr}{scope}\PY{l+s}{ }\PY{n+nc}{A}\PY{o}{:}
\PY{l+s}{ }\PY{l+s}{ }\PY{k+kd}{input}\PY{l+s}{ }\PY{n+nv}{x}\PY{l+s}{ }\PY{k+kr}{content}\PY{l+s}{ }\PY{k+kt}{integer}\PY{err}{!}\PY{l+s}{\PYZbs{}}\PY{k+kr}{label}\PY{o}{\PYZob{}}\PY{n+nv}{line}\PY{o}{:}\PY{n+nv}{x}\PY{o}{\PYZcb{}}\PY{err}{!}
\PY{l+s}{```}
\end{Verbatim}
````
The important part here being that the escaped LaTeX code is rendered (line 4) as `\PY{err}{!}\PY{l+s}{\PYZbs{}}\PY{k+kr}{label}\PY{o}{\PYZob{}}\PY{n+nv}{line}\PY{o}{:}\PY{n+nv}{x}\PY{o}{\PYZcb{}}\PY{err}{!}`. Instead, this should be `\PY{esc}{\label{line:x}}` so that it is indeed escaped when using eg the `minted` LaTeX package.

Given that [Pygment's doc](https://pygments.org/docs/formatters/#LatexFormatter) states that the `escapeinside` option has "no effect in string literals", I would suspect that some uses of the `String` token in `lexer.py` may be responsible for this behavior.

pierregoutagny:
To fix #589, I remove any use of the `String` pygments token from the lexers, that was used extensively before. They are replaced by  `Whitespace` or the generic `Text` where appropriate.

More manual testing is required: I had to update one unit tests to pass, but I don't know in what context the generated LaTeX code is supposed to be rendered, so I can't say if something is broken graphically. Similarly, I have issues compiling the `syntax.pdf` file on my computer (with or without the fix) so I can't check if it is broken.


denismerigoux (approved):
Thanks @pierregoutagny !

PR #773, opened 2025-01-15
Revert "Stop using String token in pygments lexers"

denismerigoux:
Reverts CatalaLang/catala#772 : the fix completely breaks at least the generation of the Catala syntax sheet cheat (see attached) : 

Capture d’écran du 2025-01-15 15-19-21


denismerigoux:
@pierregoutagny I'm sorry your fix was not the right solution it seems...

pierregoutagny:
Oh sorry for this break. I have trouble compiling the syntax.pdf file locally (it looks like your screenshot no matter the version), but my guess would be that if you do the same as for the [unit test](https://github.com/CatalaLang/catala/pull/772/commits/1696a9b9db868abb12aed5a51427c076411bf480#diff-3b80b027422d51c70af8c1cf14891621219b64a00f85195168c6b4be2bc0de86L83) it could work. For instance, could you check if the pdf works when you remove the `\PYG` on [lines 201 and 202](https://github.com/CatalaLang/catala/blob/d4ee59cdc4d5a453dde7ad7c01e0d9af6712132d/doc/syntax/syntax_en.tex#L201-L202) of `syntax_en.tex` ? That is replacing
```tex
\newcommand*\FancyVerbStartString{\PYG{l+s}{```catala}}
\newcommand*\FancyVerbStopString{\PYG{l+s}{```}}
```
by
```tex
\newcommand*\FancyVerbStartString{```catala}
\newcommand*\FancyVerbStopString{```}
```

PR #775, opened 2025-01-20
Add type variables and quantifiers to the AST

AltGr:
Note that they don't do much at the moment, basically this replaces `TAny` placeholders with type variables ; the translation to and from unionfind types is done on the boundaries of the typer. Structs and enums don't support quantifiers yet.

The type printer is also simplified and made less verbose by removing the `_debug` functions and not repeating the details of structs and enum every time.


AltGr:
Alright, this is now mostly ready.

Final touches remain todo:
- add the tests (I have some, but not pushed yet)
- improve printing of types with vars
- update syntax highlighting
- document

compiler/driver.ml
```diff
@@ -1001,7 +999,7 @@ module Commands = struct
     let prg, type_ordering, _ =
       Passes.lcalc options ~includes ~optimize ~check_invariants ~autotest
         ~typed:Expr.typed ~closure_conversion ~keep_special_ops:true
-        ~monomorphize_types:false ~expand_ops:true
+        ~monomorphize_types:false ~expand_ops:false
```
denismerigoux: Why did we stop expanding ops for the OCaml backend?

compiler/driver.ml
```diff
@@ -275,7 +275,7 @@ module Passes = struct
     let prg =
       if not closure_conversion then (
         Message.debug "Retyping lambda calculus...";
-        let prg = Typing.program ~fail_on_any:false ~internal_check:true prg in
+        let prg = Typing.program ~internal_check:true prg in
```
denismerigoux, line 278: Why do we switch the `fail_on_any` for the LCalc pass?

compiler/shared_ast/definitions.ml
```diff
@@ -224,7 +224,9 @@ and naked_typ =
   | TOption of typ
   | TArray of typ
   | TDefault of typ
-  | TAny
+  | TVar of naked_typ Bindlib.var
+  | TAny of (naked_typ, typ) Bindlib.mbinder
```
denismerigoux: You can rename this `TLambda` to stick to classical SystemF vocabulary?

compiler/shared_ast/operator.ml
```diff
@@ -567,7 +567,7 @@ let resolve_overload_aux (op : overloaded t) (operands : typ_lit list) :
     raise Not_found
 
 let resolve_overload
-    ctx
+    _ctx
```
denismerigoux: Remove unused parameter?

compiler/shared_ast/type.ml
```diff
@@ -90,55 +75,183 @@ let rec compare ty1 ty2 =
   | _, TArray _ -> 1
   | TDefault _, _ -> -1
   | _, TDefault _ -> 1
-  | TClosureEnv, _ -> -1
-  | _, TClosureEnv -> 1
+  | TVar _, _ -> -1
+  | _, TVar _ -> 1
+  | TAny _, _ -> -1
+  | _, TAny _ -> 1
+(* | TClosureEnv, _ -> -1
```
denismerigoux, line 82: Why leaving this commented?

compiler/shared_ast/type.ml
```diff
@@ -90,55 +75,183 @@ let rec compare ty1 ty2 =
...
+    let hash = Bindlib.hash_var
+    let format = Print.tvar
+  end
+
+  include Arg
+  module Set = Set.Make (Arg)
+  module Map = Map.Make (Arg)
+  module Hashtbl = Hashtbl.Make (Arg)
+
+  let fresh () = Bindlib.new_var (fun v -> TVar v) "'1"
```
denismerigoux, line 102: Why `"'1"`?

compiler/shared_ast/typing.ml
```diff
@@ -18,177 +18,164 @@
...
+    structs : typ StructField.Map.t StructName.Map.t;
+    enums : typ EnumConstructor.Map.t EnumName.Map.t;
+    vars : ('e, typ) Var.Map.t;
+    scope_vars : typ ScopeVar.Map.t;
+    scopes : typ ScopeVar.Map.t ScopeName.Map.t;
+    scopes_input : typ ScopeVar.Map.t ScopeName.Map.t;
+    toplevel_vars : typ TopdefName.Map.t;
+    tvars : Type.t Type.Var.Hashtbl.t;
+    printed_errors : unit Pos.Map.t ref;
+  }
```
denismerigoux, line 37: So now the typer will work on any of the intermediate representations and not just on Dcalc and Lcalc?

compiler/shared_ast/typing.ml
```diff
@@ -18,177 +18,164 @@
...
+
+  let dump ppf env =
+    let pp_sep = Format.pp_print_space in
+    Format.pp_open_vbox ppf 0;
+    (* Format.fprintf ppf "structs: @[<hov>%a@]@,"
+     *   (StructName.Map.format_keys ~pp_sep) env.structs;
+     * Format.fprintf ppf "enums: @[<hov>%a@]@,"
+     *   (EnumName.Map.format_keys ~pp_sep) env.enums;
+     * Format.fprintf ppf "vars: @[<hov>%a@]@,"
+     *   (Var.Map.format_keys ~pp_sep) env.vars; *)
```
denismerigoux: Useful to leave this here for debug?

compiler/shared_ast/typing.ml
```diff
@@ -18,177 +18,164 @@
...
+      "─➤" Type.format ty1 "─➤" Type.format ty2;
+    env.printed_errors :=
+      List.fold_left
+        (fun acc p -> Pos.Map.add p () acc)
+        !(env.Env.printed_errors) pos
 
-let pp_color_string = with_color Format.pp_print_string
+let tvar_witness tvset = Type.Var.Set.max_elt tvset
+(* anonymous type variables start with "'", so this is guaranteed to give us a
+   user-supplied name if any *)
```
denismerigoux: This comment is somewhat cryptic :)

compiler/shared_ast/typing.ml
```diff
@@ -18,177 +18,164 @@
...
-    Format.pp_print_as fmt 1 "⟨";
-    format_typ ~colors fmt t1;
-    Format.pp_print_as fmt 1 "⟩"
-  | TAny v ->
-    if Global.options.debug then Format.fprintf fmt "<a%d>" (Any.id v)
-    else Format.pp_print_string fmt "<any>"
-  | TClosureEnv -> Format.fprintf fmt "closure_env"
+(* `eqclass` gathers all current aliases of the type ; `seen` is other type
+   variables that contain `ty` *)
+let rec get_ty_aux ?(onfreevar = fun _ -> ()) env pos eqclass seen = function
```
denismerigoux: This could use proper type annotations and a comment about what this does

compiler/shared_ast/typing.ml
```diff
@@ -18,177 +18,164 @@
...
-  let open Ocolor_types in
-  blue :: cyan :: green :: yellow :: red :: magenta :: colors
+(* Expands all known type variables as much as possible in the given type *)
+let get_ty env e ty =
+  Bindlib.unbox
+    (get_ty_aux env (Expr.pos e) Type.Var.Set.empty Type.Var.Set.empty ty)
 
-let dummy_flags = { fail_on_any = false; assume_op_types = false }
-let format_typ ctx fmt naked_typ = format_typ ctx ~colors fmt naked_typ
+let get_ty_quantified env pos ty =
```
denismerigoux, line 166: When to use `get_ty_quantified` rather than `get_ty`?

compiler/shared_ast/typing.ml
```diff
@@ -224,54 +211,121 @@ let record_type_error _ctx (A.AnyExpr e) t1 t2 =
...
+    t
+  | TClosureEnv, TClosureEnv -> t2
+  | ( ( TLit _ | TArrow _ | TTuple _ | TStruct _ | TEnum _ | TOption _
+      | TArray _ | TDefault _ | TClosureEnv ),
+      _ ) ->
+    record_type_error ();
+    t2
 
-let lit_type (lit : A.lit) : naked_typ =
+let unify'
```
denismerigoux: Shouldn't this be `unify` and the other `unify'` or better `unify_and_return_repr`?

compiler/shared_ast/typing.ml
denismerigoux, line 1: Global comment: we stopped using `unionFind` here, how do we do to keep track of the equivalence classes of the unification process?

tests/array/bad/type_error_in_filter.catala_en
```diff
@@ -31,11 +31,11 @@ scope Test:
 
 ```catala-test-cli
 $ catala dcalc
-┌─[ERROR]─ 1/2 ─
+┌─[ERROR]─
 │
 │  Error during typechecking, incompatible types:
 │  ─➤ St
-│  ─➤ list of any
+│  ─➤ list of <ty1>
```
denismerigoux: This should read `<type variable 1>`

tests/bool/good/test_bool.catala_en
```diff
@@ -25,7 +25,7 @@ $ catala Typecheck --check-invariants
 
 ```catala-test-cli
 $ catala Dcalc 
-let TestBool : TestBool_in → TestBool =
+let TestBool : #[test] #[test] TestBool_in → #[test] TestBool =
```
denismerigoux, line 28: Should the printer print attributes like this?

tests/io/good/subscope.catala_en
```diff
@@ -34,8 +34,8 @@ $ catala Typecheck --check-invariants
 
 ```catala-test-cli
 $ catala Dcalc -s B
-let scope B (B_in: #[test] B_in): #[test] B =
```
denismerigoux, line 37: Why attributes are sometimes removed like here and sometimes added like above?

tests/tuples/good/letin.catala_en
denismerigoux, line 1: Thanks for the extra test!

tests/typing/bad/linked_vars.catala_en
```diff
@@ -0,0 +1,79 @@
...
+│
+│  This function has type <c> →
+│                           (list of <a>,
+│                            list of <c>,
+│                            list of <c>,
+│                            list of <d>,
+│                            list of <c>,
+│                            list of <c>,
+│                            list of <e>)
+│  which requires that <b> = <c>,
```
denismerigoux, line 41: I like this message!

tests/typing/bad/too_specialised.catala_en
```diff
@@ -0,0 +1,24 @@
...
+    in (min, max) initially (-1, -1)
+  with (if element <= min or min = -1 then element else min, if element >= max then element else max)
+```
+
+
+```catala-test-cli
+$ catala typecheck
+┌─[ERROR]─
+│
+│  In this function definition, the type <t> is specified as anything,
```
denismerigoux, line 15: I like this message!

compiler/driver.ml
AltGr: ah well it's actually unrelated to the PR I think the patch dates from when we were debugging the POC.
but it remains mostly irrelevant to do the expansion for OCaml instead of using the built-in polymorphic equality, I guesss ? If there is any doubt I'll move the patch out to be discussed separately.

compiler/driver.ml
AltGr: That flag no longer makes a lot of sense since "any" as it was before no longer exists, and we don't want to forbid polymorphism here.

compiler/shared_ast/definitions.ml
AltGr: I agree `TAny` might be a little weird in the new context, indeed,  But with `TLambda` I'd be afraid of confusion with the `TArrow` type ? Lambda __at the type level__ vs. type __of a__ lambda...

Maybe `TForAll` or `TQuantified` ?

compiler/shared_ast/type.ml
AltGr: A few reasons (note that I didn't say "good" reasons!):
- the `'` is reminiscent of the syntax of type variables in OCaml
- it's not a valid catala type ident, therefore allowing us to distinguish generated type variables
- the `1` ensures that upon renaming by bindlib the vars gets numbered 1, 2, 3, etc.
- also, `'` is before any valid identifier in lexicographical order, which allows us to favor named type-variables when choosing the representant of a variable equivalence class by just picking the max element.

Also of interest is that upon printing, this will be shown  as `<ty1>`, `<ty2>`, etc., the weird quote char won't appear

tests/array/bad/type_error_in_filter.catala_en
AltGr: Can do ; but do we wan't to expose the notion of type variables to the user ? Since the syntax is `anything of type ty`, maybe `<some type ty1>` or `<some type(1)>` would be more accessible ?

Note that here the type variable is anonymous, but if we had a named type the programmer-supplied name would be reused (within chevrons), like in https://github.com/CatalaLang/catala/pull/775/files/40e030f2efbcc08445bce227b83923b2c0bd2113#diff-66500b36753d56ce55a89631a18780eea0f413a3f63c2d753f3555ba4a34479c for example.

tests/bool/good/test_bool.catala_en
AltGr: hm, that should indeed probably be improved (the changes to the type printer were actually for simplification, I removed the inline expansion of all struct/enum types and the need for the decl_ctx parameter, but that was months ago and I don't remember why we have this difference)

Here the duplication is due to the fact that the in and out structs inherit the position (and attr) from their scope ; and the function type inherits that position as well, so it's actually `#type ( #type a -> #type b )`
We could cleanup these attrs when copying the positions, _or_ make the printer more clever.

compiler/driver.ml
denismerigoux: I changing this in this PR breaks the compilation to C of the current version of `POC-CNAF` then no we should not do it in this PR.

compiler/shared_ast/definitions.ml
denismerigoux: Yes you're right, `TForall` is better.

tests/bool/good/test_bool.catala_en
denismerigoux: I believe that attributes should not be print by default by the printers, they should only come out with the `--debug` option no?

compiler/shared_ast/type.ml
AltGr: Same convention as in `Expr`: for the sake of uniformity, I prefer to keep it here, but uncommenting it would trigger a `the match case is unused` warning.

compiler/shared_ast/typing.ml
AltGr: it already did, we call it as soon as scopelang to resolve overloads.

compiler/shared_ast/typing.ml
AltGr: Yes `Env.dump` is not used anyway, but sometimes useful for debugging... which means there is actually no reason to keep it commented 😅

tests/bool/good/test_bool.catala_en
AltGr: ok

tests/io/good/subscope.catala_en
AltGr: The dcalc is a bit quirky when supplied no scope like in the case above, it'll choose a random scope and print the program after calling `Program.to_expr` on it (hence using the expr let-in printer instead of the scope printer). That's for historical reason and I was afraid to change it at the same time when I refactored the main driver, but I think it can safely be put on par with the lcalc printer now. The removal of the attrs in this case was voluntary.

It may be better to do that change in a follow-up PR to avoid adding noise to the test result changes though ?

compiler/driver.ml
AltGr: This is unrelated to C ? Only the OCaml backend is affected. I'll move it away though.

denismerigoux: Ah yes sorry, rephrasing my previous comment: "If changing this in this PR breaks the compilation to OCaml of the current version of POC-CNAF (and thus the interpreter on multi-modules test runs) then no we should not do it in this PR".

AltGr: It shouldn't break anything, initially it was a workaround for an issue we had on the POC but that was properly fixed in the end. Now wether we want expansion in OCaml is another debate, I think we do in the end because this way we can be more confident in the consistency of its semantics, but with the current implementation we have I'm not so sure ;)

Issue #760, opened 2025-01-07
Fix trace parsing to reconstitute the call graph


denismerigoux:
Suggestion : we should move the trace parsing code from the runtime to a new little executable that parses the trace from a JSON format output by https://github.com/CatalaLang/catala/pull/754 and produces a new JSON that is the parsed trace. Parsing of the trace for big examples should be under CI as it breaks often.

PR #778, opened 2025-02-04
Fix trace printing

vincent-botbol:
This PR fixes issues identified by @rprimet.

The first issue is that the traces would be missing function call "begin" trace in specific scenarios. I.e., functions of arity > 1 defined in an included module.

The second issue is that the evaluate part is called twice which would yield a spurious `[]` when printing JSON traces.


rprimet (approved):
Looks good to me!

denismerigoux (approved):
Thanks for this important step toward #760!

PR #781, opened 2025-02-05
C backend: fix .h output of toplevel constants

AltGr:
@vincent-botbol Can you check that this fixes your issue ?


vincent-botbol:
> @vincent-botbol Can you check that this fixes your issue ?

```c
  result__1->AVVC__baremes_in = AVVC__baremes_par_defaut;
  ```
  It fails later on because it is not treated as a function call.

vincent-botbol (approved):
Looks good. Tested, it correctly fixes the bug.

PR #785, opened 2025-02-11
Make the `or if list empty` phrase optional

### LINKED ISSUE #764: Make optional all the initial and default values on list operators
We had taken the decision to force the user to put default values for list operations if the list is empty, but this turns out to be rather cumbersome and forcing the user to put dummy values. We should make these optional and the operators should raise runtime exceptions when they have an empty list and no default values.

AltGr:
on extremum and arg-extremum queries on lists. This required tuning the `reduce` internal operator to have a thunked default value.

This also includes a patch that adds support for running test scopes in Python

Todo:
- [x] update the doc and cheat-sheet
- [x] update the tree-sitter grammar

Closes #764


vincent-botbol (approved):
PR looks good. Let's not forget to update the tree-sitter as well.

AltGr:
→ https://github.com/CatalaLang/tree-sitter-catala/pull/5

PR #786, opened 2025-02-11
Fix C runtime windows build

vincent-botbol:
This PR fixes the build on windows which failed due to an invocation to `ar` which wasn't named this way on windows.
It also fixes an issue introduced in #784 which made the failing build fails even more because windows's equivalent to pkg-config doesn't consider the cygwin folder.

### DIFF
--- a/runtimes/c/discover.ml
+++ b/runtimes/c/discover.ml
@@ -1,0 +1,22 @@
+let () =
+  let open Re in
+  if Array.length Sys.argv <> 2 then
+    failwith "expected either ar or gmp_flags as argument"
+  else
+    match Sys.argv.(1) with
+    | "ar" ->
+      (* On mingw's opam windows install gcc is named: x86_64-w64-mingw32-gcc,
+         and, ar is named: x86_64-w64-mingw32-ar. We substitute gcc by ar from
+         the c compiler retrieved in the OCaml compiler's config to make the
+         build system handle it properly. *)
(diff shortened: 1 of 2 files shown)


runtimes/c/discover.ml
```diff
@@ -0,0 +1,21 @@
...
+  else
+    match Sys.argv.(1) with
+    | "ar" ->
+      (* On mingw's opam windows install gcc is named: x86_64-w64-mingw32-gcc,
+         and, ar is named: x86_64-w64-mingw32-ar. We substitute gcc by ar from
+         the c compiler retrieved in the OCaml compiler's config to make the
+         build system handle it properly. *)
+      Config.c_compiler
+      |> replace_string (compile (str "gcc")) ~by:"ar"
+      |> print_string
```
AltGr, line 14: Can it not happen (on mac ?) that this is `cc` or `clang` ? In this case this would call gcc instead of `ar` ; maybe check if `Config.c_compiler` ends with `gcc` and then only replace, otherwise returning `ar` ?

AltGr (approved):
🎉

runtimes/c/discover.ml
vincent-botbol: Oh you're absolutely right.

AltGr: Nice! `ar` never needs `.exe` ?

vincent-botbol: Windows automatically create aliases for .exe so it's safe to call it without the extension.

PR #792, opened 2025-02-20
Propagate surface's scopecall binders position

vincent-botbol:
This PR propagates the positions of the scope call's binders (underlined in the picture):
image

### DIFF
--- a/compiler/shared_ast/definitions.ml
+++ b/compiler/shared_ast/definitions.ml
@@ -565,7 +565,8 @@ and ('a, 'b, 'm) base_gexpr =
   | ELocation : 'b glocation -> ('a, (< .. > as 'b), 'm) base_gexpr
   | EScopeCall : {
       scope : ScopeName.t;
-      args : ('a, 'm) gexpr ScopeVar.Map.t;
+      args : (Pos.t * ('a, 'm) gexpr) ScopeVar.Map.t;
+          (* Map elements contain their variable's surface position *)
     }
       -> ('a, < explicitScopes : yes ; .. >, 'm) base_gexpr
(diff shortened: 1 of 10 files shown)


compiler/shared_ast/definitions.ml
```diff
@@ -565,7 +565,7 @@ and ('a, 'b, 'm) base_gexpr =
   | ELocation : 'b glocation -> ('a, (< .. > as 'b), 'm) base_gexpr
   | EScopeCall : {
       scope : ScopeName.t;
-      args : ('a, 'm) gexpr ScopeVar.Map.t;
+      args : (Pos.t * ('a, 'm) gexpr) ScopeVar.Map.t;
```
denismerigoux, line 568: A comment would be nice here to explicit what is the position referring to.

denismerigoux (approved):
Very nice, indeed we were forgetting the position of this binder...

compiler/shared_ast/definitions.ml
vincent-botbol: Indeed. As there is not a lot of comments here, I thought it made it look weird but it's very relevant.

denismerigoux (approved):

vincent-botbol:
It turns out that an invariant I believed true was actually false for some weird module paths. I patched things up, thanks to @AltGr, so that the invariant that both the modules path from the surface and the resolved modules path from `Name_resolution` have the same length.

vincent-botbol:
This PR introduces a regression in the tests outputs. It seems to only be a minor one that only affects files with complex module usages. I think it should be fine.

AltGr (approved):
LGTM!

PR #803, opened 2025-03-25
Fix error positioning affecting lsp

vincent-botbol:


AltGr:
Wow, I thought these were equivalent... shouldn't `Message.join_pos` emit an internal warning when some arguments are ignored in this way ?

AltGr (approved):

Issue #736, opened 2024-10-30
Implement attachement of code comments to AST node via an AST attribute system


denismerigoux:
In order to make progress on #673, the attribute system of #392 has entered the critical path. This issue states that work on the attribute system should begin with the following objectives : 
* capture comments inside Catala code blocks as either surfact AST nodes or annotations on neighbouring surface AST nodes (the parser and lexer should be modified) ;
* the rules for knowing where to attach comments to AST nodes can be simple at first : the comment is always attached to the AST node of the token immediately after the comment in the linear flow of tokens ;
* code comments attached to AST nodes should be pushed down the compilation chain at least to Desugared (used by #673).

AltGr:
Proposed attribute syntax (similar to [Rust](https://doc.rust-lang.org/reference/attributes.html), itself taken from C#):
- outer syntax `#[ATTR]`
- always bound to the AST node immediately following
- (optionally, we could also support `#![ATTR]` that bind to their parent node)
- inner syntax: `ATTR` is the form `VAR`, or `VAR = VAL`, where
  - `VAR` is a lident, or a dot-separated list thereof (`path.to.key`)
  - `VAL` is a `"`-delimited string (where `\"` escapes `"` and `\\` escapes `\` ?)
  - (optionally, we could extend that to also support arbitrary literals, or even arbitrary expressions ?)

example:
```catala-fr
declaration structure InputStruct:
  #[form.user_description = "This field corresponds to XXX and should be filled if YYY"]
  data user_xxx_data content money

#[test]
declaration scope Test1:
  ...
```

Implementation-wise, they will be added to the marks which already contain position and optionally type data

AltGr:
Some implementation thoughts: at the moment, extended marks are only present on expressions. Annotations will be useful on types, constructors, declarations, etc. as well.

Since these already have positions, it is tempting to simply extend the positions with annotation data ; or replace them with a hierarchical type, but this will require an update to all the places were positoin data is accessed.
Also: annotations themselves will have (raw) positions

denismerigoux:
I like your specification ! I'm also in favor of just extending the position marks everywhere to contain attributes. And yes the attribute type should carry the positions of where the attributes are in the source code :)

denismerigoux:
Good catch, Pierre. Yes, definitely, and with attributes the system could be more flexible and customizable than with hardcoded syntax additions.

AltGr:
So that would be something like:
```catala-code-en
scope S:
  #[requires]
[...]

PR #804, opened 2025-03-27
Add attributes to the language

AltGr:
This squeezes attributes into position types, and extends the parser to support them.

Implements #736, although there may be additional work needed to:
- ensure an attribute is allowed at a desired position (not all parser nodes handle them)
- ensure the attributes are propagated as expected during the compiler transformations (in particular, from surface). For example, an attribute before a parenthesized expression disappears at the moment. This might be tricky; note however that attributes on arbitrary expressions aren't the main use-case.

Still needs some work as, at the moment the type of annotations is defined inside `Surface.Ast`, which is not available from `Shared_ast` and that's pretty limitating.


compiler/surface/fill_positions.ml
denismerigoux, line 1: Why is this not necessary anymore?

compiler/catala_utils/pos.mli
```diff
@@ -18,11 +18,14 @@
...
 type t
 (** A position in the source code is a file, as well as begin and end location
-    of the form col:line *)
+    of the form col:line, and optionally, attributes *)
 
 val compare : t -> t -> int
 val equal : t -> t -> bool
 
+type attr = ..
+type attr += Law_pos of string list
```
denismerigoux, line 27: I'm not against it but why cramming the attributes in the positions instead of the `mark` type of the AST?

compiler/desugared/name_resolution.mli
```diff
@@ -162,6 +162,10 @@ val get_def_key :
   Ast.ScopeDef.t
 (** Usage: [get_def_key var_name var_state scope_uid ctxt pos]*)
 
+val surface_rule_to_def : Surface.Ast.rule -> Surface.Ast.definition
+(** Translates a {!type: Surface.rule} into the corresponding
+    {!type: Surface.definition} *)
+
```
denismerigoux, line 168: Why moving this here? Dependency cycle?

compiler/surface/ast.mli
```diff
@@ -275,70 +258,77 @@ and scope_decl_context_data = {
...
 
-and source_file = law_structure list
-[@@deriving visitors { variety = "map"; ancestors = ["Mark.pos_map"] }]
+type source_file = law_structure list
+type attr_path = string list Mark.pos
+
+type src_attr_value =
+  | String of string Mark.pos
+  | Expression of expression
+  | Unit
```
denismerigoux, line 329: To me the type should be more like this:

```ocaml
type src_attr_value =
  | Dict of src_attr_value StringMap.t
  | DocString of string Mark.pos
  | String of string Mark.pos
  | Expression of expression
  | Unit
```

tests/exception/bad/exceptions_cycle.catala_en
```diff
@@ -34,6 +34,7 @@ $ catala test-scope A
 │    │   ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 │ 10 │   definition x equals 0
 │    │   ‾‾‾‾‾‾‾‾‾‾‾‾
+├─ Test
```
denismerigoux, line 37: This PR fixes #194 !

compiler/catala_utils/pos.mli
AltGr: Marks are only on expressions. Positions are everywhere. And attributes are actually mostly expected to be useful on types, top-level declarations, structure fields, etc.
Now, indeed, we could rename `Pos` to something else (and maybe use raw positions inside of attributes instead of positions that could at the moment recursively embed attributes, although we make no use of that). But that would make the changes bigger and the actual change more difficult to review...

compiler/desugared/name_resolution.mli
AltGr: Since removing the use of `Visitors`, it became possible to make `Surface.Ast` an interface-only module by just removing these two functions.

tests/exception/bad/exceptions_cycle.catala_en
AltGr: Yes as mentionned in fe342d0 :)

compiler/catala_utils/pos.mli
denismerigoux: To make myself clearer, I was envisionning a modification to 

https://github.com/CatalaLang/catala/blob/b15b5d11c4082f0626b48e26bd02a252acf6f136/compiler/shared_ast/definitions.ml#L431-L446

to add the attributes there instead of directly into `Pos.t`.

AltGr:
Now in much better shape and should become usable. There is a small demo of testing `#[debug.print]` added to the tests.

Yay! Debug printings !

denismerigoux:
Thanks @AltGr, looks good for merging but first you should add documentation about attributes in the "Language reference" section of https://github.com/CatalaLang/catala-book!

PR #809, opened 2025-04-18
Add Catala to Java backend

vincent-botbol:
Disclaimer: Joint work with @rprimet 

This PR adds a new Java backend to Catala. This currently behaves similarly to the other backends.

One can invoke: `catala java <catala_file>` to generate an equivalent translation of a Catala program to Java.

For "modular" projects, clerk should be used, e.g., `clerk run --backend=java my_file.catala_en` will take care of building every dependency. In the near future, we will also provide a way to generate Java `jar` files to export a Catala library. In the meantime, similarly as the C backend, users need to retrieve the generated files in `_build` and pack it themselves (or simply copy the Java files into an existing project). 
Currently, the generated Java files have a few spurious warnings that can be treated as safe.

@rprimet also implemented a Java equivalent for `dates_calc`. We found it more convenient to vendor it as is while we wait for a future `dates_calc` release. We also vendored BigFraction from an Apache library which, in due time, will probably be replaced by an in-house implementation.

The call trace is currently not implemented (out of laziness) -- It will also land in a near future.


vincent-botbol:
(Not looking at the CI rn, I want a calm extended week-end)

runtimes/java/catala/runtime/CatalaInteger.java
```diff
@@ -0,0 +1,129 @@
+package catala.runtime;
+
+import java.math.BigInteger;
+
+import catala.runtime.exception.CatalaException;
+
+// We could make this class extend java.math.BigInteger
+// for consumer friendliness?
```
denismerigoux, line 8: <3

runtimes/java/catala/runtime/exception/CatalaException.java
```diff
@@ -0,0 +1,16 @@
+package catala.runtime.exception;
+
+// I guess catala exceptions should be unchecked?
+// (hence the RuntimeException base)
+// Rationale: I'm not sure an attempt to recover from a
```
denismerigoux, line 5: You would want to catch from a `CatalaException` outside of the generated Catala code, i.e in the application calling the Catala code. However, if by "recovering" you mean resuming the Catala computation with a continuation, then no we don't want that.

runtimes/java/dune
denismerigoux, line 1: So you can compile Java from `dune`, what a world we live in!

denismerigoux, line 1: So you can compile Java from `dune`, what a world we live in!

runtimes/java/org/apache/commons/numbers/core/Addition.java
denismerigoux, line 1: This vendoring is not the nicest thing but OK.

denismerigoux (approved):
I bow down to you, masters of `Java`naise! How much time did it take you all in all?

build_system/clerk_driver.ml
```diff
@@ -521,6 +529,61 @@ let run_cmd =
         copy_in ~src:(target -.- "py") ~dir:tdir;
         close_out (open_out (tdir / "__init__.py"));
         []
+      | `Java ->
+        let jar_target = target -.- "jar" in
+        (* We copy over the runtime and update the jar with the target
+           classes *)
+        copy ~src:(List.hd (get_var Var.runtime_java_jar)) ~dst:jar_target;
```
AltGr, line 536: So this means that if we generate several `jar`s, every one of them will include the runtime ? Won't that be causing conflicts ?

AltGr (approved):
Great work! This seems all good, I've just a few questions inlined.

runtimes/java/dune
vincent-botbol: Yeah, I'm wary of windows & mac though.. I'll submit a patch for the virtual java opam package so that the missing distributions are handled (in particular, alpine and windows).

runtimes/java/org/apache/commons/numbers/core/Addition.java
vincent-botbol: Yes, we'll clean it up eventually.

build_system/clerk_driver.ml
vincent-botbol: Yes, I had two choices : make a self-contained executable `jar` (i.e. `java -jar <the_jar.jar>`) or link the runtime : 
`java -cp <path_to_runtime.jar>:<path_to_the_jar.jar> <MainClassName>`.
For the backend testing, the second one makes more sense. I'll switch to this one. We can rework it eventually when we tackle clerk targets.

compiler/scalc/to_java.ml
```diff
@@ -0,0 +1,1272 @@
...
+    (pp_print_list ~pp_sep:pp_comma (fun ppf info ->
+         fprintf ppf "\"%a\"" Uid.MarkedString.format info))
+    uids
+
+let format_string_list (ppf : formatter) (uids : string list) : unit =
+  let sanitize_quotes = Re.compile (Re.char '"') in
+  fprintf ppf "new String[]{%a}"
+    (pp_print_list ~pp_sep:pp_comma (fun ppf info ->
+         fprintf ppf "\"%s\""
+           (Re.replace sanitize_quotes ~f:(fun _ -> "\\\"") info)))
```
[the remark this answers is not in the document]
vincent-botbol: I added `String.quote` in 318c1d73, let me know if that looks good to you.

build_system/clerk_driver.ml
vincent-botbol: (done)

PR #812, opened 2025-05-05
Fix incorrect file reading

vincent-botbol:
This PR fixes an issue on windows where we try to read the contents of a file based on the length of the input channel but on windows, when files are opened in text mode, some end of line translations happen and the result of `in_channel_length` is faulty. Hence, we should default to open files in binary mode instead (except in parsing as we somewhat care for such characters).

Relevant doc:
```ocaml
val in_channel_length : in_channel -> int
(** Return the size (number of characters) of the regular file
    on which the given channel is opened.  If the channel is opened
    on a file that is not a regular file, the result is meaningless.
    The returned size does not take into account the end-of-line
    translations that can be performed when reading from a channel
    opened in text mode. *)
```

### DIFF
--- a/compiler/catala_utils/file.ml
+++ b/compiler/catala_utils/file.ml
@@ -132,11 +132,11 @@ let find_in_parents ?cwd predicate =
 
 let with_out_channel filename f =
   ensure_dir (Filename.dirname filename);
-  let oc = open_out filename in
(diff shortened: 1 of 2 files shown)


denismerigoux (approved):
Clever bugfix, these small things are really annoying for inter-platform compat...

PR #817, opened 2025-06-03
Purity

AltGr:
Add a purity check on expression, and use it to guard some optimisations.


AltGr:
Ci red because github does'nt use the latest version of the rules, but it should be OK

PR #819, opened 2025-06-10
Avoid shell / native windows build

rprimet:
Needs some cleanup, but it can build + run examples from the CNAF POC natively using clerk.

Capture d’écran du 2025-06-10 22-12-29

### DIFF
--- a/compiler/catala_utils/file.ml
+++ b/compiler/catala_utils/file.ml
@@ -309,6 +309,7 @@ let () =
     in
     let count =
       if not Unix.(isatty stdin) then from_env ()
+      else if Sys.win32 then default
       else
         try
           (* terminfo *)
@@ -351,11 +352,18 @@ let check_file f =
   with Unix.Unix_error _ | Sys_error _ -> None
 
 let get_command t =
+  let cmd, args =
+    if Sys.win32 then
+      let t_exe =
+        if not (String.ends_with ~suffix:"exe" t) then t ^ ".exe" else t
+      in
+      "where.exe", [t_exe]
+    else "/bin/sh", ["-c"; "command -v " ^ Filename.quote t]
+  in
   String.trim
   @@ process_out
        ~check_exit:(function 0 -> () | _ -> raise Not_found)
-       "/bin/sh"
-       ["-c"; "command -v " ^ Filename.quote t]
+       cmd args
 
 let check_exec t =
   try
--- a/compiler/catala_utils/file.ml
+++ b/compiler/catala_utils/file.ml
@@ -355,7 +355,7 @@ let get_command t =
   let cmd, args =
     if Sys.win32 then
       let t_exe =
-        if not (String.ends_with ~suffix:"exe" t) then t ^ ".exe" else t
+        if not (Filename.check_suffix t ".exe") then t ^ ".exe" else t
       in
       "where.exe", [t_exe]
     else "/bin/sh", ["-c"; "command -v " ^ Filename.quote t]
--- a/build_system/clerk_rules.ml
+++ b/build_system/clerk_rules.ml
@@ -296,7 +296,12 @@ let[@ocamlformat "disable"] static_base_rules enabled_backends =
         ~description:["<catala>"; "tests"; "⇐"; !input];
 
       Nj.rule "dir-tests"
-        ~command:["cat"; !input; ">"; !output; ";"]
+        ~command:
+        (if Sys.win32 then
+          ["cmd"; "/c"; "type"; !input; ">"; !output; ";"]
+        else
+          ["cat"; !input; ">"; !output; ";"]
+        )
         ~description:["<test>"; !test_id];
      ]
    else [])
--- a/build_system/clerk_rules.ml
+++ b/build_system/clerk_rules.ml
@@ -233,7 +233,11 @@ let[@ocamlformat "disable"] static_base_rules enabled_backends =
   let open Var in
   [
     Nj.rule "copy"
-      ~command:["cp"; "-f"; !input; !output]
+      ~command:
+        (if Sys.win32 then
+           ["cmd"; "/c"; "copy"; "/Y"; !input; !output]
+         else 
+           ["cp"; "-f"; !input; !output])
       ~description:["<copy>"; !input];
   ] @ (if List.mem OCaml enabled_backends then [
       Nj.rule "catala-ocaml"


build_system/clerk_scan.ml
```diff
@@ -60,7 +60,10 @@ let rec find_test_scope ~lang file =
     match Seq.uncons lines with
     | Some ((_, L.LINE_TEST_ATTRIBUTE, _), _) -> true
     | Some ((_, L.LINE_INCLUDE f, _), lines) ->
-      let f = if Filename.is_relative f then File.(file /../ f) else f in
+      let f =
+        if Filename.is_relative f then Filename.concat (Filename.dirname file) f
+        else f
+      in
```
AltGr: This seems to do the same thing as before ? If not maybe it's the `/../` operator that should be fixed instead.

compiler/catala_utils/file.ml
```diff
@@ -273,72 +273,26 @@ let rec remove t =
...
-        width := Some r;
-        r)
+  with End_of_file -> (
+    let status = Unix.close_process (ic, oc) in
+    match status with
+    | Unix.WEXITED n ->
+      check_exit n;
+      Buffer.contents buf
+    | Unix.WSIGNALED n | Unix.WSTOPPED n ->
+      failwith (Printf.sprintf "Sub-process %s was killed (%d)" cmd n))
```
AltGr: I don't see the point of rewriting this function, which wasn't using a shell. The only meaningful change I can find, after untangling the syntactic rewrite, is replacing `open_process_args_in` by `open_process_args`, which returns an additional `oc` that we don't use anyway. (this might change the behaviour if the sub-process tries to read from stdin, it would now block forever instead of accepting keyboard input)

Also, the finaliser is no longer called if we get a signal or an exception (other than the expected EOF) during the execution of the sub-process. This may leave a dangling process running.

compiler/catala_utils/file.ml
```diff
@@ -351,11 +305,19 @@ let check_file f =
...
-  @@ process_out
-       ~check_exit:(function 0 -> () | _ -> raise Not_found)
-       "/bin/sh"
-       ["-c"; "command -v " ^ Filename.quote t]
+  let path = try Sys.getenv "PATH" with Not_found -> "" in
+  let paths = String.split_on_char (if Sys.win32 then ';' else ':') path in
+  let rec search = function
+    | [] -> raise Not_found
+    | dir :: rest ->
+      let full_path = Filename.concat dir t in
```
AltGr: Please use our custom operators (`dir / t`)

compiler/catala_utils/file.ml
```diff
@@ -351,11 +305,19 @@ let check_file f =
...
+      if Sys.file_exists full_path then full_path
+      else if Sys.win32 && not (String.ends_with ~suffix:".exe" t) then
+        let full_path_exe = Filename.concat dir (t ^ ".exe") in
+        if Sys.file_exists full_path_exe then full_path_exe else search rest
+      else search rest
+  in
+  search paths
```
AltGr: Looking up a command as done by the OS is, unfortunately, a bit more complex than iterating in the `PATH` (see https://github.com/ocaml/opam/blob/708d62ec2b020fb1f176f7f8184f93bd1259496d/src/core/opamStd.ml#L1266 , you also need to check for execution permission, for example).

We could try to keep the same approach as before but using `where.exe` on Windows, or add checks in this new version... But fortunately, this function doesn't seem to be actually used 🤷🏿‍♂️ so it's probably not worth the trouble !

compiler/catala_utils/message.ml
```diff
@@ -38,32 +38,47 @@ let () = ignore (unstyle_formatter Format.str_formatter)
...
+  in
+  let () =
+    try
+      Sys.set_signal 28 (* SIGWINCH *)
+        (Sys.Signal_handle (fun _ -> width := None))
+    with Invalid_argument _ -> ()
+  in
+  ref get_cols
+
+let set_terminal_width_function f = terminal_columns := f
```
AltGr: this (unpleasant) ref callback was defined in order to break the dependency cycle (the function to check screen width was defined in `File` because it needed process handling, but used in `Message`, while `File` actually depends on `Message`) ; setting the ref at runtime, during the load of the module `File` works around the problem, but exposes this `set_..._function` which is not pretty.

You are apparently making an attempt here to resolve that issue in a cleaner way, which may be a good idea (but in that case, remove this `set_...` function from the interface ! Wasn't that the point ?)

Going through a separate `Terminal` module is a good idea, but at the cost of the duplication of the `process_out` function from `File`, which creates a pitfall for future updates, in particular since this function is a bit tricky and may receive fixes for e.g. better error handling in the future. At least annotate every copy with a big `warning` comment specifying where the other copies are if this is really required.

All in all I am not sure that this strikes a better balance than what we had before and is tested.

compiler/catala_utils/message.mli
```diff
@@ -85,7 +85,7 @@ val unformat : (Format.formatter -> unit) -> string
 
 val has_color : out_channel -> bool
 val set_terminal_width_function : (unit -> int) -> unit
-val terminal_columns : unit -> int
```
AltGr: This shouldn't be a reference anymore, the function can now be set normally

compiler/catala_utils/terminal.ml
```diff
@@ -0,0 +1,57 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+let get_width () =
+  let default = 80 in
+  if Sys.win32 then (* Placeholder for Windows *)
+    default
```
AltGr: there is code to retrieve the number of columns on Windows [here](https://github.com/ocaml/opam/blob/708d62ec2b020fb1f176f7f8184f93bd1259496d/src/core/opamStd.ml#L970), but it requires C stubs and it's probably not worth it, so falling back to the default seems OK to me.

compiler/catala_utils/file.ml
rprimet: Could you add a some detail about the unused parts? `get_command` seems to be used by `check_exec` which is called from various other places in the code?

build_system/clerk_scan.ml
rprimet: Yes, reverting, thanks!

compiler/catala_utils/message.ml
rprimet: Agreed, reverting to master

PR #824, opened 2025-06-17
Customtarget

AltGr:
Two commits here:
- the first one cleans up how ninja is run and information passed around it.
- the second adds two new configurations fields in `clerk.toml`, details copied below for convenience:

## Custom targets

```toml
[[target]]
name = "french_law"
modules = ["Prestations_familiales", "Allocations_familiales"]
backends = ["c", "python", "ocaml"]
include_runtime = true
```

This will install all the concerned files (including dependencies) into
`_targets/french_law` (should "target" be singular ?), with a subdirectory for
each backend. This should make it easier to retrieve the desired artifacts in a
given app.

`clerk build` without arguments will now build all targets by default.

## Custom rules

This is more experimental, and doesn't combine with point 1 at the moment.
It allows to add one custom compilation step at the end of the compilations done
by catala, for example it could be used to generate a `.so` with

```toml
[[rule]]
out_exts = ["so"]
in_exts = ["o"]
backend = "c"
commandline = ["$CC", "$CFLAGS", "$C_INCLUDE_FLAGS", "$RUNTIME_C_LIBS", "$src", "-shared", "-o", "$dst"]
```

The commandline does variable expansion ; w e are now leaning more in the
direction of using a custom `target` and then letting the user compile directly
as they wish (perhaps providing at least Makefiles / .d files for helping with
dependencies).

However, we don't yet have a story for plugins (e.g. generation of _web_api.ml /
js files) so this is a step in the exploration of what we could do there.

### DIFF
--- a/.github/workflows/harness.yml
+++ b/.github/workflows/harness.yml
@@ -151,7 +151,7 @@ jobs:
       - name: Build examples
         run: |
           cd ~/catala-examples
-          . ~/catala/_python_venv/bin/activate && opam --cli=2.1 exec -- make all testsuite install
+          opam --cli=2.1 exec -- make all testsuite install
       - name: Generate examples test report
         if: ${{ always() }}
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ node_modules/
 build.ninja
 **/#*
 **/.#*
+*.~*~
 .envrc
 .direnv
 **/__pycache__/
--- a/Makefile
+++ b/Makefile
@@ -348,7 +348,7 @@ alltest: dependencies-python
 	$(MAKE) testsuite && \
 	$(test_title) "Running catala-examples" && \
 	$(call local_tmp_clone,catala-examples) && \
-	$(MAKE) -C catala-examples.tmp \
+	$(PY_VENV_ACTIVATE) $(MAKE) -C catala-examples.tmp \
 	  CATALA=$(CURDIR)/_build/install/default/bin/catala \
 	  CLERK=$(CURDIR)/_build/install/default/bin/clerk \
 	  BUILD=../_build/default \
--- a/build_system/clerk_cli.ml
+++ b/build_system/clerk_cli.ml
@@ -374,9 +374,14 @@ let init
     let dir =
       match build_dir with None -> config.global.build_dir | Some dir -> dir
     in
-    let d = File.clean_path dir in
-    File.ensure_dir d;
-    d
+    let dir =
+      match test_flags with
+      | [] -> dir
+      | flags -> File.((dir / "test") ^ String.concat "" flags)
+    in
+    let dir = File.clean_path dir in
+    File.ensure_dir dir;
+    dir
     (* Note: it could be safer here to use File.(Sys.getcwd () / "_build") by
        default, but Ninja treats relative and absolute paths separately so that
        you wouldn't then be able to build target _build/foo.ml but would have to
--- a/build_system/clerk_cli.ml
+++ b/build_system/clerk_cli.ml
@@ -388,6 +388,7 @@ let init
         config with
         global =
           {
+            config.global with
             build_dir;
             catala_exe;
             catala_opts = config.global.catala_opts @ catala_opts;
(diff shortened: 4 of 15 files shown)


AltGr:
Thanks for the review !
Answering your interrogations:

> Do you have a sister PR that applies this to catala-examples?

I tested some things locally, but nothing complete yet. We could use it to simplify the python and ocaml deliverables (the snippet you gave is correct and works ; the Makefile should then be adapted to build the libs from there), but a big part there is the generation of web APIs using the catala plugin and we don't have a good story for that yet.

> For me the backends parameter can be even more complex:

You can customise, but only globally, using
```toml
[variables]
CATALA_FLAGS_PYTHON = [ "--enums-as-classes" ]
```
Targets as proposed here only allow to select what should be built and make it easier to retrieve it: they can't be used to affect the build itself. Selecting specific flags for specific modules could be added using a separate mechanism, though, if needed (for a flag such as `--enums-as-classes`, the modules compiled with and without would very likely be made incompatible, so it looks like a bad idea 😅 ).

> I don't yet quite see where this would be useful. 

The use-case is basically calling the linker of your choice with the options of your choice: with it, you can instruct `clerk` on how to build a `.so` or a `.cmxa`, and then these can be used as targets to `clerk build`, for any module.

Essentially, the important feature this wraps is the recovery of the (ordered) dependencies of the given module, with the expected extension... similar to what `catala depends` allows, but without recomputing it from scratch for every command. Appropriate `.d` files (or Makefile, or dune, or whatever) generated for the targets could, maybe, fulfill this need instead.

I am reasonably confident about the `[target]` sections, but these custom rules are definitely a prototype designed to help pinpoint our exact need :)

> What are the variables allowed in commandline ?

They can be seen using `clerk list-vars`:
```bash
CATALA_FLAGS_C="-O"
CATALA_FLAGS_JAVA="-O"
CATALA_FLAGS_OCAML="-O"
CATALA_FLAGS_PYTHON="-O"
CC="cc"
CFLAGS="-std=c89 -pedantic -Wall -Wno-unused-function -Wno-unused-variable -Wno-unused-but-set-variable -Werror -g"
C_INCLUDE_FLAGS="-I ~/catala/_build/install/default/lib/catala/runtime_c"
JAVA="java"
JAVAC="javac"
JAVAC_FLAGS="-implicit:none"
OCAMLC_EXE="ocamlc"
OCAMLOPT_EXE="ocamlopt"
OCAML_FLAGS=""
OCAML_INCLUDE="-I ~/.opam/latest/lib/zarith -I ~/.opam/latest/lib/dates_calc -I ~/catala/_build/install/default/lib/catala/runtime_ocaml"
PYTHON="python3"
RUNTIME_C_LIBS="-L~/catala/_build/install/default/lib/catala/runtime_c -lcatala_runtime -lgmp"
RUNTIME_JAVA_JAR="~/catala/_build/install/default/lib/catala/runtime_java/catala_runtime.jar"
RUNTIME_OCAML_LIBS="zarith.cmxa dates_calc.cmxa ~/catala/_build/install/default/lib/catala/runtime_ocaml/runtime_ocaml.cmxa"
RUNTIME_PYTHON="~/catala/_build/install/default/lib/catala/runtime_python/src"
jar="jar"
```

These are the same that can be overriden using the `[variables]` section.

AltGr:
> The second clerk build invocation should re-build everything without the `--trace` flag since it cannot reuse the things compiled with the previous invocation of clerk build. So maybe that points to a target-specific build folder in _build, itself subdivided into backend-specific build subfolders?

Ok, this implies a complete reimplementation of the feature, but it's not like the current proto does anything complex. You are re-defining dune's [build contexts](https://dune.readthedocs.io/en/stable/overview.html#term-build-context) ;)

The subdirectories in `_build` should not be specific to a given target, but to a given set of flags, though: otherwise that would completely bar code reuse between targets. Of course, that wouldn't be visible from `_targets`.

> Reading your comments, I have the impression that for you the main way for the user to interact with clerk build is to invoke a command like clerk build <name of the file you want to build>, like clerk build french_law.so.

Not really, that is indeed the idea for the proposed version of `[rule]`, but I insist that this one is very preliminary and not intendend as the main user-facing point. For the use-case you detail, my intention was something along the lines of the following:

```toml
[[rule]]
out_exts = [ "so" ]
...

[[target]]
modules = ["Prestations_familiales", "Allocations_familiales"]
extensions = [ "so" ]
# or even : files = [ "foo.so", "bar.so" ]
...
```


On both points, maybe my remarks will make more sense with the intuition that my implicit model had the rules as an _how_ things can be built while `[target]` is purely a definition of _what_ to build.

PR #825, opened 2025-06-18
Fix spurious fence errors in parsing

vincent-botbol:
This PR fixes an issue that occurs when the parse fails to recover from a parsing error and ends up in the lexer finalizer that triggers an error regarding unclosed fences which is completely unrelated and misleading for the user. 

A test is there to exhibit the behavior and its fix.


denismerigoux (approved):
Nice quality-of-life improvement, thanks!

PR #834, opened 2025-07-18
dcalc printer: print whole program rather than the to_expr version

AltGr:
this puts it in sync with the lcalc printer and is actually more informative.


denismerigoux (approved):
Excellent, thank you!

PR #835, opened 2025-07-18
Add "impossible" to the language

AltGr:
Not supported in all backends yet


compiler/scalc/to_java.ml
```diff
@@ -562,34 +562,14 @@ let rec format_stmt ~toplevel (ctx : context) ppf (stmt : Ast.stmt Mark.pos) =
...
-    in
     let pp_default_initializer ppf =
-      match switch_return_var with
-      | None -> fprintf ppf "@\n@[<v 4>default: {@ return null;\n}@]"
-      | Some v ->
-        fprintf ppf "@\n@[<v 4>default: {@ %a = null;@ break;@\n}@]"
-          VarName.format v
+      fprintf ppf
+        "@\n\
+         @[<v 4>default: {@ throw new RuntimeException(\"Unreachable case\");@;\
```
denismerigoux, line 568: The Java runtime exception should carry with it the source code position for the `impossible` value in the source Catala code no?

runtimes/ocaml/operators.catala_en
denismerigoux, line 1: What is the purpose of this file?

runtimes/ocaml/stdlib.catala_en
denismerigoux, line 1: I think you committed in this PR some work belonging to another PR :)

denismerigoux (approved):
Thanks @AltGr! Could you do a sister PR in `catala-examples` that uses the new `impossible` keyword whenever I put comments like "should not happen" in the code?

compiler/scalc/to_java.ml
AltGr: this is not the "impossible" handling, which indeed does that already. This is the "default" case of already exhaustive pattern-matchings that Java insists we must still add (I tried "assert false", but that has some uncanny meaning in Java ; previously we were just returning `null` but there were errors in the Java typing heuristics with that)

runtimes/ocaml/operators.catala_en
AltGr: Oops, that's just some early skeleton for the stdlib that slipped here by mistake

AltGr:
> Also, this PR is a good time to respect our new PR checklist:
> ## Checklist
> 
> I confirm that have have checked and updated each of the following items if this PR impacts them:
> 
>     * [ ]  Add or update syntax constructions in the syntax cheat sheet at `doc/syntax/syntax[_fr].typ`.
> 
>     * [ ]  Add or update syntax constructions in the syntax highlighting plugins in `syntax_highlighting/` (all of them).
> 
>     * [ ]  Add or update syntax constructions in the `tree-sitter` grammar at https://github.com/CatalaLang/catala-format/blob/master/catala.scm
>       
>       * Sister PR : https://github.com/CatalaLang/catala-format/pull/XXX
> 
>     * [ ]  Complete or update the documentation for the feature on the Catala book at https://github.com/CatalaLang/catala-book.
>       
>       * Sister PR: https://github.com/CatalaLang/catala-book/pull/XXX
> 
>     * [ ]  Update the existing Catala code in case of breaking changes at https://github.com/CatalaLang/catala-examples.
>       
>       * Sister PR : https://github.com/CatalaLang/ccatala-examples/pull/XXX

Will do, but I'm making many small additions to the syntax at the moment and intend to update those in batch 😇

PR #836, opened 2025-07-18
Add missing syntax updates

vincent-botbol:
Also updated : tree-sitter, vscode extension and catala-format.


denismerigoux (approved):
Thanks very much!

PR #839, opened 2025-07-23
Add polymorphic enum `Optional` to the language

AltGr:
And various other fixes to bugs I encountered along the way, including
- all Java compiler warnings (at least on our tests)
- the irritating ninja "no work to do"
- catala color when called through clerk
- `()` printed on unit constructors

etc., see list of commits for details ; this test also led us to ensure polymorphism actually works in Java

### DIFF
--- a/compiler/plugins/api_web.ml
+++ b/compiler/plugins/api_web.ml
@@ -121,8 +121,8 @@ module To_jsoo = struct
         elts
     | TOption t ->
       Format.fprintf fmt
-        "(function Eoption.ENone () -> Js.null | Eoption.ESome x -> Js.some \
-         (%a x))"
+        "(function Optional.Absent () -> Js.null | Optional.Present x -> \
+         Js.some (%a x))"
         format_to_js t
     | TVar _ -> Format.fprintf fmt "Js.Unsafe.inject"
     | TForAll tb ->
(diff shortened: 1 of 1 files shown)


compiler/surface/lexer_fr.cppo.ml
```diff
@@ -34,6 +34,8 @@
 #define MS_OF "de"
 #define MS_LIST "liste de"
 #define MR_LIST "liste", space_plus, "de"
+#define MS_OPTION "Optionnel de"
```
denismerigoux, line 37: Not sure whether `Optional` should be capitalized here, as `liste` is not. I would be in favor of having those builtin polymorphic types to not be capitalized.

AltGr:
You can have a look at the example here: https://github.com/CatalaLang/catala/pull/839/files#diff-d5b95cd3b783ab2bea2ba4fbd8a68bf51e081ae8c48f26537e82f85a5b66edeb

It's basically presented as a usual enum with constructors `Present` and `Absent`, and the type must be written with `of xxx`, like in `Optional of integer` -- like for the type parameter of lists (btw, shouldn't the list type be capitalised for consistency ?).

I'll of course document all that shortly ; since it is a pure addition to the syntax I feel it's ok to merge the change in advance ?

compiler/surface/lexer_fr.cppo.ml
AltGr: I agree it's a problem, but was considering solving the discrepancy the other way round 😁 
the point being that `Optionnel` is still somehow a variant type and that's the convention for them...

but I guess built-in types are uncapitalised, and it may make as much sense to consider `list` and `optional` in that category vs. user-defined types which are capitalised...

What's your final takeaway ?

PR #848, opened 2025-08-05
Add Romanian language support

ly3xqhl8g9:
## Description

This PR adds full Romanian language support to Catala. The implementation follows the existing pattern for multi-language support and includes all necessary components for a complete language integration.

- Complete Romanian translation of all Catala keywords
- Proper handling of Romanian-specific characters using UTF-8 escape sequences
- All operators and language constructs properly localized
- Added Romanian lexer to the dune build configuration
- Integrated Romanian as a backend option in clerk_driver.ml
- Added Romanian language option to CLI (`--language=ro`)
- Fixed missing Romanian case in `compiler/desugared/name_resolution.ml`
- Ensured all language-dependent pattern matches include Romanian
- Created Romanian syntax documentation (`doc/syntax/syntax_ro.catala_ro`)
- Added Romanian syntax sheet build target in Makefile
- Updated CONTRIBUTING.md to mention Romanian support
- Added complete syntax highlighting support for Romanian (`syntax_highlighting/ro/`)
- Created Romanian test files covering arithmetic, boolean logic, conditionals, and literate programming
- All 627 tests pass with Romanian language support enabled

  * [ ] The corresponding PR is: (To be created - will add Romanian language documentation)


build_system/clerk_driver.ml
```diff
@@ -131,8 +131,8 @@ let backend_extensions =
     Clerk_rules.C, ["c"; "h"; "o"];
     Clerk_rules.OCaml, ["ml"; "mli"; "cmi"; "cmo"; "cmx"; "o"; "cmxs"];
     Clerk_rules.Python, ["py"];
-    Clerk_rules.Java, ["java"; "class"];
-    Clerk_rules.Tests, ["catala_en"; "catala_fr"; "catala_pl"];
+    Clerk_rules.Java, ["java"; "class"; "jar"];
```
AltGr: This change seems unexpected, probably slipped in during a rebase ?

compiler/surface/lexer_ro.cppo.ml
```diff
@@ -0,0 +1,167 @@
+(* This file is part of the Catala compiler, a specification language for tax and social benefits
+   computation rules. Copyright (C) 2020 Inria, contributors: Denis Merigoux
+   <denis.merigoux@inria.fr>, Emile Rolley <emile.rolley@tuta.io>
```
AltGr: Please set yourself as contributor :)

runtimes/c/runtime-test.c
```diff
@@ -4,7 +4,7 @@
 #include <runtime.h>
 #include <assert.h>
 
-void* test()
```
AltGr: This is harmless, but seems unrelated ?

AltGr (approved):

build_system/clerk_driver.ml
ly3xqhl8g9: Yes, fixing it.

compiler/surface/lexer_ro.cppo.ml
ly3xqhl8g9: I simply copied the Polish implementation, wasn't sure if to add or no and what. No problem if no mention, it's not a major contribution.

runtimes/c/runtime-test.c
ly3xqhl8g9: Make was giving an [warning](https://stackoverflow.com/questions/693788/is-it-better-to-use-c-void-arguments-void-foovoid-or-not-void-foo), I can revert it, not a problem.

AltGr: we can keep it, thanks

compiler/surface/lexer_ro.cppo.ml
AltGr: it can be useful to have a contact if the file is in need of maintainance later on ; no problem if you stick to your Github handle of course.

ly3xqhl8g9: Should I leave 'Denis Merigoux <denis.merigoux@inria.fr>, Emile Rolley <emile.rolley@tuta.io>' and then add myself or just add myself as the one that is responsible for the Romanian parser? Sorry, the etiquette is not clear.

AltGr: Yeah sorry, you're right to ask : you should set yourself as only contributor of the file for now, others will add themselves upon update.

AltGr:
(don't worry about the CI failures, we're updating our docker images and runners, I'll merge as soon as it's restored)

ly3xqhl8g9:
I have updated the branch. 

Running `make build_dev` hits a bug because of GNU vs BSD sed in the build system, even if it successfully builds:

In `build_system/dune` lines 58-66, there's a rule that regenerates `manpages.sexp`:

  `%{bin:clerk} 2>&1 | sed -n "s/\(^\|'\)[^']*\('\|$\)/ /gp;q"`

This sed command is supposed to extract command names from clerk's help output, but it's failing and returning nothing.

When the extraction fails:
  - The for loop only runs once (with empty `$page`)
  - Only generates 1 rule (for base clerk command)
  - The `(mode promote)` writes this truncated version

Old command (GNU-only):
  `sed -n "s/\(^\|'\)[^']*\('\|$\)/ /gp;q"`
  - Uses `\( \)` for groups ← Only works on GNU sed
  - Fails silently on BSD sed (macOS)

New command (Portable):
  `sed -E "s/[^']*'([^']*)'[^']*/\1 /g" | head -1`
  - Uses `-E` flag ← Supported by both BSD and GNU
  - Uses `( )` for groups in extended mode

AltGr:
Thanks a lot! The hand on making our build script portable is very much appreciated too.

PR #852, opened 2025-08-18
Lexing improvements ; add `#[implicit_position_argument]` attribute

AltGr:
## Description

This reworks the lexer, in particular removing lots of keywords that are now lexed as idents and recognised in their specific contexts — a good example is type names like `integer` which don't have a structuring role ; also allowing to cleanup the localisations that had started to leak outside of the lexer.

`date round increasing/decreasing` is now `date round up/down` (resp. `supérieur/inférieur` in french)

The position type is now present in the surface language, as `code_location`.

And a new attribute is added, `#[implicit_position_argument]`, which is put on a toplevel function argument of type `code_location` and is implicitely passed with the location of the **caller** whenever the function is called. This was required to enable possibly failing operations like the division to be wrapped in the standard library while still reporting the useful location in the user source code when raised.

In time, this will replace the adhoc mechanisms for handling error locations in the different backends.

  * The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/12
  * The corresponding PR is: https://github.com/CatalaLang/catala-examples/pull/31

* [X] Syntax cheat sheet at `doc/syntax/syntax_*.catala_*` and `doc/syntax/catala_*.typ`.
    - [x] The corresponding PR is: https://github.com/CatalaLang/tree-sitter-catala/pull/10
    - [x] The corresponding PR is: https://github.com/CatalaLang/catala-format/pull/5


AltGr:
(warning: PR stacked on top of #849)

AltGr:
I left this open for documentation, but please close after reading since it's superseded by #855

PR #20, opened 2020-07-03
French Tax Code example

jemsab:
Hi! As promised, here is an example from the French Tax Code (Code général des Impôts). 
I shall try and do some more articles if this one is suitable.
I have tried to follow the contribution guidelines, please let me know if anything else needs to be done

PR #855, opened 2025-08-29
Structure the runtime and stdlib to be built by Clerk

AltGr:
This aggregated quite a few changes over its course ; but the main point is that the installation of the runtimes and stdlib is a simple copy of the source files from Catala, and that Clerk now knows how to retrieve them, and compile them as needed within the catala project's `_build`.¹

The standard library can this way be defined as normal Catala modules, using externals in the standard way.

## Extra changes provided along the way

- Support for `## ` comments for docstrings (this is unrelated, but splitting it to a parallel commit led to conflicts ; I can delay it if needed though)
- Externals can now be provided in a per-backend subdir (`ocaml/foo_internal.ml` instead of the file just besides the catala file) ; this will be helpful for the stdlib to stay a little tidy
- Change required for the stdlib: the annotation `#[implicit_position_argument]` provides a way for standard library function to point to the point they were called from in case of error (this generalises the mechanism where operators like `/` would be tweaked in each backend to have an extra position argument : these custom mechanisms will be removed once we get to moving them to the stdlib). Relatedly, the `code_location` type is now a valid Catala type.

## Current status [UPDATE: 2025-09-01]

This works for the integrated tests, for all backends (in-source and using a full Catala install).
- [x] ~~... except that the stdlib is far from complete, and the `Date` functions haven't yet been bound in all backends~~ The base stdlib functions for feature parity (Date's `first_day_of_month`, etc.) have now been implemented in all backends
- [x] ~~The generation of clerk target trees has not yet been tweaked to take the new files into account (but it should be simpler than before). It might be best to actually limit it to providing all the source files in the target language without any attempt to compile them -- we do it for backend tests, but providing "libraries" across backends might be beyond our scope.~~ Clerk build now copies all stdlib artifacts in `_targets` as expected ; the `include_runtime` flag no longer exists since it is the default: the runtime and stdlib are no longer provided as backend-dependent library, only source
- [x] ~~The move of the date functions to the stdlib obviously breaks the Z3 tests that checked them ; I am not sure what to do at this point.~~ It has been agreed on to leave these tests disabled for now
- [ ] (**NEW**) A regression has been detected when combining --trace and closure conversion (in other words, the C backend). It is probably related to the changes done in detuplification / insertion of implicit args, and the time when it is done relative to the insertion of function logging calls. No time to investigate further at the moment, so I have disabled `--trace` from the CI run of catala-examples ([here](https://github.com/CatalaLang/catala/pull/855/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R347))
- [ ] (**NEW**) There are spurious `ERROR: hd` appearing in catala-examples... puzzling
- [ ] (**NEW**) The tests on catala-examples pass, but the tweaky Makefile rules that were used to generate libraries (or in this specific case, Python packages) need to be adapted ; they should probably be based on `_targets/` anyway, but were still picking up files from `_build/`

## Also note...

This supersedes and includes #852, which should be dropped when this is finalised.

¹ A notable exception is `catala_runtime.cmi`, which is installed ; but we do need OCaml anyway. The reason is that, in order for the interpreter to be able to load Catala modules compiled to `cmxs`, the interpreter and module need to agree on the interface of the runtime since they will share the implementation ; and that agreement is dependent on the specific hash of the cmi file.


AltGr:
Interesting, we had a `#20 21.17 Error: Sedlex: Malformed ASCII string` due to an `é` appearing in a sedlex regexp... but that wasn't triggered locally. My version of sedlex is 3.3 instead of 3.2 but there doesn't appear to be anything related to this in their changelog.
(if we could be rid of the manual hex unicodes in the sedlex files, they would be quite a bit nicer to maintain)

PR #860, opened 2025-09-16
flake: Update and simplify the configuration

mrdev023:
Fix the flake configuration with latest version of catala.

Maybe some dependencies is no longer required so if you have any suggestion. Don't hesitate

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1627913399,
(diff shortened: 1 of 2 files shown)


adelaett:
ok to merge.

It might be a good idea to directly submit a version of Catala to nixpkgs, since nobody anymore uses the nix version in the dev team.

mrdev023:
@AltGr Someone can try with more example if the package works well ?

```
nix build github:mrdev023/nixpkgs/add-catala#catala
```

And to prepare for the next release, do you have made changes on dependencies ?

Issue #822, opened 2025-06-17
Create non-polymorphic standard library modules


denismerigoux:
* [x] operations on money (excess of, min, max, pro-rata with the , rounding to the cent, up or down, etc.). See https://github.com/CatalaLang/catala-examples/blob/0963304b75719dc0bfdd64824518251e855005ae/impot_revenu/oracles.catala_fr#L76-L84 for pro-rata
* [x] operations on decimals (min, max, rounding to the cent, up or down, etc.)
* [x] operations on periods (intersection, union, months contained in period). See CNAF code for inspiration
* [x] operations about months in the year with a sum type for months and links to the `date` type
* [x] operations on associated lists

These modules can and should contain external operations. Each external operation should have backend implementations in OCaml, C, Java and Python. Builtin operators should be deprecated in presence of a similar standard library operation, in order to reduce the size of the runtime. 

Important : the `stdilib` should be versioned. If we introduce breaking changes to the API downstream, the old version should still be available under an old name.

AltGr:
Documentation is a crucial point of a stdlib, as well : we should probably develop doc annotations in parallel with the lib itself.

PR #867, opened 2025-09-19
Improve stdlib

denismerigoux:

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/17

### DIFF
--- a/stdlib/list_en.catala_en
+++ b/stdlib/list_en.catala_en
@@ -17,9 +17,11 @@ declaration sequence
   equals
   List_internal.sequence of begin, end
 
-## if `index` is between `1` and `count of lst`, returns `Present` with the
-## element of the list at that index. Otherwise, returns `Absent`.
+## Returns the element of the list at the given index, encapsulated within
+## constructor `Présent`
 ## **Example**: `nth_element of [$101; $102; $103], 2` is `Present content $102`
+## **Edge-case**: if the index is less than 1, or otherwise outside of the list,
+## `Absent` is returned
 declaration nth_element
   content optional of anything of type t
   depends on lst content list of anything of type t,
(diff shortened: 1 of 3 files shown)


denismerigoux:
Fixes #822

Makefile
```diff
@@ -236,7 +236,7 @@ tests: test
 TEST_FLAGS_LIST = ""\
 -O \
 --lcalc \
---lcalc,--closure-conversion,-O \
+--lcalc,-O \
```
AltGr: Erf, indeed, externals won't cope well with changing the call convention of functions. That test was useful though, I'll see if we can find another solution not disable it entirely.

(otherwise, there's still hope that closure conversion bugs will be caught by the C backend, but:
   - that includes a smaller number of tests
   - it's much less helpful for debugging them)

denismerigoux:
The CI passes but I had to do two things : 
* [Disable testing of catala-examples with closure conversion on ](https://github.com/CatalaLang/catala/pull/867/commits/a83dec4f433c961fed1f2e27b9dd58b04231e1ee), which is fine, we cannot let this in presence of externally-implemented modules in OCaml who assume a specific interface, wherease `--closure-conversion --lcalc` changes that interface.
* [Fix typing unification](https://github.com/CatalaLang/catala/pull/867/commits/c255a4715c8e95c052a88b87780773099d3db18b); this changed I'm more skeptical of. I had to change this because there was a failure in when executing https://github.com/CatalaLang/catala/blob/master/tests/typing/good/option.catala_en with `dune exec catala -- interpret tests/typing/good/option.catala_en  --whole-program -d`. See the commit message for more information.

compiler/shared_ast/typing.ml
```diff
@@ -318,13 +318,21 @@ let rec union
           t2))
   | TVar v1, _ ->
     let t =
-      match Env.get_tvar env v1 with None -> t2 | Some t1 -> union t1 t2
+      match Env.get_tvar env v1 with
+      | None -> t2
+      | Some t1 ->
+        Env.set_tvar env v1 t2;
+        union t1 t2
```
AltGr, line 325: Ehem, it seems I fixed the var-var case in 0c54026 but forgot this more generic instance ; the proposed fix seems ok but I'll go through this function again.

The normal way to resolve type variables is through `get_ty` which already solves this kind of issues completely by handling equivalence classes of type variables: only this `union` function goes through variable aliases directly because it's low-level. I'll replicate the `get_ty` mechanism (or generalise it) in the `union` recursion to make the code simpler, with more obvious termination arguments.

The other bonus is that this should give better erroring if by a bug we do end up with a recursive type (e.g. `<t> = list of <t>`)

Makefile
AltGr: Note that the line should be removed rather than kept without the flag, it becomes redundant

AltGr: Actually I think externals might work (they are not expected to be closures, just functions, closure conversion is done later on) ; but it's the basic operators `fold` or `map` that don't have the right type because they should accept closures as arguments.

AltGr: (have a hack locally that fixes these tests)

PR #871, opened 2025-09-24
catala.opam tweaks

Octachron:
This small fixes remove constraints from the local opam files that seems too strict, at least currently.

### DIFF
--- a/catala.opam
+++ b/catala.opam
@@ -30,7 +30,6 @@ depends: [
   "cppo" {>= "1"}
   "dates_calc" {>= "0.0.7"}
   "dune" {>= "3.13"}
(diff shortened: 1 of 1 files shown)


catala.opam
```diff
@@ -30,7 +30,6 @@ depends: [
   "cppo" {>= "1"}
   "dates_calc" {>= "0.0.7"}
   "dune" {>= "3.13"}
-  "js_of_ocaml-ppx" {= "4.1.0"}
```
denismerigoux, line 33: Unfortunately this is because of https://github.com/CatalaLang/catala/issues/442, we need to debug that

PR #877, opened 2025-10-03
Many bugfixes, mainly related to polymorphism

AltGr:
See the individual patches for details

### DIFF
--- a/build_system/dune
+++ b/build_system/dune
@@ -42,7 +42,9 @@
   (:standard
    (:include custom_linking.sexp)))
  ; These are actually runtime deps and not link deps
- (link_deps %{bin:catala} ../runtimes/ocaml/catala_runtime.cmi)
+ (link_deps
+  %{bin:catala}
+  (alias ../runtimes/runtimes))
  (libraries clerk_driver)
  (modules clerk)
  (package catala))
--- a/compiler/catala_utils/file.ml
+++ b/compiler/catala_utils/file.ml
@@ -156,9 +156,12 @@ let find_in_parents ?cwd predicate =
   let cwd = match cwd with None -> Sys.getcwd () | Some cwd -> cwd in
   let home = try Sys.getenv "HOME" with Not_found -> "" in
   let rec lookup dir rel =
-    if predicate dir then Some dir, rel
-    else if dir = home then None, Filename.current_dir_name
-    else
+    match predicate dir with
+    | true -> Some dir, rel
+    | exception (Unix.Unix_error _ | Sys_error _) ->
+      None, Filename.current_dir_name
+    | false when dir = home -> None, Filename.current_dir_name
+    | false ->
       let parent = Filename.dirname dir in
       if parent = dir then None, Filename.current_dir_name
       else lookup parent (rel / Filename.parent_dir_name)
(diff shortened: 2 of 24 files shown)


runtimes/c/catala_runtime.c
```diff
@@ -637,15 +637,21 @@ CATALA_BOOL o_eq_dat_dat (CATALA_DATE x1, CATALA_DATE x2) {
...
-  if (dc_period_years(x1) == dc_period_years(x2) &&
-      dc_period_months(x1) == dc_period_months(x2) &&
-      dc_period_days(x1) == dc_period_days(x2))
+  long int y1 = dc_period_years(x1);
+  long int m1 = dc_period_months(x1);
+  long int d1 = dc_period_days(x1);
+  long int y2 = dc_period_years(x2);
+  long int m2 = dc_period_months(x2);
+  long int d2 = dc_period_days(x2);
+  if (y1 == y2 && m1 == m2 && d1 == d2)
```
denismerigoux, line 646: Indeed, we have eager semantics for the logical operators...

compiler/catala_utils/file.ml
```diff
@@ -432,7 +435,10 @@ let scan_tree f t =
   in
   let not_hidden t = match t.[0] with '.' | '_' -> false | _ -> true in
   let rec do_dir d =
-    Sys.readdir d
+    (try Sys.readdir d
+     with Sys_error _ ->
+       Message.debug "Cannot read %s, skipping" t;
```
denismerigoux, line 440: Very good, should solve the problem reported on Zulip last week.

build_system/dune
denismerigoux, line 1: I guess this is purely to help us during compiler development

tests/typing/bad/poly_eq.catala_en
```diff
@@ -0,0 +1,28 @@
...
+        if elt = needle then Present content elt
+        else found
+```
+
+```catala-test-cli
+$ catala typecheck
+┌─[ERROR]─
+│
+│  Equality cannot be resolved at this point: the type of the operands is not
+│  fully known.
```
denismerigoux, line 20: I would have put : "The left and right hand side of this equality are polymorphic and can have any type. Hence, we cannot test if they're equal since we don't know their shape in advance. This comparison is thus forbidden."

denismerigoux (approved):
Thank you very much!!!

compiler/catala_utils/file.ml
AltGr: Hopefully 🤞🏿

build_system/dune
AltGr: Yes. Without it `dune exec -- clerk ...` will fail because of missing files, and it's not obvious what to build to get them.

tests/typing/bad/poly_eq.catala_en
AltGr: Won't users be more interested in something actionable rather than a more technical explanation ? I was actually hesitant to write something along the lines of "You will need to write specific functions for the different concrete types that you need to handle at this point"

We should be able to lift this limitation in the future when we reenable the monomorphisation pass ; but at the moment, without this check, it would pass in the interpreter but fail once compiled.

denismerigoux: Yes indeed so we can put : 

> The left and right hand side of this equality are polymorphic and can have any type. Hence, we cannot test if they're equal since we don't know their shape in advance. This comparison is thus forbidden, you need to write specific functions for the different concrete types that you need to handle at this point.

PR #881, opened 2025-10-10
Fix --whole-program option

vincent-botbol:
(Re-open of #878)

When compiling using the `--whole-program` option, some edge-cases are broken. Three main isssues have been identified:
- Inconsistent module path qualification between separate compilation and whole-program (e.g., in `catala-example`, `clerk test -W impot_revenu/tests/traitements_salaires.catala_fr`)
- Evaluation of external scope context variables yields an empty error (e.g., in `catala-example`, `clerk test -W impot_revenu/tests/prorata.catala_fr`)
- Whole-program fails due to ambiguous constructor in some context (e.g., in `catala-example`, `clerk test -W aides_logement/tests/tests_eligibilite_apl.catala_fr`)

This PR aims to fix these issues.

Besides, calling the `--whole-program` option directly through catala (or through clerk like `clerk test -c-w`)  tries to also inline the stdlib which confuses some cram tests or evens tries to compile the stdlib using the whole-program which yields some errors. Hence, this PR adds a `--whole-program` / `-W` to `clerk test` and `clerk run`. Clerk still builds the modules dependencies which is redundant with whole-program but this will be improved later on.

### DIFF
--- a/compiler/driver.ml
+++ b/compiler/driver.ml
@@ -35,6 +35,8 @@ let load_modules
     program :
     ModuleName.t Ident.Map.t
     * (Surface.Ast.module_content * ModuleName.t Ident.Map.t) ModuleName.Map.t =
+  Message.with_delayed_errors
+  @@ fun () ->
   let stdlib_root_module lang =
     let lang = if Global.has_localised_stdlib lang then lang else Global.En in
     "Stdlib_" ^ Cli.language_code lang


build_system/clerk_driver.ml
```diff
@@ -1278,12 +1288,23 @@ let test_cmd =
...
+      test_flags
+      report
+      out
+      file
+      whole_program =
     let catala_opts =
       catala_opts
       @ List.fold_right (fun dir opts -> "-I" :: dir :: opts) include_dirs []
     in
     let test_flags = List.filter (( <> ) "") test_flags in
```
AltGr, line 1304: I am curious with what the difference between `--whole-program` and `--test-flags=--whole-program` would be, can they be used separately, in conjunction ?

PR #886, opened 2025-10-15
Tests and fixes for options

AltGr:
- Fixed a typing regression that would cause problems in the C backend due to underspecified types
- Updated the python backend to use a custom option type that respects nesting
- Fixed `mypy` on the python runtime (`make validate-py-runtime`) and some minor improvements


denismerigoux (approved):

AltGr:
ah my ocamlformat got inadvertently uninstalled 🙄

Issue #884, opened 2025-10-14
Tests for the externals of the stdlib


AltGr:
Unit tests for all external functions are mandatory to ensure consistency between all our backends. 

These sit in `tests/stdlib`.

- [x] date_internal (`of_ymd`, `to_ymd`, `last_day_of_month` ; note that in this case it might overlap with older tests that used to test them as built-ins, maybe they could just be moved)
- [x] list_internal (`sequence`, `nth_element`, `remove_nth_element`)
- [x] period_internal (`sort`, `split_by_month`, `split_by_year`)

PR #887, opened 2025-10-15
Stdlib: Add tests for Period externals

AltGr:
Adds the missing tests to complete #884.

Won't pass until the module is actually implemented in C, Python and Java of course.


vincent-botbol (approved):
LGTM. Perhaps we should discuss whether the end period should be included or not before merging. It feels like an important design choice.

PR #890, opened 2025-10-17
Quick fix for non-latin idents

AltGr:
A better solution would rely on a transliteration where needed, preservation for the backends that support it ; this is simpler in that it will replace the names, which is robust but makes the generated code much less readable.


vincent-botbol (approved):
LGTM

PR #897, opened 2025-10-23
C backend: fix content access on matched options

AltGr:
Fixes a bug detected on `tests/stdlib/period.catala_en` on the `stdlib++` branch.


denismerigoux (approved):
Thanks for fixing this option shenanigan! Our cross-backend testing makes a good job at catching this sort of errors, nice :)

PR #899, opened 2025-10-24
Retrieve version correctly for catala installed from opam release

denismerigoux:


compiler/catala_utils/hash.ml
```diff
@@ -57,7 +57,18 @@ type full = { catala_version : t; flags_hash : Flags.t; contents : t }
...
   Flags.pass (fun flags_hash ->
-      { catala_version = !(Version.v : string); flags_hash; contents = t })
+      {
+        catala_version =
+          !(match Version.v with
+            | Some v -> v
+            | None ->
+              Option.value ~default:"dev"
+                (Option.map Build_info.V1.Version.to_string
+                   (Build_info.V1.version ()))
```
denismerigoux, line 67: I had to copy-paste this because of cyclic dependency between `Hash` and `Cli`.

AltGr:
I'd prefer setting CATALA_VERSION during build rather than relying on the complex dune-release machinery that patches generated binaries upon install: it's never completely clear to me what it does and when. Besides, if we start having 3 possible sources for the proper version, it's not clear to me what the priority should be.

For the issue with the opam package, adding `build-env: [ CATALA_VERSION = "%{version}%" ]` to the opam file should be enough. The "dev" version would remain for one attempting a manual build from the source tarball, without git or that variable set, though.

PR #900, opened 2025-10-28
Reorganize Github actions

denismerigoux:
See the thread here : https://github.com/CatalaLang/catala/commit/6d81675d332154a59c410880f5a0acbe44d38f9e#commitcomment-168978867


AltGr:
Thanks! My only complaint was to have the same name appear twice ; but I agree it's better to merge the two.

One point of having a separate workflow was that it could run directly on a non-master branch (while with master/pull_request_target, it can only be tested once merged). This was needed while first prototyping the config, but I agree this is better now!

PR #904, opened 2025-11-04
Fix small Java polymorphism bug

AltGr:
Java lambdas were sometimes missing a polymorphism annotation.

Also includes some small stdlib fixes I had pending.

### DIFF
--- a/compiler/scalc/to_java.ml
+++ b/compiler/scalc/to_java.ml
@@ -184,7 +184,8 @@ let format_visibility ppf = function
   | Private -> () (* nothing => package visibility *)
   | Public -> fprintf ppf "public "
 
-let rec format_typ ctx ppf (typ : typ) =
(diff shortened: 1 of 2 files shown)


stdlib/monthyear_fr.catala_fr
```diff
@@ -6,34 +6,32 @@
 ```catala-metadata
 déclaration structure MoisAnnée:
   donnée nom_mois contenu D.Mois
-  donnée numéro_année contenu entier
+  donnée année contenu entier
```
denismerigoux, line 9: Doesn't this breaks syntax highlighting on VSCode?

denismerigoux: And/or parsing since this is a keyword?

AltGr: it's not, "an" is

PR #907, opened 2025-11-10
Handle multiple errors during compilation passes

vincent-botbol:
This PR rework some of the delayed error mechanism and adds some ad hoc heuristics to enable more error discovery.

Conceptually, this PR transform the existing fatal errors where no satisfying value can be given to continue to delayed errors by catching the fatal errors in the caller. And, in some cases, introduce error specific ast nodes (`EBad`) & types (`TError`) to attempt to progress further.

In its simplest form, we filter out problematic rules and try to continue. In more complex scenarios, such as typing, we try to recover with the introduced error type to explore as deep as possible when it makes sense. Same idea applies when translating from an AST to another with the `EBad` node.

Empirically, it seems to work even though I noticed that spurious errors could be added up due to the existing error recovery mechanism. It looks manageable however.

Additionally, we add synchronization barriers that raises pending errors after every ad hoc error handling pass so we never should end up with any TError types or EBad AST nodes in the subsequent passes.

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -602,7 +602,10 @@ let wrap_to_delayed_error ?(kind = Generic) x f =
 let report_delayed_errors_if_any () =
   match global_errors.rev_delayed_errors with
   | [] -> ()
-  | rev_delayed_errors -> raise (CompilerErrors (List.rev rev_delayed_errors))
+  | rev_delayed_errors ->
+    (* reinitialize the global state for reentrancy *)
+    global_errors.rev_delayed_errors <- [];
+    raise (CompilerErrors (List.rev rev_delayed_errors))
 
 let combine_with_pending_errors content bt =
   List.rev ((content, bt) :: global_errors.rev_delayed_errors)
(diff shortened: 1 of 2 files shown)


compiler/shared_ast/type.ml
```diff
@@ -79,8 +81,10 @@ let rec compare (ty1 : t) (ty2 : t) =
   | _, TVar _ -> 1
   | TForAll _, _ -> -1
   | _, TForAll _ -> 1
-(* | TClosureEnv, _ -> -1
- * | _, TClosureEnv -> 1 *)
+  (* | TClosureEnv, _ -> -1
+   * | _, TClosureEnv -> 1 *)
+  | TError, _ -> -1
+  | _, TError -> -1
```
AltGr: The comment should be left at the end (it's here for completeness, but commented out to avoid a warning).
Ideally it should be uncommented and `TError` should be commented, to respect the ordering of the type definition.

AltGr (approved):
It's not yet very clear where the `wrap...` should be used and when the errors are fatal. I am also not sure about putting the wrappers in the commands of `Driver.Commands` and not directly in the functions of `Driver.Passes` (this may change the way plugins should call them?)

This said, this looks good to me.

compiler/shared_ast/type.ml
vincent-botbol: Got it, did that.

PR #909, opened 2025-11-12
Fix reentrancy issue on new delayed errors state

vincent-botbol:
This PR fixes an issue introduced by #907 that accumulates delayed errors when the compiler is re-used in the same process (e.g., the LSP).


vincent-botbol:
LSP build is fixed by https://github.com/CatalaLang/catala-language-server/pull/165 which is tightly related by this PR.

Issue #913, opened 2025-11-24
Too many non-constant constructors in OCaml


vincent-botbol:
Compilation to OCaml fails when there are too many constructors.
Possible solution: remove the unit from empty enumeration or internally split too large enumerations.

```
File "_build/parties-communes/01.donnees/ocaml/Enumerations.mli", lines 137-447, characters 2-19:
137 | ..type t =
138 |     | AADN1 of unit
139 |     | ADIN1 of unit
140 |     | ADIN2 of unit
141 |     | ADIN3 of unit
...
444 |     | T07N7 of unit
445 |     | T08N8 of unit
446 |     | T09N9 of unit
447 |     | T10N0 of unit
Error: Too many non-constant constructors
       -- maximum is 244 non-constant constructors
ninja: build stopped: cannot make progress due to previous errors.
```

PR #924, opened 2025-12-17
OCaml: use constant constructors for constructors with empty payload

AltGr:
This is a more targeted fix for #913 than what #914 proposes: only the OCaml backend is affected on the specific "constructor with no payload" case.

It is mostly orthogonal though: if there is more reason to change the representation in the compiler, #914 remains valid.

The hard part was to adjust the dark magic in the live (native runtime)/(interpreter) translations to account for the change in the OCaml data representation.

  * [x] The corresponding PR is: CatalaLang/catala-examples/pull/37

* LSP sister PR: CatalaLang/catala-language-server/pull/181


AltGr:
Closes #913

PR #925, opened 2025-12-17
Some small optims

AltGr:
- fast-path for string comparisons with the same pointer
- improvement of equality expansion for constant constructors

### DIFF
--- a/compiler/lcalc/expand_op.ml
+++ b/compiler/lcalc/expand_op.ml
@@ -75,34 +75,49 @@ let rec resolve_eq ctx pos ty args m =
       match args with [arg1; arg2] -> arg1, arg2 | _ -> assert false
     in
     let constrs = EnumName.Map.find name ctx.ctx_enums in
-    let cases =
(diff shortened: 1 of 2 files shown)


compiler/catala_utils/string.ml
denismerigoux, line 1: Do you have a sense of how much speed gain we get thanks to that?

denismerigoux (approved):

compiler/catala_utils/string.ml
AltGr: For expand_op, we should gain a lot in code size (and thus compilation / interpretation time): the code generated is now linear in size instead of quadratic in the number of constructors, when considering only the constant constructors. That could be life-saving for the cases that caused problem with OCaml's 244 constructors limitation.

For the string comparison, Vincent reported performance issues on sets and it should alleviate them, but I haven't benched yet (from reading the code, the missing fast-path was the main difference between our function and the OCaml built-in comparison ; but the latter is also based on `memcmp` which we can't compete with)

PR #926, opened 2025-12-18
Python: use "elif" when appropriate instead of nested "if"s, allow long lines

AltGr:
This should solve a bug where too many levels of nesting would cause bad indentation, breaking the python syntax.
- ensure we configure `Format` without width limit to avoid the problem
- still limit the level of nesting by using `elif` when possible


compiler/scalc/to_python.ml
```diff
@@ -628,6 +637,10 @@ let format_program
     (fmt : Format.formatter)
     (p : Ast.program)
     (type_ordering : TypeIdent.t list) : unit =
+  (* Disable the formatting line length limit which may lead to broken
+     indentation (we need to disable all line breaking for this, since Format
+     doesn't accept max_indent >= margin) *)
+  Format.pp_set_geometry fmt ~max_indent:999_990 ~margin:1_000_000;
```
denismerigoux, line 643: +1

PR #928, opened 2025-12-19
Add Json encoding of scope inputs and --json-input to interpret command

vincent-botbol:
This PR adds support for the interpretation of non-empty inputs Catala
scopes using JSON-encoded input scope structures. This relies on the
`json-data-encoding` library that allows, given a context and a type, to
dynamically generate an encoding allowing the serialization of such
values from and to JSON.

As stated, the main goal is to add the capacity to call scopes with a
JSON-input. In the CLI, this takes the following form:
```
$ clerk run a_scope_with_input.catala_en --input '{"x":1, "y":2}' --scope S
# We can omit the --scope argument when non-ambiguous
```
The `--input` flag accepts either a JSON-string, a file, or `-` for stdin.

Note: all the described commands/flags are present for `catala` and `clerk` commands.

Scope context variables are optional in the JSON, same as option types.

A `--format` flag was also added that currently accepts either `human` or `json` to output results in JSON, e.g.,
```
$ clerk run tests/json/good/optionals.catala_en --output-format json --input='{"x":"112312908312938129831983", "y":true, "z": 3 }'  --quiet
{ "r": "112312908312938129831986" }
```
Note: this also work with no-input scopes

To retrieve the JSON schema of the input and output scope structures,
the command `clerk json-schema <file> --scope <scope>` was added,
e.g.,

<details>
<summary>clerk json-schema tests/json/good/enums.catala_en --scope T</summary>

```
  [
    {
      "$schema": "http://json-schema.org/draft-04/schema#",
      "title": "Scope T input",
      "description": "Input structure of scope T",
      "$ref": "#/definitions/T_in",
      "definitions": {
        "E": { "type": "string", "enum": [ "X", "Y", "Z" ] },
        "F": {
          "oneOf": [
            {
              "type": "object",
              "properties": { "A": { "oneOf": [ { "type": "integer", "minimum": -1073741824.0, "maximum": 1073741823.0 }, { "type": "string" } ] } },
              "required": [ "A" ],
              "additionalProperties": false
            },
            { "type": "string", "enum": [ "B" ] },
            { "type": "object", "properties": { "C": { "$ref": "#/definitions/E" } }, "required": [ "C" ], "additionalProperties": false }
          ]
        },
        "T_in": { "type": "object", "properties": { "e": { "$ref": "#/definitions/F" } }, "required": [ "e" ], "additionalProperties": false }
      }
    },
    {
      "$schema": "http://json-schema.org/draft-04/schema#",
      "title": "Scope T output",
      "description": "Output structure of scope T",
      "$ref": "#/definitions/T",
      "definitions": {
        "E": { "type": "string", "enum": [ "X", "Y", "Z" ] },
        "T": { "type": "object", "properties": { "r": { "$ref": "#/definitions/E" } }, "required": [ "r" ], "additionalProperties": false }
      }
    }
  ]
```
</details>

A more human-readable output are provided in error messages. See
`tests/json/bad/simple.catala_en` for an example.

The Catala values <=> JSON correspondance is described in
`compiler/shared_ast/encoding.mli`.

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/37

### DIFF
--- a/compiler/shared_ast/encoding.ml
+++ b/compiler/shared_ast/encoding.ml
@@ -123,75 +123,60 @@ let rat_encoding : Runtime.runtime_value encoding =
     ]
 
 let date_encoding : Runtime.runtime_value encoding =
+  let date_obj =
+    obj3
+      (req "year" (ranged_int ~minimum:0 ~maximum:9999 "years"))
+      (req "month" (ranged_int ~minimum:1 ~maximum:12 "months"))
+      (req "day" (ranged_int ~minimum:1 ~maximum:31 "days"))
+  in
   def "date" ~title:"Catala date"
-    ~description:
-      "Accepts JSON strings with the following format: YYYY-MM-DD, e.g., \
-       \"1970-01-31\""
-  @@ conv
-       (function
-         | Runtime.Date d -> Format.asprintf "%a" Dates_calc.format_date d
-         | v ->
-           Message.error ~internal:true
-             "Unexpected runtime value %a instead of date while encoding to \
-              JSON"
-             Runtime.format_value v)
-       (fun s -> Runtime.Date (Dates_calc.date_of_string s))
-       string
+  @@ union
+       [
+         case
+           ~description:
+             "Accepts strings with the following format: YYYY-MM-DD, e.g., \
+              \"1970-01-31\""
+           string
+           (function
+             | Runtime.Date d ->
+               Some (Format.asprintf "%a" Dates_calc.format_date d)
+             | v ->
+               Message.error ~internal:true
+                 "Unexpected runtime value %a instead of date while encoding \
+                  to JSON"
+                 Runtime.format_value v)
+           (fun s -> Runtime.Date (Dates_calc.date_of_string s));
+         case
+           ~description:
+             "Accepts date objects: {\"year\":<int>, \"month\":<int>, \
+              \"day\":<int>}"
+           date_obj
+           (function
+             | Runtime.Date d -> Some (Dates_calc.date_to_ymd d)
+             | v ->
+               Message.error ~internal:true
+                 "Unexpected runtime value %a instead of date while encoding \
+                  to JSON"
+                 Runtime.format_value v)
+           (fun (year, month, day) ->
+             Runtime.Date (Dates_calc.make_date ~year ~month ~day));
+       ]
 
 let duration_encoding : Runtime.runtime_value encoding =
-  let parse_duration s =
-    let s =
-      String.split_on_char ' ' s
-      |> List.filter (( <> ) String.empty)
-      |> String.concat ""
-    in
-    try
-      Scanf.sscanf s "%d%s" (fun n unit ->
-          match unit with
-          | "years" | "year" -> `Years n
-          | "months" | "month" -> `Months n
-          | "days" | "day" -> `Days n
-          | _ ->
-            raise (Json_encoding.Unexpected (unit, "years, months or days")))
-    with _ ->
-      raise (Json_encoding.Unexpected (s, "number and duration unit string"))
+  let encoding =
+    obj3 (dft "years" int 0) (dft "months" int 0) (dft "days" int 0)
+    |> conv
+         (function
+           | Runtime.Duration d -> Dates_calc.period_to_ymds d
+           | v ->
+             Message.error ~internal:true
+               "Unexpected runtime value %a instead of duration while encoding \
+                to JSON"
+               Runtime.format_value v)
+         (fun (years, months, days) ->
+           Runtime.Duration (Dates_calc.make_period ~years ~months ~days))
   in
-  def "duration" ~title:"Catala duration"
-    ~description:
-      "Accepts JSON strings with the following format: [X years|months|days], \
-       e.g., \"[2 months]\" or \"[3 days]\""
-  @@ conv
-       (function
-         | `Years y -> Format.sprintf "%d years" y
-         | `Months m -> Format.sprintf "%d months" m
-         | `Days d -> Format.sprintf "%d days" d
-         | `Raw d ->
-           let s = Format.asprintf "%a" Dates_calc.format_period d in
-           String.sub s 1 (String.length s - 2))
-       (fun s -> parse_duration s)
-       string
-  |> conv
-       (function
-         | Runtime.Duration d -> begin
-           match Dates_calc.period_to_ymds d with
-           | y, 0, 0 -> `Years y
-           | 0, m, 0 -> `Months m
-           | 0, 0, d -> `Days d
-           | _ -> `Raw d
-         end
-         | v ->
-           Message.error ~internal:true
-             "Unexpected runtime value %a instead of duration while encoding \
-              to JSON"
-             Runtime.format_value v)
-       (function
-         | `Years years ->
-           Duration (Dates_calc.make_period ~years ~months:0 ~days:0)
-         | `Months months ->
-           Duration (Dates_calc.make_period ~years:0 ~months ~days:0)
-         | `Days days ->
-           Duration (Dates_calc.make_period ~years:0 ~months:0 ~days)
-         | `Raw d -> Duration d)
+  def "duration" ~title:"Catala duration" @@ encoding
 
 let position_encoding =
   let p_encoding = obj2 (req "line" int32) (req "character" int) in
--- a/tests/json/good/dates.catala_en
+++ b/tests/json/good/dates.catala_en
@@ -1,51 +1,51 @@
 ```catala-metadata
 declaration scope S:
-  input w content date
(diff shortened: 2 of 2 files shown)


build_system/manpages.sexp
```diff
@@ -2,6 +2,7 @@
 (rule (alias man) (action (with-stdout-to clerk-build.1 (run %{bin:clerk} build --help=groff))))
 (rule (alias man) (action (with-stdout-to clerk-ci.1 (run %{bin:clerk} ci --help=groff))))
 (rule (alias man) (action (with-stdout-to clerk-clean.1 (run %{bin:clerk} clean --help=groff))))
+(rule (alias man) (action (with-stdout-to clerk-json-schema.1 (run %{bin:clerk} json-schema --help=groff))))
```
denismerigoux, line 5: Very good catch! I had forgot about this list.

catala.opam
```diff
@@ -53,6 +53,7 @@ depends: [
   "ocamlformat" {?cataladevmode & cataladevmode & = "0.26.0"}
   "obelisk" {?cataladevmode & cataladevmode}
   "conf-ninja" {post}
+  "json-data-encoding"
```
denismerigoux, line 56: Should we also add `yojson` as an explicit dependency here?

compiler/shared_ast/encoding.mli
denismerigoux, line 1: Very clean and nice, I love it!

compiler/driver.ml
```diff
@@ -462,7 +462,34 @@ module Commands = struct
...
+      | [] ->
+        Message.error
+          "The program has no exported scopes.@ Please specify option \
+           @{<yellow>--scope@}@ to execute a private scope@ or add \
+           ```catala-metadata to a scope declaration."
+      | _ :: _ :: _ ->
+        Message.error
+          "The program defines multiple scopes but only one was expected.@ \
+           Please specify option @{<yellow>--scope@} to explicit what to \
+           execute.@ The program defines the following scopes:@ @[<hv 4>%a@]"
```
denismerigoux, line 473: Aaaaaah yes the lists of scopes that you can call directly in the error message <3<3<3

compiler/driver.ml
```diff
@@ -1435,6 +1509,52 @@ module Commands = struct
...
+          Encoding.scope_output_encoding scope prg.decl_ctx output_ty
+        in
+        Json_encoding.schema encoding |> Schema_repr.to_json
+      in
+      Format.fprintf (Message.std_ppf ()) "%a@\n"
+        (Yojson.Safe.pretty_print ~std:true)
+        (`List [scope_input_schema_json; scope_output_schema_json])
+    in
+    Cmd.v
+      (Cmd.info "json-schema" ~man:Cli.man_base
```
denismerigoux, line 1529: Couldn't this be factorized with the equivalent command in `clerk` ?

tests/json/bad/simple.catala_en
```diff
@@ -0,0 +1,71 @@
...
+│    At /x, no case matched:
+│      Unexpected string instead of number
+│      Unexpected string instead of numeric string
+│  
+│  Expected JSON object of the form:
+│    { "s": $Str,
+│      "x": integer ∈ [-2^30, 2^30] || string,
+│      "y": boolean,
+│      "z": integer ∈ [-2^30, 2^30] || number || string }
+│    $Str: { "x": integer ∈ [-2^30, 2^30] || string }
```
denismerigoux: <3<3<3<3

tests/json/bad/simple.catala_en
```diff
@@ -0,0 +1,71 @@
...
+┌─[ERROR]─
+│
+│  Failed to validate JSON:
+│    At /s, missing object field x
+│  
+│  Expected JSON object of the form:
+│    { "s": $Str,
+│      "x": integer ∈ [-2^30, 2^30] || string,
+│      "y": boolean,
+│      "z": integer ∈ [-2^30, 2^30] || number || string }
```
denismerigoux: So why the `2^30` bound on integers? This is pretty low, I can expect this to be broken in prod for rich people for instance. Can we improve it?

build_system/manpages.sexp
vincent-botbol: This is automatically generated :) Credits to @AltGr

catala.opam
vincent-botbol: It's already one a few lines above : 
```
  "yojson" {>= "2.1.0" & < "3"}
```

compiler/driver.ml
vincent-botbol: I thought a bit about it but did not find a convincing solution..

tests/json/bad/simple.catala_en
vincent-botbol: I found the int encoding cleaner code-wise but actually I overthought it. I switched back to int53 (which is apparently the recommended accepted range by the RFC). It covers most cases and, for big int, there is the string encoding that accepts arbitrary precision integers.

rprimet:
> > Really nice! One thing that may be confusing is the naming of the output format: ` --format json` and the fact that an `--output` option already exists, which outputs a ninja build file. I expect some unwary users will run `clerk run ... --output=json` and be surprised at the results :-)
> > (since outputing the ninja build file may be a niche use case, maybe renaming it to `--output-ninja-build-file` and renaming --format to --output might make sense because it would match `input` ?)
> 
> Agreed, I don't know what consensus is on that but @AltGr and myself agree that `-o/--output` should refer to an explicit output file. I changed --format to `--output-format/-F` and renamed `--output/-o` to `--output-ninja-file` (removing the `-o`). What do you think?

I'm OK with that!

build_system/clerk_cli.ml
```diff
@@ -242,6 +242,17 @@ let scope =
           "Used with the `run` command, selects which scope of a given Catala \
            file to run.")
 
+let scope =
+  Arg.(
+    required
+    & opt (some string) None
+    & info ["s"; "scope"] ~docv:"SCOPE"
+        ~doc:
+          "Used with the `run` command, selects which scope of a given Catala \
```
AltGr:
```suggestion
          "Selects which scope of a given Catala \
```
The option already only appears in the doc for `clerk run`, so no need to explicit it.

build_system/clerk_driver.ml
```diff
@@ -935,7 +941,7 @@ let build_test_deps ~config ~backend files_or_folders nin_ppf items var_bindings
         let filter item =
           if is_dir then
             String.starts_with ~prefix:(file / "") item.Scan.file_name
-            && Lazy.force item.Scan.has_scope_tests
+            && (Lazy.force item.Scan.has_scope_tests || not test_only)
```
AltGr: Better put the `force` on the right of the `||` no ?

compiler/driver.ml
```diff
@@ -1421,6 +1495,52 @@ module Commands = struct
...
+  let json_schema_cmd =
+    let f options includes stdlib ex_scope =
+      let mark = Expr.typed in
+      let prg, _ =
+        Passes.dcalc options ~includes ~stdlib ~optimize:false
+          ~check_invariants:false ~autotest:false ~typed:mark
+      in
+      Interpreter.load_runtime_modules
+        ~hashf:Hash.(finalise ~monomorphize_types:false)
+        prg;
```
AltGr:
```suggestion
```
Loading the runtime modules shouldn't be needed just to print the schema ?

compiler/shared_ast/encoding.ml
```diff
@@ -0,0 +1,613 @@
...
+         | Runtime.Date d -> Format.asprintf "%a" Dates_calc.format_date d
+         | v ->
+           Message.error ~internal:true
+             "Unexpected runtime value %a instead of date while encoding to \
+              JSON"
+             Runtime.format_value v)
+       (fun s -> Runtime.Date (Dates_calc.date_of_string s))
+       string
+
+let duration_encoding : Runtime.runtime_value encoding =
```
AltGr, line 165: This won't allow composite durations (e.g. `3 months + 1 days`). Those are already not possible in Catala literals, but that might become an issue here.

build_system/clerk_cli.ml
vincent-botbol: Fixed it alongside two other similar descriptions.

build_system/clerk_driver.ml
vincent-botbol: Indeed, done.

compiler/driver.ml
vincent-botbol: Good catch, fixed it

compiler/shared_ast/encoding.ml
vincent-botbol: I reworked the date & duration encodings in the latest commit.

PR #929, opened 2025-12-22
Refactor module loading

### LINKED ISSUE #915: Make catala plugins aware of built-in (stdlib) modules
To my knowledge, Catala plugins receive no information on built-in modules, and importing them explicitly is an error.

Which could be fixed by either:

- Make importing stdlib modules explicitly OK (i.e. a no-op)
- Supply a blacklist of implicitly imported modules to plugins so that they can skip these imports when generating code

AltGr:
Module loading had precently been quickfixed by preemptively loading the standard library in the expected language.

This rationalises the code by getting back to on-demand loading of all modules, including stdlib ; in particular, this means that `use`ing a module written in a different dialect won't lead to a weird error anymore.

This should also fix #915, @rprimet I'll let you check ?

### DIFF
--- a/compiler/driver.mli
+++ b/compiler/driver.mli
@@ -144,3 +144,16 @@ module Plugin : sig
     Catala_utils.Pos.attr option) ->
     unit
 end
+
+val load_modules :
+  Global.options ->
+  Global.raw_file list ->
+  stdlib:Global.raw_file option ->
+  ?more_includes:string list ->
+  ?allow_notmodules:bool ->
+  Surface.Ast.program ->
+  Shared_ast.ModuleName.t Shared_ast.Ident.Map.t
+  * (Surface.Ast.module_content
+    * Shared_ast.ModuleName.t Shared_ast.Ident.Map.t)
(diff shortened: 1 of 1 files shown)


build_system/clerk_report.ml
```diff
@@ -498,12 +498,22 @@ let summary ~build_dir tests =
...
+        "TOTAL" "RATIO";
+      let ratio ppf (m, n) =
+        let color =
+          let open Ocolor_types in
+          if m >= n then C4 green
+          else C24 { r24 = 255; g24 = m * 255 / n; b24 = 0 }
+        in
+        Format.pp_open_stag ppf (Ocolor_format.Ocolor_style_tag (Fg color));
+        Format.fprintf ppf "%8d %%" (m * 100 / n);
+        Format.pp_close_stag ppf ()
```
denismerigoux, line 511: Much color engineering, such beautiful reports <3

tests/stdlib/multilang.catala_fr
denismerigoux, line 1: Very nice!

PR #936, opened 2026-01-09
Fix code generator and explain plugin bugs

maxvonhippel:
This commit fixes two related issues preventing the `explain` plugin from working:

1. Fix OCaml code generator for `Optional.Absent` (`compiler/lcalc/to_ocaml.ml`)
   - Added special case to generate `Optional.Absent` without `unit` argument
   - Previously generated `Optional.Absent()` which failed compilation
   - This was causing `stdlib` `Period` modules to fail compilation

2. Fix `explain` plugin runtime errors (`compiler/plugins/explain.ml`)
   - Added error handling for standalone test scopes (line 761)
   - Added support for general function applications (line 952)
   - Fixed assertion failures on unexpected expression types

The explain plugin now works end-to-end and successfully generates .dot and .svg visualization files.

## Checklist

None of the below apply as far as I can tell. I don't believe the `explain` plugin is even documented.

* [ ] Syntax cheat sheet at `doc/syntax/syntax_*.catala_*` and `doc/syntax/catala_*.typ`.
* [ ] Syntax highlighting plugins in `syntax_highlighting/` (all of them).

### DIFF
--- a/compiler/lcalc/to_ocaml.ml
+++ b/compiler/lcalc/to_ocaml.ml
@@ -309,6 +309,11 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
   | EInj { e = ELit LUnit, _; cons; name } ->
     Format.fprintf fmt "@[<hov 2>%a.%a@]" format_to_module_name (`Ename name)
       format_enum_cons_name cons
+  | EInj { cons; name; _ }
+    when EnumName.equal name Expr.option_enum
+         && EnumConstructor.equal cons Expr.none_constr ->
+    Format.fprintf fmt "@[<hov 2>%a.%a@]" format_to_module_name (`Ename name)
+      format_enum_cons_name cons
   | EInj { e; cons; name } ->
     Format.fprintf fmt "@[<hov 2>%a.%a@ %a@]" format_to_module_name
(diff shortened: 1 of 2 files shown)


compiler/lcalc/to_ocaml.ml
```diff
@@ -309,6 +309,11 @@ let rec format_expr (ctx : decl_ctx) (fmt : Format.formatter) (e : 'm expr) :
   | EInj { e = ELit LUnit, _; cons; name } ->
     Format.fprintf fmt "@[<hov 2>%a.%a@]" format_to_module_name (`Ename name)
       format_enum_cons_name cons
+  | EInj { cons; name; _ }
+    when EnumName.equal name Expr.option_enum
+         && EnumConstructor.equal cons Expr.none_constr ->
+    Format.fprintf fmt "@[<hov 2>%a.%a@]" format_to_module_name (`Ename name)
+      format_enum_cons_name cons
```
vincent-botbol, line 316: This case seem to be dead-code. The option's none constructor carries, by construction, a direct `ELit LUnit` value as expression. Hence, the previous match-case already handles it. We can safely remove it. 
The PR that removes the `()` values from non-value constructors was merged not too long ago, perhaps you were in an intermediate state when you pushed this fix?

Issue #931, opened 2026-01-08
Add an error message in case of impossible or assertion failures


denismerigoux:
`impossible` and `assertion` failure should, in the interpreter as well as in each one of the backends, raise a user-specified error message when the program crashes.

This error message could be specified by the user as an attribute: 

```
#[debug.message "A married household cannot have only one individual"]
assertion (not (household.status = Married and number of household.members = 1)) 
```

And with `impossible`:

```
match household.status with pattern
-- Single : 
-- Married : 
  if number of household.members = 1 then 
    #[debug.message "A married household cannot have only one individual"]    
    impossible 
  else
    sum money of (map each member among household.members to member.income) 
```

Issue #932, opened 2026-01-08
Generalize assertions to expressions


denismerigoux:
Right now, `assertion`s can only be put as a statement under a `scope` usage, and as such can only refer to scope variables.

But we want to generalize assertions so that they can be also placed deep inside expressions, with the syntax: 

```
let x = 4 in 
assertion x >= 0 in 
x + 2
```

French version would be `assertion ... dans`.

Semantic would still be the same as statement-assertions. #931 should work with these new assertions as well.

AltGr:
This should also be decided while taking #691, and the keywords we may want to add there, into account. It might even be considered to keep `assert` for expressions, but remove it from scopes and use new keywords there instead.

PR #938, opened 2026-01-14
Allow asserts in expressions; add support for error messages

AltGr:
Closes: #931, #932

The first part of this PR adds `ensure condition in expression` (**syntax to be changed**) which introduces an inline assertion in an expression.

The second part adds a `#[error.message = "Custom test"]` attribute that is allowed
- in front of assertions (both scope-level and inline)
- in front of the `impossible` keyword

When the corresponding error is triggered at that point, the message is embedded into the error structure and printed out besides the actual error and its position.
This is done in the interpreter, and propagated to all backends, which also have a more unified, one-line error output format now.

Caveat: the interpreter messages are tested ; all the backends have been tested manually ; however, we don't support negative automated tests for validating non-regression in the error messages from the backends. Supporting this would be quite involved and probably overkill... a poor man's solution could be to insert a custom check in the Makefile, but that may be annoying to maintain. Weighing that against the risk of regression, it doesn't feel worth it at the moment.

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/42
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-examples/pull/38
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-language-server/pull/188

    - [x] The corresponding PR is: https://github.com/CatalaLang/tree-sitter-catala/pull/19
    - [x] The corresponding PR is: https://github.com/CatalaLang/catala-format/pull/12

### DIFF
--- a/compiler/desugared/name_resolution.ml
+++ b/compiler/desugared/name_resolution.ml
@@ -118,7 +118,7 @@ type attribute_context =
 let attribute_parsers :
     (string
     * string list
-    * attribute_context list
+    * (attribute_context -> bool)
     * (pos:Pos.t -> attr_value -> Pos.attr option))
     list
     ref =
@@ -144,9 +144,7 @@ let handle_extra_attributes context plugin path value pos =
         plugin;
       None
     | handlers -> (
-      match
-        List.find_opt (fun (_, _, ctx, _) -> List.mem context ctx) handlers
-      with
+      match List.find_opt (fun (_, _, ctx, _) -> ctx context) handlers with
       | None ->
         Message.warning ~pos
           "Attribute @{<magenta>#[%s.%s]@} is not allowed in this context"
--- a/compiler/desugared/name_resolution.mli
+++ b/compiler/desugared/name_resolution.mli
@@ -128,7 +128,7 @@ type attribute_context =
 val register_attribute :
   plugin:string ->
   path:string list ->
-  contexts:attribute_context list ->
+  contexts:(attribute_context -> bool) ->
   (pos:Pos.t -> Shared_ast.attr_value -> Pos.attr option) ->
   unit
 (** Registers a new callback for the translation of the [#[plugin.path]]
--- a/compiler/driver.mli
+++ b/compiler/driver.mli
@@ -138,7 +138,7 @@ module Plugin : sig
   val register_attribute :
     plugin:string ->
     path:string list ->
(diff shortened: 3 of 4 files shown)


compiler/scopelang/from_desugared.ml
```diff
@@ -736,9 +736,15 @@ let translate_rule
...
     (* we unbox here because assertions do not have free variables (at this
-       point Bindlib variables are only for function parameters)*)
+       point Bindlib variables are only for function parameters) *)
     let assertion_expr = translate_expr ctx (Expr.unbox assertion_expr) in
-    [Ast.Assertion (Expr.unbox assertion_expr)]
+    [
+      Ast.Assertion
+        {
+          e = Expr.unbox assertion_expr;
+          pos = Mark.get (D.AssertionName.get_info a_name);
```
denismerigoux, line 745: This is where we catch the attribute `#[error.message = ...]`?

compiler/surface/ast.mli
```diff
@@ -190,6 +190,7 @@ and naked_expression =
   | Dotted of expression * (path * lident Mark.pos) Mark.pos
       (** Dotted is for both struct field projection and sub-scope variables *)
   | TupleAccess of expression * int Mark.pos
+  | Assert of expression * expression * Pos.t
```
denismerigoux, line 193: A little comment here: `Assert e1 e2` is `assertion e1 in e2`?

doc/syntax/cheat-sheet.typ
```diff
@@ -2,7 +2,7 @@
 
 #let render_title(title, subtitle) = {
     place(top+left, image("logo.svg", width: 30pt))
-    place(top+right)[v1.0.0 · Révision \#1 · ⓒ #datetime.today().year()]
+    place(top+right)[v1.0.1 pre · Révision \#1 · ⓒ #datetime.today().year()]
```
denismerigoux, line 5: Can you bump this version number everywhere else?

compiler/scopelang/from_desugared.ml
AltGr: For scope-level assertions, yes

doc/syntax/cheat-sheet.typ
AltGr: Indeed, we'll need to do that, but probably not in this PR.
My motivation here was to avoid having published cheat-sheets advertising features that don't exist in the version it claims.

PR #946, opened 2026-01-22
Add Integer.sum, etc. to the stdlib

AltGr:
and remove the older syntax 'sum integer of...' from the syntax sheet as a first step to deprecation.

The syntax is still supported, but the parser was patched to no longer recognise `sum` as a keyword outside of the `sum <type> of` construct. Therefore, the change shouldn't be breaking, but leave the time to replace the old syntax with the call to the stdlib functions.

The `Duration` stdlib module was also added, since it wasn't there yet and `sum duration of ...` is supported.

The tools and grammars don't need any change (this doesn't add syntax, and they're not affected by `sum` appearing in places where it's not a keyword).

A PR updating the book to advertise the new syntax instead of the old one is proposed at https://github.com/CatalaLang/catala-book/pull/45

### DIFF
--- a/tests/array/good/aggregation_3.catala_en
+++ b/tests/array/good/aggregation_3.catala_en
@@ -13,8 +13,12 @@ scope S:
     assertion (map each i among [1; 2; 3] such that i > 2 to (i + 2) ) = [5]
 
 # Sum
+    # Old, deprecated syntax
     assertion sum integer of [1; 2; 3] = 6
     assertion sum integer of (map each i among [1; 2; 3] to (i + 2)) = 12
+    # New syntax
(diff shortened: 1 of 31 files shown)


tests/array/good/aggregation_3.catala_en
```diff
@@ -13,8 +13,12 @@ scope S:
     assertion (map each i among [1; 2; 3] such that i > 2 to (i + 2) ) = [5]
 
 # Sum
+    # Old, deprecated syntax
```
denismerigoux, line 16: Could you write an issue assigned to milestone `1.0.2` that reminds us to deprecate the old sum syntax in the next release?

PR #948, opened 2026-01-22
Stdlib: refactor day_of_week computation

AltGr:
The written implementation was a good example of proper Catala code, but in the case of the stdlib, we prefer functions over scopes. I gathered the existing code under a single function using local variables, and also switched from tuples to lists for dynamic access.

Added some tests as well.

### DIFF
--- a/tests/backends/simple.catala_en
+++ b/tests/backends/simple.catala_en
@@ -39,13 +39,13 @@ $ catala c
 
 #include <Stdlib_en.h>
 #include <Date_en.h>
+#include <List_en.h>
 #include <Duration_en.h>
 #include <MonthYear_en.h>
 #include <Period_en.h>
 #include <Money_en.h>
 #include <Integer_en.h>
 #include <Decimal_en.h>
-#include <List_en.h>
 
 typedef struct Foo {
   CATALA_BOOL x;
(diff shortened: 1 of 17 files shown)


tests/backends/simple.catala_en
```diff
@@ -39,12 +39,12 @@ $ catala c
 
 #include <Stdlib_en.h>
 #include <Date_en.h>
+#include <List_en.h>
```
denismerigoux, line 42: Shouldn't we print the includes in the backends by alphabetical order rather than what we're doing now? I noticed refactors in the Stdlib and changes in the dependency graphs changes the order of module imports in the backends like here, this is annoying because it introduces noisy diffs even though semantically nothing should change.

AltGr: This uses a topographical ordering, since there may be interdependencies between modules.
Maybe we should limit cram tests that include potentially unstable outputs... I too was surprised by the size of the `clerk reset` diff. We should have at least one test per backend (so that we notice when there are changes, even small), but most cli tests with `catala ocaml` are probably superfluous. Note also that there is already some post-processing hacks to avoid diffs on changed source hashes, for example.

Cram tests using `--debug` are probably not such a good idea either, the debug output has no purpose of stability. We should probably implement debug scopes at some point, so that we could test on a specific one.

PR #952, opened 2026-01-28
Rewire JSOO pipeline and add online playground

rprimet:
Adds

- An updated JSOO pipeline 
- A web-based playground

The part that should be critiqued / reviewed most thoroughly is the first commit ( a54be1e0733675103ee9145ffb8b810a426ea61a ) that targets the ocaml side, since the playground will probably live in catala-book?

Can be tested at https://idyllic-kringle-f90220.netlify.app/learn.html or https://idyllic-kringle-f90220.netlify.app/learn.html?bookBase=https://book.catala-lang.org/fr for the French version

PR #955, opened 2026-02-03
Restore js_of_ocaml web interpreter

rprimet:
Supersedes #952 in part (other PR to come in catala-book for the web playground)

### DIFF
--- a/compiler/web/catala_web_interpreter.ml
+++ b/compiler/web/catala_web_interpreter.ml
@@ -130,15 +130,18 @@ let cleanup_files module_files =
       try Sys.remove path with Sys_error _ -> ())
     module_files
 
-(* Drain and return error notifications, keeping only warnings in the acc *)
-let drain_error_notifications () =
+(* Drain all accumulated notifications, emptying the accumulator. Returns
+   (warning_diags, error_notifs) where warning_diags are already converted to JS
+   diagnostic objects and error_notifs are raw lsp_errors. *)
+let drain_all () =
   let all = List.rev !notifications_acc in
-  let errors =
-    List.filter (fun (e : Message.lsp_error) -> e.kind <> Message.Warning) all
+  notifications_acc := [];
+  let warnings, errors =
+    List.partition (fun (e : Message.lsp_error) -> e.kind = Message.Warning) all
   in
-  notifications_acc :=
-    List.filter (fun (e : Message.lsp_error) -> e.kind = Message.Warning) all;
-  errors
+  let warning_diags = List.map (diagnostic_of_lsp_error "warning") warnings in
+  let error_notifs = errors in
+  warning_diags, error_notifs
 
 (* Extract JS position objects from a list of lsp_errors, filtering stdlib *)
 let positions_of_notifications notifs =
@@ -151,9 +154,9 @@ let positions_of_notifications notifs =
       | _ -> None)
     notifs
 
-(* Handle errors and return a list of error diagnostics *)
-let handle_error exn =
-  let error_notifs = drain_error_notifications () in
+(* Format an exception into a list of error diagnostic JS objects. Does not
+   touch the notification accumulator. *)
+let error_diagnostics error_notifs exn =
   let positions = positions_of_notifications error_notifs in
   match exn with
   | Message.CompilerError content ->
@@ -185,16 +188,6 @@ let handle_error exn =
         val positions = Js.array [||]
       end;
     ]
-
-(* Drain accumulated warnings and return as diagnostic objects *)
-let drain_warnings () =
-  let all = List.rev !notifications_acc in
-  let warnings =
-    List.filter (fun (e : Message.lsp_error) -> e.kind = Message.Warning) all
-  in
-  notifications_acc :=
-    List.filter (fun (e : Message.lsp_error) -> e.kind <> Message.Warning) all;
-  List.map (diagnostic_of_lsp_error "warning") warnings
 
 let () =
   Js.export_all
@@ -234,7 +227,7 @@ let () =
                 detection *)
              let _ = Dcalc.From_scopelang.translate_program prg in
              Message.report_delayed_errors_if_any ();
-             let warning_diags = drain_warnings () in
+             let warning_diags, _error_notifs = drain_all () in
              Message.results ~ppf:Format.str_formatter
                [(fun ppf -> Format.fprintf ppf "Typechecking successful!")];
              let output = Format.flush_str_formatter () in
@@ -244,8 +237,8 @@ let () =
                val diagnostics = Js.array (Array.of_list warning_diags)
              end
            with exn ->
-             let warning_diags = drain_warnings () in
-             let error_diags = handle_error exn in
+             let warning_diags, error_notifs = drain_all () in
+             let error_diags = error_diagnostics error_notifs exn in
              object%js
                val success = Js._false
                val output = Js.string ""
@@ -307,15 +300,15 @@ let () =
                in
                format_results Format.str_formatter scope language results);
              let formatted = Format.flush_str_formatter () in
-             let warning_diags = drain_warnings () in
+             let warning_diags, _error_notifs = drain_all () in
              object%js
                val success = Js._true
                val output = Js.string formatted
                val diagnostics = Js.array (Array.of_list warning_diags)
              end
            with exn ->
-             let warning_diags = drain_warnings () in
-             let error_diags = handle_error exn in
+             let warning_diags, error_notifs = drain_all () in
+             let error_diags = error_diagnostics error_notifs exn in
              object%js
                val success = Js._false
                val output = Js.string ""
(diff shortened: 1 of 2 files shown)


compiler/catala_web_interpreter.ml
```diff
@@ -1,34 +1,283 @@
...
+
 open Catala_utils
 open Driver
 open Js_of_ocaml
 
+(* Register embedded stdlib files in jsoo's /static/ virtual filesystem *)
+let () = Stdlib_embedded.register_stdlib ()
+
+(* Force stdlib_internals modules to initialize by referencing them. Without
+   this, jsoo may optimize away the module initialization. *)
```
denismerigoux: Wouahou this is tricky! I'm afraid we will forget to update this list if we add new externals to the standard library. Worth putting a note somewhere else we remember?

compiler/web/test_web_interpreter.js
denismerigoux, line 1: Can you move all the files related to the web interpreter to a `web` folder inside `compiler`?

rprimet: done

compiler/catala_web_interpreter.ml
rprimet: Good point, there's a `linkall` flag that seems to ensure that externals won't be removed so I added the flag and removed this code.

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+
+open Catala_utils
+open Driver
+open Js_of_ocaml
+
+(* Register embedded stdlib files in jsoo's /static/ virtual filesystem *)
+let () = Stdlib_embedded.register_stdlib ()
+
+(* Format interpretation results as human-readable text *)
+let format_results language results =
```
AltGr: This should be `results_to_string`, the convention is that format functions take a `Format.formatter` argument.

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+  let buf = Buffer.create 256 in
+  let ppf = Format.formatter_of_buffer buf in
+  List.iter
+    (fun ((var, _), result) ->
+      Format.fprintf ppf "@[<hov 2>%s =@ %a@]@\n" var
+        (Shared_ast.Print.UserFacing.value language)
+        result)
+    results;
```
AltGr: Better to use a `vbox` with cuts rather than `"@\n"`
```suggestion
  Format.pp_open_vbox ppf ();
  List.iter
    (fun ((var, _), result) ->
      Format.fprintf ppf "@[<hov 2>%s =@ %a@]@," var
        (Shared_ast.Print.UserFacing.value language)
        result)
    results;
  Format.pp_close_box ppf ();
```

Also, this could be simplified by using `Format.str_formatter` without an explicit buffer.

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+  Format.pp_print_flush ppf ();
+  Buffer.contents buf
+
+(* Format error content as human-readable text *)
+let format_error content =
```
AltGr: Same remarks

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+        if not (String.starts_with ~prefix:Stdlib_embedded.stdlib_path file)
+        then
+          let msg =
+            match pos_message with
+            | Some fmt ->
+              let buf = Buffer.create 64 in
+              let ppf = Format.formatter_of_buffer buf in
+              fmt ppf;
+              Format.pp_print_flush ppf ();
+              Buffer.contents buf
```
AltGr:
```suggestion
              Format.asprintf "%t" fmt
```

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+    (object%js
+       method typecheck (js_options : 'a Js.t) =
+         let language, main_filename, main_contents, module_files =
+           setup_files js_options
+         in
+         let options =
+           Global.enforce_options
+             ~input_src:(Contents (main_contents, main_filename))
+             ~language:(Some language) ~debug:false ~color:Never ~trace:None
+             ~path_rewrite:(fun f -> (f :> Global.file))
```
AltGr:
```suggestion
             ~path_rewrite:(fun f -> (f :> File.t))
```
Global.file is an alias needed for bootstrapping but File.t is the visible type.

compiler/web/catala_web_interpreter.ml
```diff
@@ -0,0 +1,274 @@
...
+             (Js.Unsafe.get js_options (Js.string "trace"))
+             (fun () -> Js._false)
+           |> Js.to_bool
+         in
+         let options =
+           Global.enforce_options
+             ~input_src:(Contents (main_contents, main_filename))
+             ~language:(Some language) ~debug:false ~color:Never
+             ~trace:(if trace then Some (lazy Format.std_formatter) else None)
+             ~path_rewrite:(fun f -> (f :> Global.file))
```
AltGr:
```suggestion
             ~path_rewrite:(fun f -> (f :> File.t))
```

compiler/web/dune
```diff
@@ -0,0 +1,45 @@
...
+   catala_en
+   -e
+   catala_fr
+   -e
+   catala_pl
+   -m
+   plain
+   -o
+   %{target}
+   ../../stdlib)))
```
AltGr: Crunch + manual registration is the reliable old way to make files available at js runtime, but more recently `js_of_ocaml` added built-in support for this.

The rule would become `js_of_ocaml build-fs %{deps} -o stdlib_embedded.js` and the `ml` file would no longer be needed

compiler/web/dune
```diff
@@ -0,0 +1,45 @@
...
+
+(executable
+ (name catala_web_interpreter)
+ (modes byte js)
+ (modules catala_web_interpreter stdlib_files stdlib_embedded)
+ (link_flags
+  (:standard -linkall))
+ (js_of_ocaml
+  (flags
+   (:standard --effects=cps)))
```
AltGr: I am curious why the flag, I don't believe we make any use of effects at the moment ?

catala.opam
```diff
@@ -42,6 +42,9 @@ depends: [
   "uucp" {>= "10"}
   "ubase" {>= "0.05"}
   "zarith" {>= "1.12"}
+  "zarith_stubs_js" {>= "v0.17.0"}
+  "js_of_ocaml" {>= "6.0"}
+  "js_of_ocaml-ppx" {>= "6.0"}
```
AltGr, line 47: Should we make these optional dependencies ? (Open question...)

Makefile
```diff
@@ -208,6 +208,10 @@ CLERK_TEST=$(CLERK_BIN) test --exe $(CATALA_BIN) \
 unit-tests: .FORCE
 	dune build @for-tests @runtest
 
+web-interpreter-tests: .FORCE
+	dune build compiler/catala_web_interpreter.bc.js
+	node compiler/test_web_interpreter.js
```
AltGr: This should probably be part of the `alltest` target, that we use to simulate the CI runs.

compiler/web/test_web_interpreter.js
AltGr, line 1: Thanks for adding these tests. I'm a little worried about their maintenance though; ideally at some point we should be able to run on the scope tests in `tests/` ; although this can only cover "positive" cases, not verify proper error cases.

Another solution would be to turn them into `catala-test-cli`, but that would require a liitle bit of additional support for `clerk runtest` to be able to run `node catala_web_interpreter.js`, feed it with the program as input and retrieve the output.

compiler/web/dune
rprimet: That was due to a build warning: 
```
Warning [missing-effects-backend]: your program contains effect handlers; you
     should probably run js_of_ocaml with option '--effects=cps'
```
I guess this implies that effects are being used by some library?

catala.opam
rprimet: I'll defer to you on this!

Makefile
rprimet: done

compiler/web/test_web_interpreter.js
rprimet: I agree with your remark about the maintenance burden however these tests also exercise behaviors that are specific to the web version (e.g. the web interpreter API, error position reporting, user-supplied multiple files) and were sometimes written after uncovering a bug, so that may be reasonable?

There is now a loop for running (a subset of) scope tests in `tests/` ; note that doing this however required a few changes in `messages.ml / mli` (because of unix / jsoo behavior differences) so a further review would be useful ; the nice thing is that we now surface warnings in the JS API which will help displaying them nicely in the playground editor (previously warnings were disabled and only errors were reported)

PR #1000, opened 2026-04-03
Better support for external types

AltGr:
The encoding for external types wasn't expressive enough, but a safe dose of additional GADTs solved it elegantly.
An external type extends a gadt defined in the runtime so that it can be matched and recovered by type (exactly like `Type.Id` works in recent OCaml stdlib)

This is transparent to the user, who only needs to apply a functor to a module with the few functions required defined. The generated templates pave the way to this nicely.

While this works in the interpreter, the handling of the autotests generated further headaches, but that is being solved. I'll detail later on, but basically the autotests require us to insert computed literals into the generated code, which is hard for abstract types. The ongoing solution handles that through json serialisation, which we wanted anyway.


AltGr:
Ok, this is now done but for the Java backend (@vincent-botbol Help! 😿 )

The following features, which were needed, have been added along the way:
- external types now require to specify a `from_json` function ; in Python, a default is inherited ; in C, the template includes the boilerplate.
- there is an internal `ValueFromJson` operator that takes a (static) string and is compiled into the corresponding intialisation of the object. This is used by the autotest generator to inject the interpreted results, so that they can be checked against the results from the computation within the chosen backend.
- as a bonus, the operator has been made accessible from the surface syntax (the specific syntax should probably be changed though): if external type `Foo` was defined, you can write `#[json = "{ json representation of the object }"] external (Foo)`.
  (I used an attribute because they already had strings, but since this adds new syntax for the external and type, there is nothing mandatory about that: `external ("json string", Foo)`, or even maybe `Foo("json string")` might do the trick)

PR #1006, opened 2026-04-14
Creates module in backend library for each backend

Arnaud-Bihan:
Put backends code for clerk out of clerk_rules

This PR is not the complete feature as there is still a dispatch in Clerk_rules on different backends


AltGr:
Huh I was going to say that you could stack this PR over the other one (by targetting its branch instead of master), but actually I don't think that works across forks 😒

Arnaud-Bihan:
> Huh I was going to say that you could stack this PR over the other one (by targetting its branch instead of master), but actually I don't think that works across forks 😒

Yes unfortunately I didn't find how to make it work, there should be no change needed for catala-lsp-server with this branch but as I can't rebase on it, it will have a red CI until it's merged on master

build_system/backend/backend.ml
```diff
@@ -0,0 +1,63 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+open Clerk_utils
+open Catala_utils
+
+module type S = sig
```
AltGr, line 21: This warrants some documentation: please explain that this is the expected common interface for the backend implementations, and the expectations for each `val`.

build_system/backend/backend.ml
```diff
@@ -0,0 +1,63 @@
+(* This file is part of the Catala build system, a specification language for
+   tax and social benefits computation rules. Copyright (C) 2020-2025 Inria,
+   contributors: Denis Merigoux <denis.merigoux@inria.fr>, Emile Rolley
+   <emile.rolley@tuta.io>, Louis Gesbert <louis.gesbert@inria.fr>
```
AltGr, line 4: Don't hesitate to add yourself ☺

build_system/backend/c.ml
```diff
@@ -0,0 +1,192 @@
...
+        ~test_flags
+        ~include_dirs =
+      let open Common.Flags in
+      let catala_flags_c =
+        Common.Flags.catala_backend_flags ~autotest ~use_default_flags
+          ~test_flags ~accepts_closure_conversion:false
+      in
+      let def = def ~variables in
+      [
+        def Var.catala_flags_c (lazy catala_flags_c);
```
AltGr: Hm, ideally each backend should be able to add its own vars, they're still centralised in `var.ml`. I guess that may not be so easy though ?

build_system/backend/java.ml
```diff
@@ -0,0 +1,183 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+open Clerk_utils
+open Catala_utils
+open Var
+open File
```
AltGr, line 20: Just adding a note here to remember to double-check that all fixes done by vbot in the meantime got correctly included here.

build_system/clerk_utils/ninja.mli
```diff
@@ -17,15 +17,21 @@
...
-    This module is sensitive to the CWD at first use. Therefore it's expected
-    that [chdir] has been run beforehand to the project root. *)
-
-val ocaml_include_flags : string list Lazy.t
-val ocaml_link_flags : string list Lazy.t
-val ocaml_runtime_dir : File.t Lazy.t
-val c_runtime_dir : File.t Lazy.t
-val python_runtime_dir : File.t Lazy.t
-val java_runtime_dir : File.t Lazy.t
+val extern_src :
```
AltGr, line 20: Some documentation would be welcome as well in this interface
(I am sorry that this falls on to you because of this refactoring, while the previous code wasn't very nicely documented... but better benefit from your understanding effort, and undocumented interfaces aren't very helpful)

AltGr (approved):
This is a very nice reorganisation, thanks! I'd be happy with some documentation of the new interfaces (e.g. `backend.mli`), and we should go even further (e.g. split `var.ml`) but that is probably a next step.

build_system/backend/c.ml
Arnaud-Bihan: I've moved each var related to a backend from var.ml to the specified backend, doing that I also moved linking_command and run_artifact out of clerk_driver.
This has been done in those 3 commits:
https://github.com/CatalaLang/catala/pull/1006/changes/bfb63eb22f607140a5b8f9ee91e1fd03a6e2df28
https://github.com/CatalaLang/catala/pull/1006/changes/66cda6b22538448d91abc6743c6c62772c4b37c7
https://github.com/CatalaLang/catala/pull/1006/changes/90f47e9ec63c437b45cf46b0b3e544bcae732191

build_system/backend/java.ml
Arnaud-Bihan: I've checked the MR against vbot changes in https://github.com/CatalaLang/catala/pull/1010/changes, eberything is still in place the only modification that I've observed is for the java backend in the external copy while before we were relaying on the is_stdlib flag pass through the function here we're relaying on the is_stdlib flag coming from the item (each catala item has a flag telling if it's coming from the stdlib).

Those boolean should always be the same (or there is an error somewhere, for more safety if you want I can pass the stdlib flag in parameter along the item and use that instead of item.is_stdlib

build_system/clerk_utils/ninja.mli
Arnaud-Bihan: I've done it in this commit let me know if you think it's okay or not :+1: https://github.com/CatalaLang/catala/pull/1006/changes/a3b4ef17ce58988044e982ef3d8cd6ed271ec377

AltGr:
The code-base for Clerk is already feeling much nicer, thanks !

PR #1001, opened 2026-04-07
Split clerk build system in multiple library

Arnaud-Bihan:
This PR add new library in clerk build system:
- clerk_utils: holds modules uses in everywhere in clerk
- clerk_lib: holds library specific to certain aspect of clerk (cli, toml, config)
- backend: not much thing at first but it's to prepare future backend in it
### If this PR adds a feature or has breaking changes

* [x] Update the Catala language server in case of breaking changes at https://github.com/CatalaLang/catala-language-server
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-language-server/pull/218

PR #1011, opened 2026-04-16
Remove unused SubScopeVarDefinition

rprimet:
After digging a little with @vincent-botbol and @AltGr  we came to the conclusion that `SubScopeVarDefinition` is unused. 

This is a cleanup PR to remove it.

### DIFF
--- a/compiler/dcalc/from_scopelang.ml
+++ b/compiler/dcalc/from_scopelang.ml
@@ -591,19 +591,11 @@ let translate_rule
     'm Ast.expr scope_body_expr Bindlib.box)
     * 'm ctx =
   match rule with
-  | S.ScopeVarDefinition { var; typ; e; _ }
-  | S.SubScopeVarDefinition { var; typ; e; _ } ->
+  | S.ScopeVarDefinition { var; typ; e; io } ->
     let scope_var = Mark.remove var in
     let decl_pos = Expr.no_attrs_pos (Mark.get (ScopeVar.get_info scope_var)) in
     let pos_mark, _ = pos_mark_mk e in
-    let scope_let_kind, io =
-      match rule with
-      | S.ScopeVarDefinition { io; _ } -> ScopeVarDefinition, io
-      | S.SubScopeVarDefinition _ ->
-        ( SubScopeVarDefinition,
-          { io_input = NoInput, decl_pos; io_output = false, decl_pos } )
-      | S.Assertion _ -> assert false
-    in
+    let scope_let_kind = ScopeVarDefinition in
     let a_name = ScopeVar.get_info (Mark.remove var) in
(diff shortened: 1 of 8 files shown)


AltGr (approved):
The best kind of PR

denismerigoux (commented):
Wow wow calm down here :) It may not be used in the codebases so far but to me these `SubScopeVarDefinitions` are an important part of the language! How do you do to define `context` variable for a sub-scope you are calling?

rprimet:
Not sure what failure in CI ( https://github.com/CatalaLang/catala/actions/runs/24504347347 ) is about? Possibly needs to be re-triggered after https://github.com/CatalaLang/catala/pull/1001 is part of the container that builds the LSP?

denismerigoux:
For instance imagine the program:

```catala
declaration scope A:
   context output x content integer

scope A:
    definition x equals 1

declaration scope B:
   output y content integer
   a scope A

scope B:
   definition a.x equals 2
   definition y equals a.x
```

Isn't `definition a.x equals 2` a `SubscopeVarDefinition`?

denismerigoux:
It seems to me that if you merge this PR my example above will fail to compile, forcing you to instead write:

```catala
declaration scope A:
   context output x content integer

scope A:
    definition x equals 1

declaration scope B:
   output y content integer

scope B:
   definition y equals (result of A with { -- x: 2 }).x
```

rprimet:
@denismerigoux after a rebuild I get no issue

```
romain@enclume:~/temp/subscope_test$ cat foo.catala_en 
` ``catala-metadata
declaration scope A:
   context output x content integer

scope A:
    definition x equals 1

declaration scope B:
   output y content integer
   a scope A

scope B:
   definition a.x equals 2
   definition y equals a.x
`  ``

romain@enclume:~/temp/subscope_test$ catala interpret foo.catala_en --scope B
┌─[RESULT]─ B ─
│ y = 2
└─
romain@enclume:~/temp/subscope_test$ 

```

(extra spaces in backtick series because nesting code sections in github is uneasy)

AltGr:
@denismerigoux a while ago; explicit subscopes where transformed into direct scope applications at scopelang time. IIRC, we had concluded that it was pointless to keep two concurrent mechanisms doing basically the same thing. The code removed here is a leftover that we forgot at that time.

Subscopes of course are kept and unchanged in the surface and desugared stages.

denismerigoux:
> explicit subscopes where transformed into direct scope applications at scopelang time. IIRC, we had concluded that it was pointless to keep two concurrent mechanisms doing basically the same thing. The code removed here is a leftover that we forgot at that time.

Ah nice I had forgot about that, it looks like the right decision! I bless this PR then <3

PR #1013, opened 2026-04-20
Improve tests output

AltGr:
The main change is to ensure the code generated keeps the definition order when possible (rather than depending on a hash function ; this is simply changing `Graph.Topological.Make` into `Graph.Topological.Make_stable`)

This makes lots of things much nicer: error messages in the file order, tests reported in file order, etc.

Also quite a few improvements to printing, esp. regarding failed asserts. Maybe look at the commits separately to review since they each have their part of noise in the changed test outputs (in particular the ordering one!)


compiler/shared_ast/print.ml
```diff
@@ -496,8 +513,8 @@ module ExprGen (C : EXPR_PARAM) = struct
     let var = C.var in
     let operator = C.operator in
     let e = C.pre_map e in
-    let paren ~rhs ?(colors = colors) expr fmt e1 =
-      if Precedence.needs_parens ~rhs ~context:e (C.pre_map e1) then (
+    let paren ?(context = e) ~rhs ?(colors = colors) expr fmt e1 =
```
denismerigoux, line 516: If you're curious you may read https://cambium.inria.fr/blog/a-toy-type-language-1/ or look for "unparsing" for more principled ways of putting parenthesis in the pretty-printer :)

compiler/shared_ast/print.ml
```diff
@@ -1077,11 +1113,9 @@ module UserFacing = struct
      Trying would lead to inconsistencies where some comparable numbers are in %
      and some others not, adding confusion. *)
 
-  let date (lang : Global.backend_lang) ppf d =
+  let date (_lang : Global.backend_lang) ppf d =
```
denismerigoux, line 1113: OK, then might as well delete the parameter no?

compiler/shared_ast/print.ml
```diff
@@ -1157,7 +1191,9 @@ module UserFacing = struct
     | EInj { name = _; cons; e = ELit LUnit, _ } ->
       Format.fprintf ppf "@[<hov 2>%a@]" EnumConstructor.format cons
     | EInj { name = _; cons; e } ->
-      Format.fprintf ppf "@[<hov 2>%a@ %a@]" EnumConstructor.format cons
+      Format.fprintf ppf "@[<hov 2>%a %a@ %a@]" EnumConstructor.format cons
+        keyword
+        (match lang with Fr -> "contenu" | _ -> "content")
```
denismerigoux: Don't you want to match the other languages explicitly here instead of `_`?

tests/assertions/bad/inline.catala_en
```diff
@@ -49,8 +47,7 @@ $ catala test-scope A
 └─
 ┌─[ERROR]─ 3/3 ─
 │
-│  Assertion failed: Check that a is more than -6.
-│  The condition resolved to: ~ (6 > -6)
+│  Assertion failed: Check that a is more than -6: ~not~ (6 > -6)
```
denismerigoux: Why `~not~` here and not simply `not` ? Also, why the "The condition resolved to :` disappeared ? I find this particular error message very confusing.

denismerigoux (approved):
Thanks @AltGr, this is very helpful :)

compiler/shared_ast/print.ml
AltGr: Yup this would be better ; but well, this printer is for debugging, but not sure it's worth the effort.

compiler/shared_ast/print.ml
AltGr: Indeed ; it's in case we change our minds on how we want to print the dates, localised vs. international and similar to the source language: here I am changing for the second option, but keeping the arguments gives some easier leeway if we decide to refine on it.

compiler/shared_ast/print.ml
AltGr: I am assuming `english` is the "default case" :)

tests/assertions/bad/inline.catala_en
AltGr: The expressions printed on assert failures have been negated, compared to before.

here it's supposed to be crossed-over `not` (and should be displayed as such if your terminal supports it)
The idea was to be consistent with the other cases, where e.g. `assert 2+2 = 5` would lead to
- previously "Assertion failed: the condition resolved to 4 = 5"
- now just "Assertion failed: 4 ≠ 5" — shorter and clearer IMO

Here, maybe it would be clearer to just *remove* the `not` ?
`assert not (2+2 = 4)` → `Assertion failed: 4 = 4` ?

denismerigoux: Yes I would prefer "no `not`" rather than "crossed-out not".

compiler/shared_ast/print.ml
denismerigoux: Don't you want to put the Polish keyword here since we still support the Polish syntax?

tests/assertions/bad/inline.catala_en
AltGr: Ok, let's not be intuitionistic then ;)

PR #1014, opened 2026-04-21
Fix qualification of constructors after disambiguation

AltGr:
This is the proper fix for b25a3d6.
Also includes a fix for parens in the OCaml printing.
(note: based upon #1013)

### DIFF
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -264,10 +264,6 @@ let rec check_formula (op, pos_op) e =
     check_formula (op1, pos_op1) e2
   | _ -> ()
 
-(** Helper that restores surface positions in module paths. *)
-let restore_position path_item mname =
-  ModuleName.map_info (fun (mn, _) -> mn, Mark.get path_item) mname
-
 let translate_literal l pos =
   match l with
   | S.LNumber ((Int i, _), None) -> LInt (Runtime.integer_of_string i)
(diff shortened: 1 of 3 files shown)


compiler/desugared/from_surface.ml
```diff
@@ -644,7 +640,8 @@ let rec translate_expr
         let uid = Name_resolution.get_scope ctxt id in
         (* Retain the correct positions *)
         ScopeName.map_info
-          (fun (_ml, pos) -> List.map2 restore_position path resolved_path, pos)
+          (fun (path1, pos) ->
+            (if resolved_path = [] then path1 else resolved_path), pos)
```
denismerigoux, line 644: This pattern needs a little comment to explain what is going on here. More generally, knowing where paths are qualified and when they are not would require more global doc, maybe in the `.mld` files?

runtimes/ocaml/catala_runtime.ml
```diff
@@ -533,8 +533,8 @@ module type ExternalTypeSpec = sig
       the left-hand side is respectively smaller, equal or greater than the
       right-hand side *)
 
-  (* val to_expr : t -> string (\** Must output a valid OCaml expression that
-     encodes the given value *\) *)
+  (* val to_expr : t -> string (** Must output a valid OCaml expression that
```
denismerigoux, line 536: You can remove this `val` if it's commented no?

denismerigoux (approved):
OK for me!

compiler/desugared/from_surface.ml
AltGr: They should always be. This line of code was supposed to keep proper positions in the paths of qualified idents (for the LSP, mostly), but had the unfortunate side-effect of destroying the qualification of already resolved, unqualified idents. This was the real cause of the comparison bug we got on monday, took me a while to find out 😅

PR #1015, opened 2026-04-22
Clerk loop with clerk start when there are external module

Arnaud-Bihan:
When writing a catala project with an external module, I came across a loop in the error messages that was unexpected.
image

Using clerk build hints me to use catala gen external that hints me to use clerk start but I can't use clerk start as there is an external module. It's a loop that can be easily fixed by creating the external OCaml file but it's nicer if clerk start would work.

To do so instead of building the whole repo when using clerk start, we only build the stdlib (which is represented by the flag include dir). Also instead of force build the OCaml rules, I've made the code to build the backend specified by the targets. If there are two targets one in c and one in java, clerk start will build the stdlib for java and c.

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -1490,16 +1490,28 @@ let runtest_cmd =
 
 let start_cmd =
   let run config quiet (ninja_flags : string list) =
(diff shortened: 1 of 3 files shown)


AltGr:
Good catch ! Indeed that was the expected behaviour of `clerk start`

PR #1016, opened 2026-04-22
Allow multiple bindings in match cases

AltGr:
i.e. `match foo with pattern -- Constructor content (a, b): a + b`

NOTE: stacked on top of #1014

* [x] Update the Catala language server in case of breaking changes at https://github.com/CatalaLang/catala-language-server *(no changes needed)*

    - [x] already merged (this is backw compatible, and actually was already supported in matches, only not in the short test-match syntax)
    - should be unaffected

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -393,7 +393,10 @@ module Content = struct
       List.iteri
         (fun i (c, bt) ->
           let header = Printf.sprintf "%d/%d" (succ i) len in
-          emit_raw ~ppf ~header c target;
+          (try emit_raw ~ppf ~header c target
+           with err ->
+             Format.fprintf ppf "@[<hov 4>failed to print:@ %a@]@,@."
+               Format.pp_print_text (Printexc.to_string err));
           if Global.options.debug then Printexc.print_raw_backtrace stderr bt)
         contents
 
--- a/compiler/desugared/from_surface.ml
+++ b/compiler/desugared/from_surface.ml
@@ -413,37 +413,18 @@ let rec translate_expr
   in
   match Mark.remove expr with
   | Paren e -> rec_helper (Mark.set (Pos.join pos (Mark.get e)) e)
-  | Binop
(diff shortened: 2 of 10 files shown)


AltGr:
This probably doesn't need an update of the cheat-sheet ; I'll see about the doc, though

compiler/catala_utils/message.ml
```diff
@@ -393,7 +393,10 @@ module Content = struct
       List.iteri
         (fun i (c, bt) ->
           let header = Printf.sprintf "%d/%d" (succ i) len in
-          emit_raw ~ppf ~header c target;
+          (try emit_raw ~ppf ~header c target
+           with err ->
+             Format.fprintf ppf "@[<hov 4>failed to print:@ %a@]@,@."
```
denismerigoux, line 398: I see why this is useful for debugging, but if a user sees that, shouldn't they also have a prompt to file an issue to https://github.com/CatalaLang/catala/issues?

compiler/scopelang/from_desugared.ml
```diff
@@ -206,6 +206,12 @@ let rec translate_expr (ctx : ctx) (e : D.expr) : untyped Ast.expr boxed =
       (List.map (translate_expr ctx) args)
       tys_explicit
       (fun args -> Expr.eapp ~f ~tys m ~args:(add_implicit_args args))
+    (* | EMatch ->
+     *   (\* *\)
+     *   Expr.detuplify_application
+     *     (List.map (translate_expr ctx) args)
+     *     tys_explicit
+     *     (fun args -> Expr.eapp ~f ~tys m ~args:(add_implicit_args args)) *)
```
denismerigoux: Since we do not need to de-tuplify anymore we can clean up this comment?

denismerigoux (approved):
Thanks @AltGr!

compiler/scopelang/from_desugared.ml
AltGr: Indeed, sorry for the noise!

PR #1018, opened 2026-04-27
Fix publiccode.yml validation errors

bfabio:
Fix validation errors and add a GitHub Actions workflow to catch future ones.


denismerigoux (approved):
Thank you very much for this update and extra CI check!

PR #1021, opened 2026-04-30
Refactor operator handling, add Find and Sort

AltGr:

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/61
  * (no update was actually needed)
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-language-server/pull/224
* [x] Update `catala-format`:
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-format/pull/14


compiler/catala_utils/message.ml
```diff
@@ -19,6 +19,33 @@
 
 (**{1 Terminal formatting}*)
 
+type Format.stag += Link of string
+
+let add_link_tags ppf =
```
denismerigoux, line 24: What is this sorcery? You can have clickable links in terminal output 😱😱 ?

compiler/catala_utils/pos.ml
denismerigoux, line 1: I suspect this will break some of @rprimet's tools that parse the terminal output of the terminal, for instance in the playground?

compiler/desugared/from_surface.ml
```diff
@@ -120,16 +120,133 @@ let translate_binop :
...
+  | S.ListMap ->
+    let ty_elt = Type.fresh_var op_pos in
+    let ty_relt = Type.fresh_var (Expr.pos lhs) in
+    op_expr Map [TArrow ([ty_elt], ty_relt); TArray ty_elt]
+  | S.ListFilter ->
+    let ty_elt = Type.fresh_var op_pos in
+    op_expr Filter [TArrow ([ty_elt], (TLit TBool, op_pos)); TArray ty_elt]
+  | S.ListMax | S.ListMin ->
+    (* -> match (reduce (fun x y -> if x > y then x else y) lhs) with | Some x
+       -> x | None -> rhs // impossible *)
```
denismerigoux, line 201: Thanks for the comments, they're very helpful! Though for verification purposes we might want in the future to propagate high-level lists operation further down the compilation line down to Lcalc for instance ?

compiler/shared_ast/typing.ml
```diff
@@ -110,12 +110,11 @@ let unification_error env ~pos ?fmt_pos fmt ty1 ty2 =
       | None -> Some (List.map (fun p -> "", p) pos)
     in
     Message.delayed_error () ~kind:Typing ?extra_pos ?fmt_pos
-      ("Error during typechecking, incompatible types:@\n\
-        @[<v>@{<blue>@<2>%s@} @[<hov>%a@]@,\
-        @{<blue>@<2>%s@} @[<hov>%a@]"
+      ("@[<v 2>Incompatible types:@ @[<h>got@;\
+        <6 0>@]@[<hov>%a@],@ expected @[<hov>%a@]"
```
denismerigoux, line 114: Ok so this is replicating OCaml's typing error message. You know  @AltGr what I'm going to say here <3<3<3

Consider this simple Rust program: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=acbfac11c44308083f7f1d52e8ac5087

```Rust
fn main(){
    let x : u32 = 4;
    let b : bool = false;
    if x {b} else {false}
}
```

The type error message here is: 

```
error[E0308]: mismatched types
 --> src/main.rs:4:8
  |
4 |     if x {b} else {false}
  |        ^ expected `bool`, found `u32`

error[E0308]: mismatched types
 --> src/main.rs:4:11
  |
1 | fn main(){
  |          - expected `()` because of default return type
...
4 |     if x {b} else {false}
  |           ^ expected `()`, found `bool`

error[E0308]: mismatched types
 --> src/main.rs:4:20
  |
1 | fn main(){
  |          - expected `()` because of default return type
...
4 |     if x {b} else {false}
  |                    ^^^^^ expected `()`, found `bool`
```

Can we try to have more info in the message about where do these incompatible types come from?

compiler/surface/ast.mli
```diff
@@ -112,7 +126,7 @@ type builtin_expression =
   | ToDecimal
   | ToMoney
   | Round
-  | External of typ
+  | External of typ  (** Payload is in the attribute [#[json=""]] *)
```
denismerigoux, line 129: The doc about this is missing on https://book.catala-lang.org/en/5-8-2-external-modules.html#external-types

tests/array/good/arg_extremum.catala_en
```diff
@@ -5,7 +5,7 @@ declaration structure S:
   data id content integer
   data income content money
 
-#[test] declaration scope A:
+declaration scope A:
```
denismerigoux, line 8: Why did you remove this?

tests/array/good/sort.catala_en
denismerigoux, line 1: Thank you for the tests!

tests/default/bad/conflict.catala_en
```diff
@@ -11,9 +11,7 @@ scope A:
 
 ```catala-test-cli
 $ catala Interpret -s A --message-format=gnu
-tests/default/bad/conflict.catala_en:8.32-8.36: [WARNING] Multiple conflicting definitions: these have the same conditions and will always trigger a conflict at runtime.
-tests/default/bad/conflict.catala_en:9.32-9.36: [WARNING]
```
denismerigoux, line 15: Very nice!

.ocamlformat
```diff
@@ -2,7 +2,7 @@ profile = default
 margin = 80
 exp-grouping = preserve
 break-fun-decl = fit-or-vertical
-wrap-comments
+wrap-comments = false
```
denismerigoux, line 5: I see you tried to fit this snugly at the end of a monstruous diff :) I would prefer local non-wrapping `[@@` directives to the comments where you don't want to wrap rather than a global setting but I understand these are difficult to place.

denismerigoux (commented):
Thanks @AltGr! Here are already some comments while you finish this beauty

compiler/catala_utils/message.ml
AltGr: Surprise ! 😁 😁 😁 
They work in VScode too

compiler/catala_utils/pos.ml
AltGr: Using Format's tag magic, they are removed if not on a tty (even if color is enabled) ; I enable them with `--color=always` though, because clerk uses that and we want them to get through, but we can easily refine the CLI flags if it's a problem (e.g. `--links=never|always|auto`)

compiler/desugared/from_surface.ml
AltGr: The PR goes quite a bit in that direction, with 3 new operators, but it keeps some of the desugaring, mostly for min/max constructs

compiler/shared_ast/typing.ml
AltGr: The extra positions still detail it so we are actually pretty close:
image

The actual reason for this change was that the previous message was pretty poor when tucked in the LSP's tooltips (we had stg like `Error during typechecking: incompatible types ─➤ boolean ─➤ integer` with extra spacing to make it uglier ; now this reads `Incompatible types:  got boolean, expected integer` which is much nicer)

compiler/surface/ast.mli
AltGr: Noted, I'll remember to add it as I document the new list ops

tests/array/good/arg_extremum.catala_en
AltGr: it just defines a constant used in the next test, it doesn't test anything

.ocamlformat
AltGr: it's unclear how to apply them to comments ; the thing that got me there is the pseudo-code explaining the resulting AST in the desugaring code in `from_surface.ml`. After it got mangled one too many times, which really makes it unreadable (even though I wrap it in `verbatim` blocks with `{v v}`), I got tired. Changing verbatim blocks is not acceptable for a code formatter.

AltGr: (I wouldn't have put it in a dedicated commit if I really wanted to pass it without it being found ;) )

AltGr:
As @denismerigoux noted, this ended up including a few unrelated goodies:
- multi-line markdown titles (694e6a8fe)
- allow trailing semicolons in lists (7b57be264)
- better type error messages (esp. when reformatted for the LSP) (fda1805da)
- fixes for parens in the OCaml backend output (33230ba19)
- fixed the order of the autotests in generated files (e68fb204b)
- added terminal hyperlinks when printing file positions, etc. (cdd144ce2)

Of course, I could split these commits away if this is a problem or it makes the reviews easier.

vincent-botbol:
The `Fix another case of missing parens in the OCaml printer` commit is empty. Let's remove it from the history.

vincent-botbol (dismissed):
Before merging, we need to update the syntax highlighting in the vscode extension. I'll do it next week if you don't.

vincent-botbol:
Awesome work ! Well done

PR #1023, opened 2026-04-30
Draft: A backend for each file instead of having a enabled backends for all

Arnaud-Bihan:
When creating the clerk.ninja file to execute commands, clerk_rules retrieve the enabled backend by concatenation of all possible bakcends for all backends.

But this lead to an issue on external file for example, if I create a target A in OCaml with an external file (written in OCaml) and a target B in Java, Clerk will expect my external file from the target A to have an implementation in Java because the enabled_backends are OCaml and Java for all the files.

This is an attempt not finished to fix that (this is clearly not optimal the way I did it for now)

### DIFF
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -277,8 +277,8 @@ let build_clerk_target
     |> normalize_backends
   in
   let install_targets, all_modules_deps =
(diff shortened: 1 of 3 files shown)


AltGr:
Superseded by #1072 ; it's part of a rewrite, but I took heavy inspiration from this PR on the relevant part.

PR #1024, opened 2026-04-30
Draft: Catala command to create a pandoc markdown

Arnaud-Bihan:
A new catala command to create a pandoc-markdown, from that pandoc markdown you can use pandoc to generate any kind of file pandoc supports. To give an example I've also created a backend to generate a pdf . Markdown html and pdf files are generated but there are still some issues (for example pdf links are not working properly while html and markdown are working fine)


AltGr:
Awesome! Once this is mature enough we can backport the `html` and `latex` commands to just be sugar for a call of `pandoc` over it.

compiler/literate/md.ml
```diff
@@ -0,0 +1,175 @@
...
+      (Format.pp_print_list
+         ~pp_sep:(fun fmt () -> Format.fprintf fmt "\n")
+         (separate_law_and_code language print_only_law complete_headings))
+      children fmt_details_close ()
+  | A.LawInclude _ -> ()
+  | A.ModuleDef _ | A.ModuleUse _ -> () (* TODO: show somehow ?*)
+
+let rec fmt_toc
+    (parents_headings : string list)
+    fmt
```
denismerigoux, line 128: So here's a long-standing Markdown feature we currently do not support: whatv about really long titles that need to be split into multiple lines like:


~~~markdown
## Such a very long title we need 
## two lines to write it
~~~

Right now Catala parses it as two different titles of the same level. I don't think it's in the scope of this PR to support this but this gives you an idea of where we would want to go in terms of functionnalities. The pain point to support this feature for instance is that heading parsing currently goes through the Catala parser, so we would need to tweak the Catala parser just for that small feature of Markdown, while we ignore in the Catala parser most of the other Markdown features (because we delegate them to Pandoc later).

compiler/literate/md.mli
```diff
@@ -0,0 +1,41 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2020 Inria, contributor:
+   Denis Merigoux <denis.merigoux@inria.fr>
```
denismerigoux, line 3: You can change the name and date here for something more accurate like : 


    Copyright (C) 2026 Inria, contributor: Arnaud Bihan <XXXXX>

compiler/driver.ml
```diff
@@ -480,6 +508,33 @@ module Commands = struct
         $ Cli.Flags.wrap_weaved_output
         $ Cli.Flags.extra_files)
 
+  let pdf options output print_only_law wrap_weaved_output =
```
denismerigoux, line 559: This is very nice but the PDF generated by Pandoc is very ugly no? Right now our pipeline of choice to generate PDF is through LaTeX with the preamble set out in 

https://github.com/CatalaLang/catala/blob/master/compiler/literate/latex.ml#L52-L165

I would prefer not to roll out an out-of-the-box PDF feature that provides an uglier PDF than the one we're already providing with our current preferred pipeline.

syntax_highlighting/en/pandoc/catala_en.xml
denismerigoux, line 1: Amazing, did not know `pandoc` supported that,!!

denismerigoux (commented):
Thank you!! I guess now we' could lift these commands to `clerk` so that we can produce a `.md` or `.tex` or `.html` file per target ;)

AltGr:
While your added syntax highlighting is pretty neat, I am a bit wary of having yet another syntax coloring spec in yet another format (I am losing count of how many we have :/)

An option could be to try and leverage the fancy highlighting based on the tree-sitter syntax that we already use for the [catala-book](https://github.com/CatalaLang/catala-book/tree/main/treesitter). One possible way to do this would be to preprocess all catala code blocks with something like `tree-sitter highlight --html --scope catala-code-en` and insert the HTML blocks directly in the intermediate markdown. I don't think there exists yet a way to insert this highlighting directly in pandoc (unless of course we write [our own filter](https://github.com/smimram/ocaml-pandoc/blob/master/examples/include.ml))
- upsides:
  * better coloring
  * less maintenance burden for the rules
- downsides:
  * we depend on the tree-sitter binary in addition to pandoc. it's nothing in size compared to texlive, but still one thing to worry about in our installation (note that catala-format already has this dependency)
  * the intermediate markdown is less readable (but that may not be a problem, if you want readable, you have the raw sources on one side and the html/pdf on the other)

compiler/driver.ml
Zylon04: Yes you're right as of right now the feature is far from being completed. That command was just to give an example of what was possible and also to test the pdf result with a single command instead of two (generating the pandoc markdown and then call pandoc targetting pdf)

Arnaud-Bihan: Mmmh after checking I may have introduced something wrong in the last version because indeed the pdf is not working as expected, it was working way better before now there are a lot of `{...}` that should not appear

PR #1028, opened 2026-05-11
Stdlib/dates: add primitives for explicit rounding mode

AltGr:
This is required when using date operations from top-level functions at the moment, as those don't have a way to select the rounding mode.

The function names have been chosen to match the existing `date round up·down` / `date arrondi inf·supérieur` scope-level constructs.

This is needed at the moment, but hopefully at some point we can propagate rounding options in a better way like in scopes. Nevertheless, it's sufficient to close #986.

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -108,8 +108,8 @@ let formatter_of_out_channel ?(nocolor = false) oc =
   let ppf =
     lazy
       (if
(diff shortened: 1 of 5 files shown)


denismerigoux (approved):
Thanks very much! Could you also backport the signature and doc of the new functions to https://book.catala-lang.org/en/5-7-standard-library.html#module-date?

PR #1033, opened 2026-05-21
Add a DebugPrint op, print relevant vars on failed assert

AltGr:
This makes assertion failures in the backends much nicer and easier to debug, by giving the expected insight into what is wrong and how.

Basically, for asserts of the form `assert var = 42`, the assertion failure message giving the source code position will be preceded by `var = 43` on the previous stderr line, and this for all backends. In addition, this scales for all forms of assertions.

**Note:** while the new Op.DebugPrint could be used to translate the #[debug.print] attrs, this is **not** done at the moment.

### DIFF
--- a/compiler/dcalc/autotest.ml
+++ b/compiler/dcalc/autotest.ml
@@ -85,8 +85,8 @@ let test_scope_outs ctx lang acc env name scope =
     | exception e ->
       if e <> Exit then
         Message.warning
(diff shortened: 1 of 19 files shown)


vincent-botbol:
Tested, works as intended. Looks good to merge.

vincent-botbol (approved):

vincent-botbol (approved):

PR #1041, opened 2026-05-29
Turn the 'No plugin registered' warning into debug

AltGr:
This is too verbose, and it's legit to have attributes intended for a third-party plugin but ignored by Catala


denismerigoux (approved):

vincent-botbol (approved):
Awesome.

PR #1045, opened 2026-06-01
Fix, complete and unify the JSON outputs with the different backends

AltGr:
At the moment, this just implement JSON output in the Python backend and fixes some small things in the OCaml one

~~NOTE: stacked on top of #1031~~


denismerigoux (approved):

AltGr:
Ah, that's annoying: CI fails in closure-conversion mode because of:
`{"r":{"f":"<function>"}}`  vs.  `{"r":{"f":["<function>",[]]}}`

Testing the interpreter / OCaml backend in closure conversion mode is hackish anyway¹... a nice validation while we could get it passing, but it might be time to disable it from CI if it's becoming more of a burden.
- the closure conversion is well tested already by the C backend
- if we remove the `-O` from this test it's already pretty broken anyway.

¹ the interpreter dynamically detects if it has a closure instead of a function for application, and some other tweaks; but the runtime operators are still not designed to take closures.

denismerigoux (approved):

PR #1050, opened 2026-06-03
Improve clerk handling of the exception-tree building

AltGr:
This relaxes the need to run ninja by fine-tuning the args needed by Catala.

NOTE: I removed the `make_absolute` from the JSON printing -- the relative path is now correct -- but the result in VScode should be checked for surprises

Also: adds a missing mli and hyperlinks (you can now click the conditions in the trace in terminal output)

### DIFF
--- a/build_system/backend/common.ml
+++ b/build_system/backend/common.ml
@@ -76,7 +76,13 @@ module Flags = struct
       if inplace then
         {
           options with
-          global = { options.global with build_dir = Filename.current_dir_name };
+          global =
+            {
+              options.global with
+              build_dir =
+                File.make_relative_to ~dir:File.original_cwd
+                  Filename.current_dir_name;
+            };
         }
       else options
     in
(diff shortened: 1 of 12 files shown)


vincent-botbol (changes requested):
In this current state, it doesn't work properly in the extension. The path is relative to the project's root which is not necessarily our cwd when we are in VScode, hence the links are broken. Perhaps, the extension could do some work on its side but, for resilience sake, I think it's best to leave this path as absolute.

AltGr:
I am still not convinced by absolute paths (although it certainly is the fastest reliable fix in this case) ; but you actually point to an issue outside of this consideration: when running `clerk`, the paths are generally expected to be relative to the CWD, not the project root. For example, you can run `clerk test .` from a subdirectory of your project, and the file names reported will be relative to that subdirectory.

Do you think that fixing this would solve your issue in any way, or is it unrelated ?

AltGr:
I fixed that, and according to my tests that fixed it in the extension as well, @vincent-botbol what do you think ?

vincent-botbol (approved):

PR #1051, opened 2026-06-03
Detect duplicate declaration of enum cases and struct fields

### LINKED ISSUE #1049: `string_enum: duplicate case` error while encoding from and to JSON
An internal error `Unexpected error: Invalid_argument("Json_encoding.string_enum: duplicate case")` may be triggered when trying to outputing or input JSON encoded Catala structures.

This seems to be triggered by the fact that there are duplicate enumeration constructors in enum declarations.
The compiler should either trigger an error at compile-time or at least a warning and remove duplicate from its internal structures.

AltGr:
Fixes #1049


denismerigoux (approved):

AltGr:
CI red because of a race condition with changes in formatting in another PR and the merge of the corresponding changes in repo catala-examples ; we can assume it's passing.

PR #1053, opened 2026-06-04
Debug: report CPU time rather than absolute time

AltGr:
This should alleviate issues with debug output changing on loaded machines.


denismerigoux (approved):

AltGr:
CI red because of a race condition with changes in formatting in another PR and the merge of the corresponding changes in repo catala-examples ; we can assume it's passing.

PR #1059, opened 2026-06-10
Optimisation pass: limit width of sub-expression exploration

AltGr:
This is a quick-fix, but should avoid some explosions for now.

### DIFF
--- a/compiler/shared_ast/optimizations.ml
+++ b/compiler/shared_ast/optimizations.ml
@@ -28,17 +28,25 @@ let binder_vars_used_at_most_once
   (not (Array.exists Fun.id (Bindlib.mbinder_occurs binder)))
   ||
   let vars, body = Bindlib.unmbind binder in
-  let rec vars_count (e : (('a, 'b) dcalc_lcalc, 'm) gexpr) : int array =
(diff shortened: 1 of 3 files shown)


compiler/shared_ast/optimizations.ml
```diff
@@ -28,17 +28,25 @@ let binder_vars_used_at_most_once
   (not (Array.exists Fun.id (Bindlib.mbinder_occurs binder)))
   ||
   let vars, body = Bindlib.unmbind binder in
-  let rec vars_count (e : (('a, 'b) dcalc_lcalc, 'm) gexpr) : int array =
+  let credit = ref 14 in
+  (* Raises [Exit] on either var used twice, or credit exhausted *)
```
denismerigoux, line 32: A high-level comment linking the credit to the size of sub-expressions would be perfect :)

denismerigoux (approved):

compiler/shared_ast/optimizations.ml
AltGr: I am rewriting this pass in a more algorithmically-decent way anyway ;)

PR #1061, opened 2026-06-12
Test usage of modules from external modules

### LINKED ISSUE #1056: External modules should allow usages from other Catala modules
This should be allowed : 

~~~markdown
> Module Foo external 

> Usage of Bar 

```catala-metadata
declaration baz contenu money depends on arg content Bar.Enum
```
~~~

AltGr:
Both usage of a Catala module and an external module are tested

Closes #1056, which doesn't appear to require any additional work ?


denismerigoux (approved):
Amazing news !! We should tell @guillaumeFerard, the limitation we had when we tested two weeks ago was probably a 1.1.0 limitation that was fixed in 1.2.0.

AltGr:
Ah, according to CI that didn't work as well as in my tests. Probably just something to fix in Clerk, which may be missing a dependency edge, then.

AltGr:
Haha, the issue was actually neither with Catala nor Clerk, but just with the test that was obsolete and didn't use the newer gen-external template that instanciates the comparison methods.

AltGr:
Ah, of course. Should have thought of it earlier, but whole-program interpretation becomes impossible in this case.

The proposed solution makes sure we dynlink the required Catala modules **even in whole-program mode**. This will be a duplicate implementation, but the two should be able to coexist since they remain in parallel realms, and we already support translation of the data structures between native & AST.

AltGr:
CI only fails on the LSP with an apparently unrelated error (PR race ?)

vincent-botbol (approved):

PR #1067, opened 2026-06-19
Error out on use of private types

AltGr:
Name resolution didn't check it, so before this patch, it would give ugly error messages in the backends.

This also includes nice hints detecting that the type exists but isn't exposed, and suggesting to use catala-metadata blocks.

Also includes some more cosmetic changes, see the patches.

### DIFF
--- a/compiler/catala_utils/message.ml
+++ b/compiler/catala_utils/message.ml
@@ -661,7 +661,7 @@ let register_content_as_delayed_error
     | None -> true
   in
   if register_error then (
-    let bt = Printexc.get_raw_backtrace () in
+    let bt = Printexc.get_callstack 12 in
     if Global.options.stop_on_error then
       Printexc.raise_with_backtrace (CompilerError m) bt;
     global_errors.rev_delayed_errors <-
--- a/build_system/clerk_report.ml
+++ b/build_system/clerk_report.ml
@@ -268,8 +268,10 @@ let print_diff ppf p1 p2 =
       (* Account for "\r\n" *)
         else pos.Lexing.pos_cnum
     in
-    seek_in ic (pos_char pstart);
-    really_input_string ic (pend.Lexing.pos_cnum - pstart.Lexing.pos_cnum)
+    try
+      seek_in ic (max 0 (pos_char pstart));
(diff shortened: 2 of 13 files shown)


AltGr:
Don't forget to merge https://github.com/CatalaLang/catala-examples/pull/46
~~and https://github.com/CatalaLang/catala-language-server/pull/240~~

compiler/catala_utils/message.ml
```diff
@@ -448,25 +448,19 @@ module Content = struct
...
-      emit_raw ?ppf content target;
-      if Global.options.debug then Printexc.print_raw_backtrace stderr bt
+  let emit_n ?ppf (errs : t list) (target : level) =
+    match errs with
+    | [content] -> emit_raw ?ppf content target
     | contents ->
       let ppf = match ppf with Some ppf -> ppf | None -> get_ppf target in
       let len = List.length contents in
       List.iteri
-        (fun i (c, bt) ->
```
denismerigoux, line 463: We don't display the backtrace any more in errors with setting `--debug`?

denismerigoux (approved):
Thanks! This would benefit adding lines into https://github.com/CatalaLang/catala/blob/nopub/CHANGELOG.md documenting the cosmetic changes for the end-user :)

compiler/catala_utils/message.ml
AltGr: In the case where we handle multiple errors. I'd like to discuss this with @vincent-botbol  actually: there are three cases for errors:
- `Message.error` caught at the toplevel: in this case we display the backtrace just like before
- `Message.delayed_error`: we don't have a backtrace available since nothing was raised, so I disabled printing (it was giving garbage). We could probably get a backtrace by using an effect here... 😇 
- `Message.error`, but within a `Message.wrap_to_delayed_error` block: in this case, there was a `try`, although in a limited scope. So, we can actually retrieve the backtrace from the `try` block down: this used to be printed but I have disabled it altogether by focusing on the previous case. We could restore it and make the backtrace argument an option, but I don't know how prevalent it is and would like @vbot's opinion.

AltGr: Nevermind, `Printexc` has had a `get_callstack` function for a while but I had forgotten about it.

PR #1070, opened 2026-06-25
Clerk: strip trailing CR from scope-test output (fix 0 tests on native Windows)

rprimet:
On native Windows, `clerk test` reports **0/0/0** ("ALL TESTS PASSED", zero counted) for any project whose tests are `#[test]` *scopes* — even though the tests run fine. 

**Cause.** `clerk runtest` parses `catala interpret --quiet`'s `"<scope>: passed"` lines (`compiler/driver.ml`, commented "parsed by Clerk"). On a native Windows build catala's stdout is in text mode, so those lines arrive as `…passed\r\n`. `run_catala_test_scopes` reads them with `In_channel.input_line`, which strips the `\n` but keeps the `\r`, and the result is matched under `whole_string` — so `passed\r` never matches and every scope result is dropped. CLI (`catala-test-cli`) tests were unaffected because they already go through `sanitize` (which strips `\r\n`); the scope-test path did not.

**Fix.** Hoist the existing `re_endline` out of `sanitize` to module scope and strip trailing line endings from each scope-test output line in `run_catala_test_scopes`.

### DIFF
--- a/build_system/clerk_runtest.ml
+++ b/build_system/clerk_runtest.ml
@@ -46,7 +46,7 @@ let out_line output_buf str =
 (* Strips trailing line endings. Notably, on Windows catala's stdout is in text
    mode, so lines we parse below arrive with "\r\n"; [input_line] only drops the
    "\n", leaving a '\r' that would defeat the [whole_string] matches. *)
-let re_endline = Re.(compile @@ seq [rep (set "\r\n")])
+let re_endline = Re.(compile @@ rep (set "\r\n"))
 
 let sanitize =
   let re_endtest = Re.(compile @@ seq [bol; str "```"]) in
--- a/build_system/clerk_runtest.ml
+++ b/build_system/clerk_runtest.ml
@@ -43,6 +43,11 @@ let out_line output_buf str =
       pos_bol = pos_cnum;
     }
 
+(* Strips trailing line endings. Notably, on Windows catala's stdout is in text
+   mode, so lines we parse below arrive with "\r\n"; [input_line] only drops the
+   "\n", leaving a '\r' that would defeat the [whole_string] matches. *)
+let re_endline = Re.(compile @@ seq [rep (set "\r\n")])
+
 let sanitize =
   let re_endtest = Re.(compile @@ seq [bol; str "```"]) in
   let re_modhash =
@@ -59,7 +64,6 @@ let sanitize =
              char '"';
            ])
   in
-  let re_endline = Re.(compile @@ seq [rep (set "\r\n")]) in
   let re_backslash = Re.(compile (repn (char '\\') 1 (Some 2))) in
   fun str ->
     str
@@ -208,7 +212,10 @@ let run_catala_test_scopes
   in
   Unix.close cmd_out_wr;
   let out_lines =
-    Seq.of_dispenser (fun () -> In_channel.input_line command_ic)
+    Seq.of_dispenser (fun () ->
+        Option.map
+          (Re.replace_string re_endline ~by:"")
+          (In_channel.input_line command_ic))
   in
   let parse_error line =
     let re_error =


vincent-botbol (changes requested):
Looking at the doc, it says:
```
val input_line : t -> string option
(** [input_line ic] reads characters from [ic] until a newline or the end of
    file is reached.  Returns the string of all characters read, without the
    newline (if any).  Returns [None] if the end of the file has been reached.
    In particular, this will be the case if the last line of input is empty.

    A newline is the character [\n] unless the file is open in text mode and
    {!Sys.win32} is [true] in which case it is the sequence of characters
    [\r\n]. *)
```
I think the issue could be solved by invoking `Input_channel.set_binary_mode command_ic false`. That would be more efficient.

Actually, the channel is created using `Unix.in_channel_of_descr` which also states that :
```
On Windows: {!Stdlib.set_binary_mode_out} always fails on channels created
with this function. 
```
So perhaps it's not the best idea..

vincent-botbol:
~~Ok forget, about the fix. I don't understand why this fix is needed. Look at the code, we have:~~
```ocaml
  let out_lines =
    Seq.of_dispenser (fun () ->
        Option.map String.trim_end (In_channel.input_line command_ic))
  in
  let success, expected =
    Seq.fold_left
      (fun (success, expected) result_line ->
        (*                vvvvvvvvvvvvvvvvvvvvv *)
        let result_line = sanitize result_line in
        out_line result_line;
        match Seq.uncons expected with
        | Some (l, expected) -> success && String.equal result_line l, expected
        | None -> false, Seq.empty)
      (true, expected) out_lines
  in
  ...
  ```
~~This call should already remove the trailing `\r`. Perhaps, I'm missing something. Can you confirm that this patch indeed fixes the issue?~~

My bad, I was looking at the wrong function (`run_catala_test` instead of `run_catala_test_scopes`).

File: build_system/clerk_runtest.ml [hunk 1]
```diff
@@ -59,7 +64,6 @@ let sanitize =
              char '"';
            ])
   in
-  let re_endline = Re.(compile @@ seq [rep (set "\r\n")]) in
```
vincent-botbol, line 62: I don't understand why this was also removed from `sanitize`. Doesn't `run_catala_test` need it ?

vincent-botbol (changes requested):

build_system/clerk_runtest.ml
rprimet: Yes but it's just hoisted up I think? `run_catala_test` still references it

vincent-botbol: Oh right, my bad...

vincent-botbol (approved):

build_system/clerk_runtest.ml
```diff
@@ -43,6 +43,11 @@ let out_line output_buf str =
       pos_bol = pos_cnum;
     }
 
+(* Strips trailing line endings. Notably, on Windows catala's stdout is in text
+   mode, so lines we parse below arrive with "\r\n"; [input_line] only drops the
+   "\n", leaving a '\r' that would defeat the [whole_string] matches. *)
+let re_endline = Re.(compile @@ seq [rep (set "\r\n")])
```
AltGr:
```suggestion
let re_endline = Re.(compile @@ rep (set "\r\n"))
```
`seq` with a singleton list is useless (not from your patch I know 😅, I just spotted it now)

AltGr (commented):
Thanks for spotting this bug! Actually, using the same `Seq.map String.trim_end` as in the other function should have worked here as well, the function takes care of `\r`.

build_system/clerk_runtest.ml
rprimet: Oh yes, good point thanks!

PR #1072, opened 2026-07-06
Clerk refactoring, targets classification and dependency handling

AltGr:
## Checklist

* [x] Update the CHANGELOG.md

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-book/pull/75
      [PREVIEW](https://catala-book-preview-pr-75.catala-book-preview.workers.dev/)
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-examples/pull/56
  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-language-server/pull/246


.woodpecker/main.yaml
```diff
@@ -44,8 +44,21 @@ steps:
     depends_on: build
     environment: *build_env
     commands:
+      # This reproduces 'make testsuite', but splitting in individual commands
```
denismerigoux, line 47: Then we should delete the alias `testsuite` in the Makefile, otherwise people will add things to it thinking they will be tested by the CI but actually no.

build_system/backend/c.ml
```diff
@@ -19,6 +19,8 @@ open Clerk_utils
 open Catala_utils
 open Clerk_lib
 
+type Clerk_config.backend += T
```
denismerigoux: What's `T`? You meant `C`?

build_system/clerk_lib/clerk_cli.ml
```diff
@@ -305,15 +314,15 @@ let code_coverage =
     value
     & flag
     & info ["code-coverage"]
-        ~env:(Cmd.Env.info "CATALA_MEASURE_COVERAGE")
+        ~env:(Cmd.Env.info "CLERK_CODE_COVERAGE")
```
denismerigoux, line 314: These are user-facing changes and should be documented in the `CHANGELOG.md`

build_system/clerk_driver.ml
```diff
@@ -161,42 +152,41 @@ let target_backend config t =
         with
         | Some rule -> rule.Config.backend
         | None ->
-          Message.error
-            "Unhandled extension @{<red;bold>%s@} for target @{<red>%S@}" ext t)
+          if List.mem_assoc ext (extensions_backend ()) then
+            Message.error
+              "Extension @{<red;bold>%s@} of target@ @{<red>%S@} is not \
+               supported by the enabled backends"
```
denismerigoux: Which ones? We could throw in some more love for our users here :)

build_system/clerk_driver.ml
```diff
@@ -161,42 +152,41 @@ let target_backend config t =
...
   | ext -> aux ext
 
+(*
 let config_backend = function
-  | Clerk_config.OCaml -> `OCaml
-  | Clerk_config.C -> `C
-  | Clerk_config.Python -> `Python
-  | Clerk_config.Java -> `Java
+  | Clerk_backends.Ocaml.T -> `OCaml
+  | Clerk_backends.C.T -> `C
```
denismerigoux: Ah OK, I understood, `T` is for type.

build_system/manpages.sexp
```diff
@@ -1,6 +1,5 @@
 (rule (alias man) (action (with-stdout-to clerk.1 (run %{bin:clerk}  --help=groff))))
 (rule (alias man) (action (with-stdout-to clerk-build.1 (run %{bin:clerk} build --help=groff))))
-(rule (alias man) (action (with-stdout-to clerk-ci.1 (run %{bin:clerk} ci --help=groff))))
```
denismerigoux, line 3: Should not be deleted here

compiler/scalc/to_python.ml
```diff
@@ -218,7 +218,7 @@ let rec format_typ ctx (fmt : Format.formatter) (typ : typ) : unit =
          ~pp_sep:(fun fmt () -> Format.fprintf fmt ",@ ")
          format_typ_with_parens)
       t1 format_typ_with_parens t2
-  | TArray t1 -> Format.fprintf fmt "List[%a]" format_typ_with_parens t1
+  | TArray t1 -> Format.fprintf fmt "Array[%a]" format_typ_with_parens t1
```
denismerigoux, line 221: Is this an answer to @edelagnier's comments about the Python backend?

denismerigoux (commented):
A first batch of comments on this draft :)

.woodpecker/main.yaml
AltGr: Added a comment there to avoid that 😅 . It still needs to be duplicated so that we can have something that is almost the same as the CI with a simple `make` command.

AltGr:
Ok, so the bulk of the work is done here, tests should pass and it is now in sync with master + the changes in the [clerk-windows-fixes](https://github.com/CatalaLang/catala/pull/1073) branch (boy, was that tedious)

There are still quite a few things to handle, though, as there are no integration tests for clerk:
- [x] Update the changelog
- [ ] Update the doc for the newer changes in clerk.toml
- [x] `clerk build` and `clerk test` should be getting reasonaby stable, but `clerk run` needs more care and testing, especially with json input
- [x] reimplement `clerk ci` (by combining the ninja targets + finalisers of `clerk test` with all backends enabled and `clerk build`)
- Fix some regressions on information and error messages:
  - [x] Restore the information message on the results of building targets
  - [x] Restore the proper error when the implementation is missing for an external module
- [x] Fix `clerk build <dir>` trying to build modules with unsupported backends
- [x] ~~Check the release of ninja_utils 2.0~~ Ninja_utils 2.0 is [already merged](https://github.com/ocaml/opam-repository/pull/30428), so remove the pin-depends from the opam file (and possibly re-generate the CI image). This is more related to #1073, but since we may merge both at once (see at the bottom of this post)
- [x] Test and fix remaining bugs
- [ ] Add the backend test results to the XML and JSON test reports (they only are in the console report at the moment)

Further work (not related to regressions, so could be postponed -- it may be best to merge this asap, I don't want to rebase again):
- [ ] `clerk typecheck` should be refactored to be read-only (like `clerk exceptions`) and work on targets
- [ ] Implement the `install-sources` clerk config (default yes?) to copy the relevant Catala sources into each target (preferably as a zip file. Rationale: when you get an error message, you should have access to the file it points to). Conversely, remove `install-objects` that should now only be triggered by the new `--obj` cli flag.
- [ ] Generate C dependency Makefiles and Java project files, like is now done with dune for OCaml
- [ ] Add specific integration tests for clerk [using dune CRAM](https://dune.readthedocs.io/en/stable/reference/cram.html), possibly with hard-coded project skeletons.
- [ ] Cleanup the situation with `include-directories` which do not serve a meaningful purpose at the moment. The plan is to scan everything recursively instead, the only downside being that module name conflicts are no longer tolerated. `include-directories` should be re-purposed into a list of directories to recursively scan (defaulting to `.`), and maybe deprecated in favor of an `exclude-directories` field. This will allow for quite some simplification, like the possibility to pass the same `-I` flags to the backend compilers whatever the module being built, or removing the check for "sibling" modules while building the rules.

Note: it's not necessary to manually merge #1073 first, merging this will automatically mark both PRs as merged since this includes #1073's head commit.

vincent-botbol (approved):

AltGr:
Small issues found remaining:
- `clerk build file -b interpret` always returns `does not support any of the selected backends and will be ignored.` (instead of compiling the @catala/obj target ?)
- `clerk test tests/modules/bad/mod_def1.catala_en` triggers a ninja error

vincent-botbol:
Fixed up a patch in head commit to restore stderr coloring in typecheck command.

PR #1073, opened 2026-07-08
Clerk windows fixes

rprimet:
An assortment of fixes to bring up a native windows toolchain.

Mostly path-related plus a few windows-specific limitations (cmdline length on windows...).


build_system/backend/java.ml
```diff
@@ -24,6 +24,36 @@ let javac = Var.make "JAVAC"
 let javac_flags = Var.make "JAVAC_FLAGS"
 let jar = Var.make "jar"
 let java = Var.make "JAVA"
+let backend_name = "java"
+
+(* Java's classpath entry separator is OS-dependent: ';' on Windows (':' would
+   collide with the drive-letter colon in 'C:\...'), ':' elsewhere. Parameterized
+   by [win32] so it is unit-testable off-Windows. *)
```
AltGr:
```suggestion
(* Classpath entry separator. Parameterized by [win32] so it is unit-testable off-Windows. *)
```

The fist sentence is a tautological comment, says exactly the same thing as the code in more verbose and less clear way.

build_system/backend/ocaml.ml
```diff
@@ -84,7 +84,17 @@ let linking_command ~build_dir ~var_bindings link_deps item target =
...
+     Var.quote_arg at the include sites). That quoting is required because the
+     variable is spliced into the ninja *rule* commands, which ninja runs through
+     a shell: an unquoted path with spaces (e.g. C:\Program Files\Catala\...\
+     zarith) would be word-split into two arguments. This link step is the one
+     consumer that is NOT a ninja rule — it is a direct argv exec
+     (Clerk_cli.run_command_line -> Unix.create_process, no shell), where each
+     list element is already exactly one argument and the surrounding quotes
+     would be passed literally as part of the path (ocamlopt: "Cannot find file
+     zarith.cmxa"). So strip the shell-quoting back off for this path only. *)
+  @ List.map Var.unquote (Var.get_var var_bindings ocaml_include)
```
AltGr: This quoting is a bit tedious, and may prove fragile in the long run, In the OCaml side of things, we have an argument list so there is no ambiguity at this point, and no quoting is needed.

I believe it should be the responsibility of `Nj.rule.format` to properly quote, when the output is done.

build_system/backend/python.ml
```diff
@@ -22,6 +22,12 @@ open Clerk_lib
 let catala_flags_python = Var.make "CATALA_FLAGS_PYTHON"
 let python = Var.make "PYTHON"
 
+(* PYTHONPATH entries are separated by ';' on Windows (':' would be read as part
+   of a 'C:\...' drive letter and mis-split the entry), ':' elsewhere.
+   Parameterized by [win32] so it is unit-testable off-Windows. *)
+let path_sep ~win32 = if win32 then ";" else ":"
```
AltGr: That's true for all PATH variables actually, probably worth putting this in `Catala_utils/file.ml` (not sure about the unit testing bit in that case though)

build_system/clerk_utils/var.ml
```diff
@@ -58,6 +58,29 @@ let cat_files = make "cat_files" (* Useful on Windows only *)
...
+   word-split. Ninja expands ${VAR} inside the quotes, and double quotes work
+   for both cmd.exe and /bin/sh. Use ONLY in command contexts — never for ninja
+   paths in inputs/outputs, which need ninja's own escaping instead. *)
+let quoted x = "\"" ^ !x ^ "\""
+
+(* Same as [quoted] but for a literal string (e.g. an absolute include dir)
+   rather than a variable reference. Ninja's own escaping ($ , $:) makes ninja
+   PARSE the file, but is un-escaped before the command runs, so a spaced path
+   still needs shell quotes to survive word-splitting by /bin/sh or CommandLineToArgvW. *)
+let quote_arg s = "\"" ^ s ^ "\""
```
AltGr: Also see `String.quote`, which provides JSON-compatible quoting: I am not sure if that's fully compatible with what we want here (which may depend on the shell / the OS ?)
But that function handles arguments containing quotes or newlines, which is not the case here.

Anyway. as mentioned above, this should probably be in `ninja_utils`

compiler/catala_utils/file.ml
```diff
@@ -30,76 +30,189 @@ let finally f k =
...
 
-let path_to_list path =
-  let p = String.re_split_delim dir_sep_re path in
-  let drive, p =
-    if not Sys.win32 then None, p
+(* Platform-parameterized path logic: the [~win32] flag and an explicit [~cwd]
+   keep these pure and OS-independent, so they can be unit-tested on Linux. The
+   public path functions below are thin wrappers passing [Sys.win32] and
+   [Sys.getcwd ()]. *)
+module Path = struct
```
AltGr: Splitting into a separate module may not be a bad idea indeed. But maybe it should be besides, and not within `File` ?

Also, for consistency,  please rename the functions with `path` in their names : `Path.path_to_list` → `Path.to_list`, `Path.list_to_path` → `Path.from_list`, etc.

I an also not convinced by the `win32` parameter, which adds a lot of noise.  Are the unit tests really that important ? A less invasive solution could be to define a global reference `let win32 = ref Sys.win32`, and overwrite it only in the testing scope.

build_system/clerk_utils/var.ml
AltGr: Quoting the Ninja manual: this may be interesting

> On Unixes, commands are arrays of arguments. The Ninja command variable is passed 
> directly to sh -c, which is then responsible for interpreting that string into an argv array. 
> Therefore, the quoting rules are those of the shell, and you can use all the normal shell 
> operators, like && to chain multiple commands, or VAR=value cmd to set environment 
> variables.
> 
> On Windows, commands are strings, so Ninja passes the command string directly to 
> CreateProcess. (In the common case of simply executing a compiler this means there is 
> less overhead.) Consequently, the quoting rules are determined by the called program, 
> which on Windows are usually provided by the C library. If you need shell interpretation 
> of the command (such as the use of && to chain multiple commands), make the 
> command execute the Windows shell by prefixing the command with cmd /c. Ninja 
> may error with "invalid parameter" which usually indicates that the command line 
> length has been exceeded.
```

File: compiler/scalc/to_c.ml [hunk 6]
```diff
@@ -482,6 +482,13 @@ let format_op (fmt : Format.formatter) (op : operator Mark.pos) : unit =
   | Add_dat_dur _ -> assert false (* needs specific printing *)
   | op -> Format.fprintf fmt "@{<blue;bold>%s@}" (Operator.name op)
 
+(* Emit a source position as a C initialiser struct. [%S] escapes the filename
+   so a Windows path's backslashes survive as a C string literal. *)
+let format_pos (fmt : Format.formatter) (pos : Pos.t) : unit =
+  Format.fprintf fmt "@[<hv 2>{%S,@ %d, %d, %d, %d}@]" (Pos.get_file pos)
```
AltGr: `%S` is the OCaml escaping and is not valid for C. Please use `String.quote` instead.

AltGr (commented):
Thanks for the tedious efforts on Windows compat 😅 
Some comments:
- I believe the command quoting in Clerk would be more reliably handled in `Ninja_utils` directly, avoiding the quote/unquote forward and back.
- Please keep the comments concise, I don`t need the code read out to me.
- I am not sure what the best solution is, but it's annoying to have interfaces polluted for the purpose of unit testing.

compiler/catala_utils/file.ml
rprimet: module Path moved besides File and not within, renamed functions

compiler/scalc/to_c.ml
rprimet: Thanks! Fixed

build_system/backend/ocaml.ml
rprimet: regarding moving quote/unquote to ninja_utils, this is probably the right solution but would require API changes in ninja_utils plus a release, so could be deferred to a later PR?

compiler/catala_utils/file.ml
rprimet: Also good point re. `win32` parameter everywhere, that was too noisy. Removed in favor of a ref.

build_system/backend/ocaml.ml
AltGr: We have full control over `ninja_utils` and can add a `pin-depends` until it gets released, so it shouldn't be too much trouble. My worry is that doing this here adds quite some code complexity (in particular, unquotings)) that we will later have to remove (well, and to be honest I have a huge pending PR on Clerk that I will have to rebase upon this)

If you are worried about complexifying your dev workflow (having to juggle between multiple git repos is never pleasintg), you can just `ln -s ../ninja_utils .` from the catala source, and dune should handle the linked version of `ninja_utils` transparently.

I can take care of the release after that if needed.

build_system/backend/backend_paths.mli
```diff
@@ -0,0 +1,27 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+(** Backend path/command helpers. Windows behaviour is testable on any host via
+    [Path.win32]. Internal to [Clerk_backends]. *)
+
+val pythonpath : string list -> string
+val classpath : backend:string -> string list -> string
```
denismerigoux, line 24: `classpath` should be used to make the paths where Java source code files corresponding to a class are stored? Not obvious from the name of the function.

build_system/backend/backend_paths.ml
```diff
@@ -0,0 +1,43 @@
+(* This file is part of the Catala build system, a specification language for
+   tax and social benefits computation rules. Copyright (C) 2020-2025 Inria,
+   contributors: Denis Merigoux <denis.merigoux@inria.fr>, Emile Rolley
+   <emile.rolley@tuta.io>, Louis Gesbert <louis.gesbert@inria.fr>
```
denismerigoux: You should put your name here instead of copy-pasting old irrelevant contributors attributions :)

build_system/backend/backend_paths.mli
denismerigoux: Apparently it's something else, you should document what these functions do with a comment

compiler/catala_utils/path.ml
```diff
@@ -0,0 +1,184 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2020 Inria, contributor:
+   Emile Rolley <emile.rolley@tuta.io>, Louis Gesbert <louis.gesbert@inria.fr>
```
denismerigoux: Ditto authors here

compiler/catala_utils/path.mli
```diff
@@ -0,0 +1,41 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2020 Inria, contributor:
+   Emile Rolley <emile.rolley@tuta.io>, Louis Gesbert <louis.gesbert@inria.fr>
```
denismerigoux: Ditto authors here

denismerigoux (approved):
Thanks Romain! We discuss it this afternoon

build_system/backend/backend_paths.mli
rprimet: comment added

compiler/catala_utils/path.ml
```diff
@@ -0,0 +1,184 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+(* Pure path/URL helpers. The target OS is read from [win32] (defaulting to the
+   host); tests override it to exercise Windows behaviour on any host. Internal:
+   [File]/[Message] expose the host-specialised versions. *)
```
AltGr, line 19: Ah, if making this a top-level module, we should expose its functions directly and skip the indirection through `File` ? The overridable `win32` flag doesn't affect the interface, and it seems the only other addition is the `~cwd` flag, which could be made optional (`?(cwd = Sys.getcwd())`) ?

compiler/catala_utils/path.ml
```diff
@@ -0,0 +1,184 @@
...
+
+(* Separator between entries of a path-list env var (PATH, PYTHONPATH,
+   CLASSPATH). *)
+let list_sep () = if !win32 then ";" else ":"
+
+let is_relative n =
+  if !win32 then
+    (String.length n < 1 || (n.[0] <> '/' && n.[0] <> '\\'))
+    && (String.length n < 2 || n.[1] <> ':')
+  else String.length n < 1 || n.[0] <> '/'
```
AltGr, line 38: So here we are duplicating the OCaml stdlib function to make the win32 flag overrideable in the purpose of running windows-targeted unit-tests on Linux, is that correct ?

I took some time wondering why this function was necessary compared to before and trying to spot the difference. Maybe explain why it needs to be here ? Or just run the Windows unit tests on Windows 😬 .

compiler/scalc/to_c.ml
```diff
@@ -482,6 +482,15 @@ let format_op (fmt : Format.formatter) (op : operator Mark.pos) : unit =
   | Add_dat_dur _ -> assert false (* needs specific printing *)
   | op -> Format.fprintf fmt "@{<blue;bold>%s@}" (Operator.name op)
 
+(* [String.quote] escapes the filename with the portable escape set so Windows
+   backslashes survive as a C string literal ([%S] would use OCaml's decimal byte
+   escapes, invalid in C). *)
```
AltGr, line 487: Indeed, `String.quote` is the correct escaping here, no need to refer to Windows in the comment.

build_system/clerk_utils/var.ml
```diff
@@ -20,14 +20,21 @@ open Catala_utils
 include Ninja_utils.Var
 (** Ninja variable names *)
 
+let cmd_only_names = ref String.Set.empty
```
AltGr: I am not in favour of this global ref: it's likely to blow up in our faces sometime in the future (e.g. ninja variables have some scope and can be shadowed, which this can't handle).

Some kind of variable typing is probably the best solution indeed, but the information should be stored within the `Var.t` themselves.

compiler/catala_utils/path.ml
rprimet: Done :-)

build_system/clerk_utils/var.mli
```diff
@@ -21,6 +21,20 @@ include module type of struct
...
+      positions -> {!make}. Dual-use (CATALA_EXE): [make], quote each command
+      ref.
+    - command ref: expands to one argument -> [quote_arg !var]; to several ->
+      bare [!var].
+    - command literal: spaceable (a path) -> {!quote_arg}; a flag -> bare
+      string.
+    - variable definition values: never contain a quote char ({!binding_words}
+      quotes cmd_only vars at emit; direct execs read values as argv).
+    - {!get_var} output is direct-exec argv: use as-is, never quote or strip. *)
+
```
AltGr, line 31: I tried to re-read this module's documentation several times, but it remains very hard to fully understand.

From how I understand it, we have two types that apply to both expressions and variables, in the same way:
- `atoms`: singletons that should never be split ; they may contain references to other variables only if those are atoms as well.
-  `lists`: may contain either atoms, or other lists. In the latter case, elements of the list are taken one by one just like elements of their parent (ie lists of lists are flattened)

Double-quoting must happen whenever an atom appears in a list context::
- either when defining a variable of type `list` from an `atom`
- or when inserting a variable of type `atom` inside a list
This only when outputting ninja variables or commands (and when calling out processes on Windows ??), and in addition to the ninja `$` quoting ; excluding the in/outs of the build statement which take unquoted atoms.

From what I understand, you also differentiate  atoms between the ones that may contain spaces and the ones that can't and shouldn't be quoted. This may indeed be needed if an `atom` variable appears in a `list` context where it should normally be quoted, am I correct ? What would be the cases where quoting such references by default would hurt ?

With all this, I get the feeling that quoting could be automatically handled when resolving an expression ?

rprimet:
(temporarily move back to draft to ensure no merging until we land the ninja_utils changes)

rprimet:
Integrated in #1072

PR #1075, opened 2026-07-20
Fix too large generated java code

vincent-botbol:
This PR fixes the issue with the java backend generated too large methods which would yield a "code too large" error.
We introduce an expression split mechanism in the lcalc pass which lifts large lambda-terms into topdefs and replaces the original ones with topdef calls Too large arrays are replaced in the same fashion by concatening the topdef call.
This is enabled by default in java but may also be freely used in the other backends.

Also, scalc scope variable definitions can be split in several functions. This is needed for java as the current concatenation of all expressions as a single block generates too large code that escape the previous mechanism detection. 

We introduce a `--split-threshold` argument that expects an integer (>100) that represents the limit on an expression size before splitting. As we assume a relation between the lcalc lambda terms and the java bytecode, we empirically determined that a "safe" upper bound would be of 10K nodes in the considered lambda-terms. Hence, that is the default value (only in java) if none was provided.

## Checklist

* [x] Update the CHANGELOG.md

  * [x] The corresponding PR is: https://github.com/CatalaLang/catala-examples/pull/58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,3 +14,9 @@ in behavior visible for the end-users of the tooling.
   the `--trace` mechanism:
   - Added support in the `Java` backend;
   - Added `clerk run --trace ...` options.
+
+* [#1075](https://github.com/CatalaLang/catala/pull/1075) Fixes a bug
+  where, for large Catala programs, the java generated code is too
+  large and would yield "code too large" error. We now split large
+  methods into smaller methods whenever necessary.
+  Also, removed law headings from positions in Java (same as C).
--- a/compiler/lcalc/split.ml
+++ b/compiler/lcalc/split.ml
@@ -1,0 +1,633 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2026 Inria, contributor:
+   Vincent Botbol <vincent.botbol@inria.fr>
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you may not
+   use this file except in compliance with the License. You may obtain a copy of
+   the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations under
+   the License. *)
+
+open Catala_utils
+open Shared_ast
+
+type size_mark = (int * typed mark) custom
+
+type 'a split_ctx = {
+  split_threshold : int;
+  decl_ctx : decl_ctx;
+  topdefs : (lcalc, size_mark) Shared_ast__Definitions.gexpr Var.Set.t;
+}
+
+(** Bottom-up fold mapping typed marks to size marks, i.e., each node's mark
+    will contain its sub-term (approximated) size. *)
+let add_size (type a) : (a, typed) gexpr -> (a, size_mark) boxed_gexpr =
+ fun e ->
+  let location_size = 6 in
+  let op_size : a operator -> int = function
+    (* Traces in backends outputs a [begin] corresponding to the tag
+       and its location and an [end] node with its value. *)
+    | Tag (ScopeCall _) -> 3 + location_size + 2
+    | Tag (FunCall _) -> 3 + location_size + 2
+    | Tag (ScopeVarDef _) -> 2 + 2 + 2 + location_size + 2
+    | Tag (LocalVarDef _) -> 3 + location_size + 2
+    | Tag (LocalTupDef { names }) -> 2 + List.length names + location_size + 2
+    | Tag BranchingCondition -> 2 + location_size + 2
+    | Tag (Branching None) -> 2 + location_size + 2
+    | Tag (Branching (Some _)) -> 3 + location_size + 2
+    | Tag Assertion -> 2 + location_size + 2
+    | Tag (Exception { label = None; _ }) -> 3 + location_size + 2
+    | Tag (Exception { label = Some _; _ }) -> 4 + location_size + 2
+    | DebugPrint _ | Sort _ | Add_dat_dur _ | Sub_dat_dur _ -> 2
+    | _ -> 1
+  in
+  let rec add_size (e : (a, typed) gexpr) : int * (a, size_mark) boxed_gexpr =
+    let (Typed { pos; ty = _ } as m) = Mark.get e in
+    let mk_mark n : size_mark mark = Custom { pos; custom = n, m } in
+    match Mark.remove e with
+    | EVar _ | EExternal _ | ELit _ | EEmpty | ECustom _ | EBad | EFatalError _
+      ->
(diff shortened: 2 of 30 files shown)


compiler/lcalc/split.ml
denismerigoux, line 1: I love it, extra clean and crisp

compiler/scalc/from_lcalc.ml
```diff
@@ -573,6 +578,117 @@ let rec translate_scope_body_expr ctx (scope_expr : 'm L.expr scope_body_expr) :
     RevBlock.rebuild (decl ++ statements)
       ~tail:(translate_scope_body_expr { ctx with ren_ctx } scope_let_next)
 
+let translate_split_scope_body_expr
```
denismerigoux, line 581: So this is the function that is called in the Lcalc -> Scalc translation for Java only?

compiler/scalc/to_java.ml
```diff
@@ -256,6 +252,10 @@ let rec format_lit (ppf : formatter) (l : lit Mark.pos) : unit =
   match Mark.remove l with
   | LBool true -> pp_print_string ppf "CatalaBool.TRUE"
   | LBool false -> pp_print_string ppf "CatalaBool.FALSE"
+  | LInt i when Z.fits_int64 i ->
+    fprintf ppf "new CatalaInteger(%s%s)"
```
denismerigoux, line 256: Nice readability optim

compiler/scalc/to_java.ml
```diff
@@ -268,6 +268,10 @@ let rec format_lit (ppf : formatter) (l : lit Mark.pos) : unit =
         (Mark.copy l (LInt (Q.num i)))
         format_lit
         (Mark.copy l (LInt (Q.den i)))
+  | LMoney e when Z.fits_int64 e ->
```
denismerigoux, line 271: Could even put a case where `e` is a round number of currency units to call `foUnits` instead of `ofCents`

compiler/scalc/to_java.ml
```diff
@@ -754,6 +754,18 @@ and format_block ?(toplevel = false) ctx ppf (block : Ast.block) =
       fprintf ppf "@[<hov 4>%a;@]" (format_expression ctx) expr;
       pp_print_space ppf ();
       format_stmts ~toplevel r
+    | (SLocalDecl { name = n, _; typ }, _)
```
denismerigoux, line 762: A little comment here would be good to describe which pattern this is matching?

CHANGELOG.md
```diff
@@ -14,3 +14,8 @@ in behavior visible for the end-users of the tooling.
   the `--trace` mechanism:
   - Added support in the `Java` backend;
   - Added `clerk run --trace ...` options.
+
+* [#1075](https://github.com/CatalaLang/catala/pull/1075) Fixes a bug
+  where, for large Catala programs, the java generated code is too
+  large and would yield "code too large" error. We now split large
+  methods into smaller methods whenever necessary.
```
denismerigoux, line 21: Also you removed `law_headings` from the code positions stored in the Java backend.

denismerigoux (approved):
Wonderful, thank you!

compiler/scalc/from_lcalc.ml
vincent-botbol: Yes, through the scalc `config.split_scope_var_defs` config flag

compiler/scalc/to_java.ml
vincent-botbol: Good idea. Fixed it up

CHANGELOG.md
vincent-botbol: Yes, amended.

compiler/scalc/to_java.ml
vincent-botbol: Done, this whole module is indeed quite cryptic.

compiler/lcalc/split.ml
```diff
@@ -0,0 +1,633 @@
...
+    | Tag (ScopeVarDef _) -> 2 + 2 + 2 + location_size + 2
+    | Tag (LocalVarDef _) -> 3 + location_size + 2
+    | Tag (LocalTupDef { names }) -> 2 + List.length names + location_size + 2
+    | Tag BranchingCondition -> 2 + location_size + 2
+    | Tag (Branching None) -> 2 + location_size + 2
+    | Tag (Branching (Some _)) -> 3 + location_size + 2
+    | Tag Assertion -> 2 + location_size + 2
+    | Tag (Exception { label = None; _ }) -> 3 + location_size + 2
+    | Tag (Exception { label = Some _; _ }) -> 4 + location_size + 2
+    | DebugPrint _ | Sort _ | Add_dat_dur _ | Sub_dat_dur _ -> 2
```
AltGr, line 47: This could take into accound ` Scalc.To_<backend>.op_needs_pos` as well, for more precision 🤔 
... but since it depends on the backend...

compiler/scalc/print.ml
```diff
@@ -188,27 +188,67 @@ and format_block
...
-  Format.pp_open_hovbox ppf 4;
-  Print.keyword ppf "let ";
   let () =
     match def with
     | SVar { var; expr; typ = _; visibility = _ } ->
+      Format.pp_open_vbox ppf 2;
+      Format.pp_open_hovbox ppf 4;
+      Print.keyword ppf "let ";
       format_var_name ppf var;
       Print.punctuation ppf " =";
```
AltGr, line 208: Unrelated, but just noticing this now: `Print.punctuation` assumes a 1-width punctuation, so this will lead to misalignment: the ` ` and `=` must be printed separately.

compiler/shared_ast/expr.mli
```diff
@@ -499,6 +499,7 @@ val compare : ('a, 'm) gexpr -> ('a, 'm) gexpr -> int
 
 val is_value : ('a any, 'm) gexpr -> bool
 val free_vars : ('a any, 'm) gexpr -> ('a, 'm) gexpr Var.Set.t
+val free_vars_marked : ('a any, 'm) gexpr -> (('a, 'm) gexpr, 'm mark) Var.Map.t
```
AltGr, line 503: Maybe detail that this returns the mark for the **first** occurence of each free var ?
(or are they expected to be equal for multiple occurences ?)

AltGr (approved):
Beautiful work, thanks!
Maybe add the new pass to `compiler/index.mld` ;)

compiler/scalc/print.ml
vincent-botbol: Should we do 
```
Format.fprintf ppf " ";
Print.punctuation ppf "=";
```
?
That looks weird :)

compiler/shared_ast/expr.mli
vincent-botbol: Indeed, added some doc.

compiler/lcalc/split.ml
vincent-botbol: Yes, this part is quite unsatisfying. I settled for this but it really depends on the backend.

vincent-botbol:
> Beautiful work, thanks! Maybe add the new pass to `compiler/index.mld` ;)

Thanks ! Done

PR #1077, opened 2026-07-23
Bump version to 1.2.1 everywhere

vincent-botbol:
While releasing 1.2.1, I forgot to bump `master` version. It doesn't impact the release as the opam files in the public/private repo are consistent but master's executables will say 1.2.0 instead of 1.2.1.


denismerigoux (approved):
Oh, thanks for checking! I missed that when reviewing your PR for 1.2.1 back a few weeks ago, I should have been more careful...

PR #1079, opened 2026-07-27
Better description for List.find

denismerigoux:


AltGr:
Checked and verified that it doesn't overflow the layout. Thanks.

PR #1080, opened 2026-07-27
Fix toplevel leftover traces

vincent-botbol:
This PR removes some leftover traces from the first of the "double-evaluation" that we perform. Incidentally, it makes the trace less precise as we lose what was done in the first beta-reduction. We might want to offer different level of details.


vincent-botbol:
Actually, the current behavior is probably what we want. It's closer to what the backends should output.

PR #1083, opened 2026-08-10
Merge back adjacent LawText nodes together before calling pandoc

denismerigoux:
This PR restores the correct styling by pandoc of blocks of legal text containing multi-line artefacts like table. Since we call pandoc separately on each `LawText` block and that our lexer/parser creates a new `LawText` block from each different line, pandoc never saw the whole tables but only lines of it, preventing it from rendering it correctly.


AltGr:
Ah, indeed, we overlooked this when we changed how lawtext is handled by the parser a while ago. The fix feels fine, but I am wondering why lawtext items are only getting merged when below a header, and not at toplevel ?

PR #1086, opened 2026-08-23
lexer: require a separator after multi-word keywords

### LINKED ISSUE #830: Parser is a bit too tolerant after multi-word keywords
Fun fact: space is not mandatory after multi-word keywords. For example, this is accepted:
```catala
declaration scope Foo:
  output f content integer depends onx content integer

scope Foo under conditiontrue:
  definition f of x equals 0
```

normally, this is refused because there is an ambiguity between identifier and keyword, that is solved by being greedy, so `definitionfoo` is lexed as an identifier. Here, because of the space within the keyword, we no longer have this ambiguity (or rather, it's hidden behind a different branch : keywords are not a subset or identifiers)

This is not critical, but can be surprising, and could lead to breakages later on if unnoticed in the code and we fix the bug.

ludovicc:
Fixes #830

## Problem

Multi-word keywords (`depends on`, `list of`, `optional of`, `with pattern`, `under condition`, `we have`, `map each`, `and then`, `or if list empty`, `but replace`) were matched by lexer regexps that did not enforce any separator afterwards. Input such as:

\`\`\`catala
scope Foo under conditiontrue:
  definition f of x equals 0
\`\`\`

was silently accepted, lexing `conditiontrue` as the keyword followed by a glued identifier.

## Fix

In the action of each multi-word keyword in the shared `lex_code` (used by the English, French and Polish lexers via the `MR_*` macros), check that the character immediately following the lexeme is not an identifier character; otherwise raise a lexing error at the keyword position.

The peek is implemented with `Sedlexing.mark`/`next`/`backtrack` so that the lexbuf state is restored when no error is raised.

## Testing

- New regression test `tests/parsing/bad/multiline_keyword_space.catala_en`
- Full clerk suite passes: 281 files / 705 tests, 100%

### DIFF
--- a/compiler/surface/lexer.cppo.ml
+++ b/compiler/surface/lexer.cppo.ml
@@ -481,6 +481,7 @@ let rec lex_code (lexbuf : lexbuf) : token =
       L.update_acc lexbuf;
       DATA
   | MR_DEPENDS ->
+      L.check_keyword_boundary lexbuf (Pos.from_lpos prev_pos) prev_lexeme;
       L.update_acc lexbuf;
       DEPENDS
   | MR_DECLARATION ->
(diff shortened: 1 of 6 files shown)


vincent-botbol:
Thanks for your contribution!
With our (cumbersome) multi-language lexer, we have some corner-cases that are not matched by your fix. 
For instance, this snipped would be valid:
```catala
scope A under condition$1=$2:
  ...
```
I would say that, in this particular situation, it would be acceptable. We could also exhaustively check for all authorized following tokens. That would be annoying to maintain however we don't have to patch the lexer regularly so this could be a reasonable solution.
We'll discuss this with the rest of the team when summer vacations are over.

PR #1088, opened 2026-08-25
Add machine-readable JSON output format for dcalc

ludovicc:
## Purpose

The `dcalc` command currently ignores `--output-format=json`: it always prints the human-readable AST via `Print.program`. This PR adds a machine-readable JSON export of the dcalc intermediate representation, enabling direct AST-level differential testing against external oracles (e.g. formal ports such as a Lean 4 runtime).

## Proposal

- New module `compiler/dcalc/json_export.ml` serializes the dcalc AST to `Yojson.Safe.t`. Every node is a JSON object with a `"tag"` field naming the constructor (`lit`, `var`, `app`, `op`, `abs`, `if`, `struct`, `struct_access`, `inj`, `match`, `tuple`, `tuple_access`, `default`, `pure_default`, `empty`, `error_on_empty`, `fatal_error`, `pos`), plus an `unsupported` fallback for constructors not expected in dcalc.
- Literals are serialized with the existing runtime string formatters (integers as strings to avoid precision loss).
- Scope bodies export `input_struct`, `output_struct`, the list of lets (each with `kind`/`var`/`typ`/`expr`) and the `return` expression.
- `driver.ml` branches on `Global.output_format = JSON` in both code paths of the dcalc command (whole program, and `-s Scope` selection which emits a JSON array of selected scopes).

Example:

```console
$ catala dcalc --output-format=json -s A test.catala_en
[{"scope":{"input_struct":"A_in","output_struct":"A","lets":[
  {"kind":"get","var":"x","typ":"integer",
   "expr":{"tag":"struct_access","e":{"tag":"var","name":"A_in"},"field":"x_in"}},
  {"kind":"set","var":"y","typ":"integer",
   "expr":{"tag":"error_on_empty","e":{"tag":"default", ...}}}}],
  "return":{"tag":"struct","name":"A","fields":{"y":{"tag":"var","name":"y"}}}}]
```

Human output is unchanged when `--output-format=json` is absent.

## Implementation notes

- Follows the same `--output-format=json` convention already used by `interpret` and `dependency-graph`.
- Operator names reuse `Print.operator_to_string` so JSON tags match the existing surface syntax.

### DIFF
--- a/compiler/dcalc/json_export.ml
+++ b/compiler/dcalc/json_export.ml
@@ -1,0 +1,182 @@
+(* This file is part of the Catala compiler, a specification language for tax
+   and social benefits computation rules. Copyright (C) 2026 Inria.
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you may not
+   use this file except in compliance with the License. You may obtain a copy
+   of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   License for the specific language governing permissions and limitations
+   under the License. *)
+
+(** JSON export of the [dcalc] intermediate representation.
+
+    Intended for machine consumption, e.g. differential testing of the Catala
+    runtime against external oracles. Every node is a JSON object with a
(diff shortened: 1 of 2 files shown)


vincent-botbol:
Thanks for the PR. I need to discuss this with the other maintainers first. My idea for this feature would have been to reuse the `json-data-encoding` library to increase maintainability and to be able to also generate a JSON schema but perhaps that's overkill. We'll let you know when we reach a decision.

PR #1091, opened 2026-08-28
Propagate documentation as javadoc in the generated java code

vincent-botbol:
See `tests/attributes/good/java_doc.catala_en`


compiler/scalc/to_java.ml
```diff
@@ -31,6 +31,46 @@ let pp_print_list_padded ?pp_sep pp ppf l =
...
+  let param_to_text (n, d) =
+    let d = String.trim d in
+    let s = Format.asprintf "@[<hov>%a@]" pp_print_text d in
+    asprintf "@@param %s @[<hov>%a@]" n pp_print_text s
+  in
+  let pp_content ppf s =
+    let s = Format.asprintf "@[<hov>%a@]" pp_print_text s in
+    let l = String.split_on_char '\n' s |> List.map (fun l -> "* " ^ l) in
+    let pp_item ppf s = fprintf ppf "@[<h>%s@]" s in
+    (pp_print_list ~pp_sep:pp_print_cut pp_item) ppf l
```
AltGr, line 48: It's not really an issue, but this will format the text blocks 80-columns wide without taking into accound the left margin.

AltGr (approved):
Nice, thanks !
Ideally the same work should be done for all backends, so that the APIs generated from Catala code can be understood on their own.

compiler/scalc/to_java.ml
vincent-botbol: Yes, it seemed like the most straightforward way to "pretty"-print it.

PR #1092, opened 2026-08-29
Fix --gen-external template to match ExternalType functor signature

Sarztak:
Fixes a bug in the `--gen-external` template generator where the generated OCaml code did not match the `ExternalType` functor signature: `equal` and `compare` were missing the `_pos` parameter, and `from_json` had incorrect signature `_pos t` instead of `_pos _s`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,3 +45,9 @@ in behavior visible for the end-users of the tooling.
 * [#1087](https://github.com/CatalaLang/catala/pull/1087/) Escape
   non-ascii characters in Java backend strings as unicode. All Java
   strings are now agnostic of encodings.
+
+* [#1092](https://github.com/CatalaLang/catala/pull/1092) Fixes a bug
+  in the `--gen-external` template generator where the generated OCaml
+  code did not match the `ExternalType` functor signature: `equal` and
+  `compare` were missing the `_pos` parameter, and `from_json` had
+  incorrect signature `_pos t` instead of `_pos _s`.
--- a/compiler/lcalc/to_ocaml.ml
+++ b/compiler/lcalc/to_ocaml.ml
@@ -945,6 +945,24 @@ let format_module_registration ctx fmt exports modname hash is_external =
   Format.pp_close_box fmt ();
   Format.pp_print_char fmt ' ';
   Format.pp_print_string fmt "]";
+  if is_external then begin
+    Format.pp_print_space fmt ();
+    Format.pp_print_string fmt "~types:[";
+    Format.pp_print_list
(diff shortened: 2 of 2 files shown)


build_system/clerk_backend/ocaml.ml
```diff
@@ -28,7 +28,7 @@ let ocaml_include = Var.make_vector "OCAML_INCLUDE"
 module OCaml_Flags = struct
   let ocaml_include_and_lib : (string list * string list) Lazy.t =
     lazy
-      (let link_libs = ["zarith"] in
+      (let link_libs = ["zarith"; "yojson"] in
```
vincent-botbol: Even though it looks very natural to do that, this would currently break a lot of stuff. Adding a runtime dependency is a heavy choice. 
We indeed want to add JSON capabilities but we did not yet settle on the right solution so we cannot have this as of yet.

CHANGELOG.md
```diff
@@ -45,3 +45,9 @@ in behavior visible for the end-users of the tooling.
 * [#1087](https://github.com/CatalaLang/catala/pull/1087/) Escape
   non-ascii characters in Java backend strings as unicode. All Java
   strings are now agnostic of encodings.
+
+* [#1092](https://github.com/CatalaLang/catala/pull/XXXX) Fixes a bug
```
vincent-botbol:
```suggestion
* [#1092](https://github.com/CatalaLang/catala/pull/1092) Fixes a bug
```

vincent-botbol (changes requested):
Thanks for the fixes! 
However, as mentioned, the e8f383a commit shouldn't be part of this PR.
For the time being, the workaround is to patch your clerk's environment variables through the `clerk.toml` file or via the clerk's `--vars` option to modify the `OCAML_INCLUDE` flag to add the `yojson` directory.
E.g.,:
```toml
... # clerk.toml content

[variables]
OCAML_INCLUDE=["-I","<path to your yojson lib>/yojson","-I","<path to your zarith lib>/zarith"]
```
(you can use `opam var lib:yojson` to retrieve the proper path).

build_system/clerk_backend/ocaml.ml
Sarztak: Hi Vincent,

Thank you for the alternative solution. I understand, that adding yojson directly could break things.

Sarztak:
Applied changes to CHANGELOG.md as suggested

vincent-botbol:
I deleted the commit adding yojson dependency and squashed the new changelog commit into the original one in order to speed up merging stuff. I hope that's fine with you.

vincent-botbol (approved):

PR #1093, opened 2026-09-01
Patch java files target installations and generate maven project files

### LINKED ISSUE #1094: Java targets are not easily exportable
The generated java files of `_targets` are not easily exportable as they require a `package <package_name>;` directive which should correspond to their parent directory name.

Ideally, users should be able to copy over the java target's directory in their project as is. As such, the installed runtime and stdlib and should be in under the current `_targets/java/libcatala/catala/{runtime/stdlib/..}` directories but directly in the `_targets/java/catala/{runtime/stdlib/..}` to match their package declaration (i.e., `package catala.stdlib;`, `package catala.runtime;`, etc.). This way, a simple `cp _targets/java/ my_project/src/main/java/` is sufficient to import the catala generated code.

vincent-botbol:
This PR attempts to fix an issue with java target generation where the generated files could not be imported easily.
These files should have a `package <package_name>;` directive so that the target directory can be copied over directly in an existing java project. Moreover, the `<package_name>` must follow java convention, which is capitalized camel case. Hence, for java, we adapt the directory name (and the `package` directive) as such.

Also fix duplicate copying of the stdlib in java.

Fixes #1094

### DIFF
--- a/build_system/clerk_backend/ocaml.ml
+++ b/build_system/clerk_backend/ocaml.ml
@@ -347,7 +347,8 @@ module Spec : Sig.Spec = struct
         | None -> ""
         | Some n -> Printf.sprintf "\n (public_name %s.%s)" n target.tname)
         (String.concat " "
-           (List.map String.to_id ("libcatala" :: target.dependencies)))
+           (List.map String.to_id (Scan.libcatala :: target.dependencies)
+           |> List.sort_uniq compare))
 
   let install_runtime ~config =
     let open File in
--- a/build_system/clerk_driver.ml
+++ b/build_system/clerk_driver.ml
@@ -685,6 +685,7 @@ let install_backend_targets
   if not (List.exists (fun t -> List.mem bk t.Clerk_config.backends) targets)
   then ()
   else
+    let is_java = config_backend bk = `Java in
     let bk_dir = target_dir / backend_subdir bk in
     let extensions =
       B.src_extensions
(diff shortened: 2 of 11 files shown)


vincent-botbol:
This PR is a bit hackish, I'd like to wait for @AltGr 's input to reach a better solution. This can be deployed for testing though.

vincent-botbol:
Ok, this looks reviewable. 
I also bit the bullet, read the maven documentation and added the generation of maven project files. One main for all targets in `_targets/java/pom.xml` and one for each target that are supposed to be self-contained as long as you installed its dependencies.

build_system/clerk_backend/ocaml.ml
```diff
@@ -347,7 +347,8 @@ module Spec : Sig.Spec = struct
         | None -> ""
         | Some n -> Printf.sprintf "\n (public_name %s.%s)" n target.tname)
         (String.concat " "
-           (List.map String.to_id ("libcatala" :: target.dependencies)))
+           (List.map String.to_id
+              (Scan.libcatala :: target.dependencies |> List.sort_uniq compare)))
```
AltGr: Normally there should be no possible conflict, but I'd have put the `uniq` after the `to_id`

build_system/clerk_driver.ml
```diff
@@ -694,15 +695,34 @@ let install_backend_targets
       else []
     in
     B.install_runtime ~config;
+    let target_transitive_deps (t : Clerk_config.target) =
```
AltGr: It's a bit sad to recompute this here since we have the full dependency graph and its transitive closure during `Clerk_rules.organise_modules` ;

it might be cleaner to add this info to `Clerk_rules.callback_info` to return it from that point.

build_system/clerk_driver.ml
```diff
@@ -727,9 +747,28 @@ let install_backend_targets
                       / basename src
                     else src
                   in
-                  copy_in ~dir ~src)
+                  if not is_java then copy_in ~dir ~src
```
AltGr, line 735: I think it would still be nicer to add this processing to the Clerk_backend implementation -- we already have `B.install_runtime` after all. The best granularity might be to have a shared `install_target` function parametrised with the `copy_file` function, and use that in every backend spec.

AltGr (approved):
Very nice, thanks!

build_system/clerk_driver.ml
vincent-botbol: Done

build_system/clerk_driver.ml
vincent-botbol: Gave it an honest try but it gets messy really quickly as a lot of clerk_rules stuff are involved. It would require to lift those as a weird bunch of arguments yielding a shaky abstraction.

PR #1098, opened 2026-09-07
Further clerk fixes

AltGr:
This is a follow-up for #1072

See the individual patches for details:
- fixes color forwarding to catala sub-processes
- fixes `clerk build` with cmxs targets
- changes `clerk run` to always use the cli-supplied backend(s), defaulting to `interpret`, notwithstanding what backends are declared in clerk.toml (was previously taking the intersection, which could lead to silently doing nothing)
- fixes backend analysis of non-target modules that depend on targets (resolving `clerk build` issue with `impot/revenu/tests/Interface` on `catala-examples`)
- changes `clerk ci` to only build targets instead of all files it could find (tests are still run on all files)

### DIFF
--- a/build_system/clerk_backend/flags.ml
+++ b/build_system/clerk_backend/flags.ml
@@ -109,11 +109,7 @@ let default ~code_coverage ~(trace : bool) ~inplace ~config =
           | Some e -> File.check_exec e
           | None -> Lazy.force Poll.catala_exe);
         ]);
-    def Var.catala_flags
-      (lazy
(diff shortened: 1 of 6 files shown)


vincent-botbol (approved):
Thanks, looks good. Tested each fixes, it worked as expected.

One unrelated issue is that ninja is yielding a weird error when you call `clerk test` on a module without any declared test (e.g., in `catala-examples`: `clerk test impot_revenu/tests/interface.catala_fr`)

PR #1099, opened 2026-09-08
Clerk: generate simplistic dependency Makefiles in the C backend

AltGr:


build_system/clerk_utils/module_graph.ml
```diff
@@ -0,0 +1,559 @@
...
+                (function
+                  | None -> Some info
+                  | Some conflict ->
+                    (* Note: until now this was allowed. However, targets
+                       select their contents by module name only, so this
+                       could only be for local modules ? We could switch to
+                       UIDs to support this, or somehow namespace the
+                       modules by dir.
+                       We need to implement something else than picking
+                       randomly, in any case *)
```
denismerigoux, line 112: I agree!

build_system/clerk_utils/module_graph.ml
```diff
@@ -0,0 +1,559 @@
...
+        then
+          Message.warning
+            "The module@ @{<blue>%s@}@ belongs@ to@ no@ target@ and@ appears@ \
+             to@ be@ unused"
+            m;
+        let check_conflicts () =
+          let _, conflict_targets =
+            (* We allow a module to get attached to multiple targets
+               (`base_targets`). However, this creates a conflict between any two
+               indepenednt users of the same module. The fix would be to attach
```
denismerigoux, line 439: typo

build_system/clerk_utils/module_graph.ml
```diff
@@ -0,0 +1,559 @@
...
+               uses@ it):@]@,\
+              \    @[<v>%a@]@,\
+               @,\
+               @[<hov>@{<bold>Hint:@} @{<blue>%s@}@ should@ be@ included@ in@ \
+               a@ unique@ target@ that@ is@ listed@ in@ the@ \
+               @{<cyan>dependencies@}@ field@ of@ all@ targets@ that@ might@ \
+               use@ it.@]@,\
+               @,\
+               @[<hov>The dependency graph in Dot format is available in@ \
+               @{<bold;blue>%t@}.@]@]"
```
denismerigoux, line 473: Have you tried running it on the CNAF project? Do you think they will be able to patch things up and understand this error message?

denismerigoux (approved):
Thanks @AltGr!

build_system/clerk_utils/module_graph.ml
AltGr: This PR merely moved the code to a separate file :) ; nevertheless, my hope was indeed that they would be able to grasp its meaning, yes. The problem would not occur at the moment, since they don't have target dependencies specifications at all (so each targets would independently include all the module it needs).

It *would* happen if you were to define a target that depends on both A and B, and there was an overlap between modules of A and B. It can be tested on `catala-examples` by trying to depend on both `allocations-familiales` and `fonpeps`, without having defined the intermediate `smic` target.

The solutions are to make A depend on B, B depend on A ; or, as in the `catala-examples` case above, define a shared dependency for the shared modules. The text tries to give hints in this direction, and since the error would appear after you meddled with dependencies I believe it should make sense...

Of course, if you have a better formulation in mind ? ;)
Another possibility is to get into the issue more in depth in the catala-book, and link to it in the error (I found recent `gcc` actually do that!)

denismerigoux: > Of course, if you have a better formulation in mind ? ;)

For the moment no :) 

> Another possibility is to get into the issue more in depth in the catala-book, and link to it in the error (I found recent gcc actually do that!)

I like this very much!

vincent-botbol (approved):
I am satisfied with the changes. I did not find any problems with it. I tried it on several scenarios including the vscode test plugin

PR #1100, opened 2026-09-11
clerk: keep catala's stderr out of the JSON output of exceptions

rprimet:
This keeps JSON output clean for parsing even in presence of warnings


AltGr (approved):
The fix seems correct! I'd maybe have found it clearer if the flag was `forward_stderr` rather than `merge_stderr`.

PR #1101, opened 2026-09-11
Introduce check-expected option  in clerk and catala to check variabl…

Arnaud-Bihan:
…e from the trace generated

## Checklist

* [ ] Update the CHANGELOG.md

* [ ] Update the Catala language server in case of breaking changes at https://github.com/CatalaLang/catala-language-server

* [ ] Syntax cheat sheet at `doc/syntax/syntax_*.catala_*` and `doc/syntax/catala_*.typ`.
* [ ] Syntax highlighting plugins in `syntax_highlighting/` (all of them).


build_system/clerk_backend/flags.ml
```diff
@@ -109,7 +108,8 @@ let default ~code_coverage ~(trace : bool) ~inplace ~config =
           | Some e -> File.check_exec e
           | None -> Lazy.force Poll.catala_exe);
         ]);
-    def Var.catala_flags (lazy (catala_flags @ includes));
+    def Var.catala_flags
+      (lazy (catala_flags @ (if trace then ["--trace"] else []) @ includes));
```
denismerigoux, line 112: Why such a move? It does not make the code clearer IMHO.

build_system/clerk_lib/clerk_cli.ml
```diff
@@ -386,6 +386,15 @@ let objects =
          compiler to compile those into objects (e.g. $(b,.o) files. Note that \
          this is done anyway when necessary, like when running backend tests."
 
+let check_expected =
```
denismerigoux, line 389: Rename variable, CLI option and everything else to `--check-trace-assertions`

build_system/clerk_lib/clerk_config.ml
```diff
@@ -146,6 +152,9 @@ let project_encoding () =
        (dft_field ~name:"target_dir" ~default:default_global.target_dir string)
        (dft_field ~name:"include_sources"
           ~default:default_global.include_sources bool)
+       (* Left as an option rather than defaulted: an absent key is reported as a
+          likely oversight, while an explicit "false" is a deliberate choice. *)
+       (opt_field ~name:"check_expected" @@ bool)
```
denismerigoux, line 157: I agree, but then what's the semantic difference between `--check-trace-assertions=false` and nothing on the CLI ?

build_system/clerk_lib/clerk_config.mli
```diff
@@ -32,6 +32,12 @@ type global = {
   catala_opts : string list;
   default_targets : string list;
   include_sources : bool;
+  check_expected : bool option;
+      (** Compare the values of the [#[testcase.variable]] attributes against
```
denismerigoux, line 36: OK for `#[testcase.variable]`, I'll add it to https://github.com/CatalaLang/catala-book/pull/62

build_system/clerk_utils/expected.ml
```diff
@@ -0,0 +1,307 @@
...
+
+(* Expected values are written by hand in the Catala source, with the surface
+   syntax of the file's language ("100,25 €", "$100.25", "2024-1-3"...), whereas
+   the trace holds them as rendered by [Catala_runtime.Json]: money as "100.25"
+   (two decimals, dot separator, no currency symbol), dates zero-padded, exact
+   decimals as a "num/den" rational, etc.
+
+   [normalize_value] parses the surface form and re-renders it through the very
+   same encoder the trace uses, so the two are comparable as plain strings by
+   construction rather than by replicating the runtime's formatting rules. *)
```
denismerigoux, line 45: I don't agree with this string/string comparison, even if re-using the existing JSON parsing/serializing to normalize the value is clever. I would argue that values should be compared by evaluating a deep-embedded Catala equality expression between the parsed expected value and the deep-embedding of the trace value. @vincent-botbol what do you thing ?

build_system/clerk_utils/expected.mli
```diff
@@ -0,0 +1,50 @@
...
+      (** What the trace holds at the declared path, [None] when the path leads
+          nowhere or the element carries no value *)
+}
+
+val normalize_value : string -> string
+(** Re-renders a value written in the surface syntax ("100,25 €", "$100.25",
+    "2024-1-3"...) the way the trace encoder does, so that both can be compared
+    as plain strings. *)
+
+val read_trace : File.t -> Yojson.Safe.t option
```
denismerigoux, line 36: You shouldn't need to have to read a trace serialized into JSON into a file, but merely access the in-memory representation of the trace accessible in the interpreter, no?

build_system/clerk_utils/var.mli
```diff
@@ -40,6 +40,10 @@ val catala_exe : string t
 val catala_flags : Expr.t t
 val runtime : string t
 
+val trace : Expr.t t
```
denismerigoux, line 43: Again, I don't see why the trace should be stored in a file between `clerk` and `catala` if we move `expected.ml` into `catala`.

build_system/clerk_utils/expected.ml
denismerigoux, line 1: This file should belong to `catala` and not `clerk` ; `catala` should parse the `#[testcase.variable]` attributes and check expected values in the interpreter, merely reporting to clerk any errors or statistics about passing or failing these assertions. This is the way we already do for regular assertions in tests.

build_system/clerk_driver.ml
```diff
@@ -983,8 +986,41 @@ let run_targets
...
+      let expected_empty = Scan.M.is_empty item.Scan.expected_variables in
+      let check_expected, trace_format =
+        match trace with
+        | None when not expected_empty ->
+          if check_expected then
+            Message.warning
+              "@[<hov>There are expected variables in %a but no trace file has \
+               been provided:@ disabling the \"check expected\" feature.@]"
+              pp_context ();
+          false, trace_format
```
denismerigoux, line 1006: I don't see why you would provide a trace file ahead of `clerk run`, since `clerk run`'s interpretation is supposed to create the trace you will check expected values from?

build_system/clerk_report.ml
```diff
@@ -634,11 +658,26 @@ let print_json ~(build_dir : string) (tests : file list) =
         success + file.successful, total + file.total)
       (0, 0) tests
   in
+  (* type expected = Clerk_utils.Expected.expected = {
+  name : string;
+  expected : string;
+  current_value : string option;
+} *)
```
denismerigoux, line 665: Forgotten comment

build_system/clerk_report.mli
```diff
@@ -31,13 +31,20 @@ type inline_test = {
   i_result : pos;  (** Same for the actual result in the destination file *)
 }
 
+type expected = Clerk_utils.Expected.expected = {
+  name : string;
+  expected : string;
+  current_value : string option;
+}
```
denismerigoux, line 38: Is it really useful to repeat the type definition here, why not leave the simple alias?

compiler/surface/lexer.cppo.ml
denismerigoux, line 1: This is not how we parse and declare an attribute! See https://github.com/CatalaLang/catala/blob/04f06f3f6cb3adc1dbad104e7a0ec7ff1111777d/compiler/shared_ast/definitions.ml#L456

build_system/clerk_lib/clerk_cli.ml
Arnaud-Bihan: Done, don't hesitate to point if I've forgotten a "expected" somewhere I tried to get rid of that notion

build_system/clerk_lib/clerk_config.mli
AltGr: In principle, `#[testcase.*]` is the namespace for the `testcase` plugin, so I wouldn't expect it to be handled by the core compiler ?

compiler/surface/lexer.cppo.ml
AltGr: Also note that attribute values don't have to be strings, they can be parsed to the surface AST.

PR #1108, opened 2026-09-16
Fix JSON input propagation from clerk to catala

### LINKED ISSUE #1105: `clerk run --input FILE` passes the file's contents on the command line
`clerk run --help` says `--input=FILE|JSON` "Reads a JSON value from the given string or file (- for stdin)". Given a file, clerk reads it and passes its contents to `catala` as a single command-line argument. Linux limits a single argument to 128 KB (`MAX_ARG_STRLEN`), so every input larger than that fails, although passing a file is exactly what one does for a large input. Reading from stdin fails the same way. `catala interpret` reads the same file without trouble.

**File** `big_input.catala_en`:

````
# Large input

```catala
declaration scope S:
  input xs content list of integer
  output n content integer

scope S:
  definition n equals number of xs
```
````

**Input** `big_input.json` (269 KB), created with `python3 -c "import json; json.dump({'xs': list(range(40000))}, open('big_input.json', 'w'))"`

**Command:** `clerk run big_input.catala_en -s S --input big_input.json`

**Output** (exit 125):

```
┌─[ERROR]─
│
│  Unexpected error: Unix.Unix_error(Unix.E2BIG, "create_process",
│  "/opt/opam/default/bin/catala")
│
└─
```

**Same with stdin:** `clerk run big_input.catala_en -s S --input - < big_input.json` fails with the same error.

**For comparison:** after that clerk run has built the project, `catala interpret _build/big_input.catala_en --stdlib=_build/libcatala --directory=_build -s S --input "$PWD/big_input.json"` prints `n = 40,000`.

**Expected:** clerk passes the file's path (or a temporary file for stdin) to `catala`. Otherwise a proper error message in place of the uncaught `Unix_error`.

### LINKED ISSUE #1107: A JSON number read as a `decimal` goes through a binary float, silently
The input schema accepts a JSON number for a `decimal`. It is read as a double, so `0.1` becomes 0.1000000000000000055… before Catala sees it, and an exact type turns inexact without a warning. The same value given as the string `"0.1"` is read exactly. JSON number literals are decimal text, so they could be parsed exactly as well.

This is essentially the input side of #1104 .

**File** `decimal_input.catala_en`:

````
# Decimal input

```catala
declaration scope S:
  input x content decimal
  output y content decimal

scope S:
  definition y equals x * 3.0
```
````

**Commands and outputs:**

| Command | Output |
|---|---|
| `clerk run decimal_input.catala_en -s S --input '{"x": "0.1"}'` | `y = 0.3` |
| `clerk run decimal_input.catala_en -s S --input '{"x": 0.1}'` | `y = 0.300,000…` |

**Expected:** JSON numbers for `decimal` and `money` parsed from their text, or at least a note in the documentation and the schema that exact values need strings.

### PR DESCRIPTION
**Pull request by vincent-botbol**
Fixes #1105 : clerk directly used catala's cli flag which parses the file/stdin and call catala through the CLI with the expanded JSON file potentially leading to an error due to a too large invoked command. We now simply forward the string argument and let catala parse the file.
Fixes #1107 : convert float to rational using `Q.of_string` in JSON inputs =>
                      `Q.of_float 0.1` => `- : Q.t = 3602879701896397/36028797018963968`
                      `Q.of_string (string_of_float) 0.1` => `- : Q.t = 1/10`
Following #1106 discussion: allow optional field input values to be `null` or an explicit option `"Absent" | { "Present": <val> }`

### CODE REVIEW

File: compiler/shared_ast/encoding.ml [hunk 1]
```diff
@@ -118,7 +118,7 @@ let rat_encoding : Val.t encoding =
                  "Unexpected runtime value %a instead of decimal while \
                   encoding to JSON"
                  Val.format v)
-           (fun f -> Val.V (Decimal, Q.of_float f));
+           (fun f -> Val.V (Decimal, Q.of_string (string_of_float f)));
```
AltGr, line 121: Wow, that's pretty ugly 😳 
It's... probably the best way though ? I mean... floats, anyway.
According to https://www.rfc-editor.org/info/rfc8259/#section-6,  Json numbers are actually arbitrary precision **but** may expect the parser to be limited to IEEE754 : this makes the above hack OK since OCaml floats are IEEE754. If we wanted to do better we should discard `Json_encoding`'s float encoding and wrap ours to `Q` ; we're unlikely to get such Json input from the wild though.

See also [this blog post](https://blog.json-everything.net/posts/numbers-are-numbers-not-strings/) ; we could always output our arbitrary-precision numbers into valid Json numbers, but since the purpose is to be read by third-party parsers we better use strings as a workaround.

compiler/shared_ast/encoding.ml
```diff
@@ -281,10 +281,16 @@ and generate_option_encoder ctx typ =
           (function None -> 0, "Absent", None | Some x -> 1, "Present", Some x);
       }
   in
+  let proj_null = function
+    | V (Enum { name = "Optional" | "Optionnel"; constr }, v) -> (
```
AltGr, line 285: I don't wamt to think about how we're going to handle the scaling of the internationalisation of this type 🙈

AltGr (approved):
A couple comments, but nothing that gets in the way of merging.

